PR #9, opened 2016-07-22
Add dropwizard

oibe:


oibe:
this diff was based on top of add_jaxrs2, so there are a lot of redundant changes.

PR #17, opened 2016-07-26
add jaeger-crossdock to opensource

oibe:


jaeger-java-crossdock/src/main/thrift/tracetest.thrift
```diff
@@ -0,0 +1,47 @@
+namespace java com.uber.jaeger.crossdock.tracetest
```
yurishkuro: this is already defined in /idl afaik

yurishkuro:
:shipit:

Issue #14, opened 2016-07-24
Improvement in equals method performance , comparing reference


nklmish:
We can improve the performance of `equals` method by first comparing whether two instances of the object belongs to the same reference.

E.g. In class `RemoteControllerSampler`, the `equals` method looks like this:
```java   
    @Override
    public boolean equals(Object sampler) {
        if (sampler instanceof RemoteControlledSampler) {
            RemoteControlledSampler remoteSampler = ((RemoteControlledSampler) sampler);
            synchronized (this) {
                ReentrantReadWriteLock.ReadLock readLock = remoteSampler.getLock().readLock();
                readLock.lock();
                try {
                    return this.sampler.equals(remoteSampler.sampler);
                } finally {
                    readLock.unlock();
                }
            }
        }
        return false;
    }
```

Now consider a scenario :

``` java
RemoteControllerSampler original = new RemoteControllerSampler(...);
RemoteControllerSampler copy = original;
IntStream.range(0, 100).parallel().forEach(
       loopCounter -> System.out.println(original.equals(copy))
);
```
We need to acquire and release lock and execute the equals logic N (100) times.

But if we perform the following check in the beginning of `equals` method:
``` java
if (this == sampler) return true
```

then the code will have no need to acquire any lock if both instances refer to the same reference.

To summarise, the improved `equals` method will look like this:
```java
   @Override
    public boolean equals(Object sampler) {
       if (this == sampler) return true;
        ....
    }
```

PR #18, opened 2016-07-26
equals comparing reference #14

nklmish:


yurishkuro:
looks good, thanks for the PR, @nklmish

yurishkuro:
Fixed #14

PR #22, opened 2016-07-27
add metrics emission to ServerFilter, and ClientFilter

oibe:


jaeger-jaxrs2/src/main/java/com/uber/jaeger/filters/jaxrs2/ClientFilter.java
```diff
@@ -43,10 +45,12 @@
     private final Tracer tracer;
     private final TraceContext traceContext;
     private final Logger logger = LoggerFactory.getLogger(ClientFilter.class);
+    private final Metrics metrics;
 
-    public ClientFilter(Tracer tracer, TraceContext traceContext) {
+    public ClientFilter(Tracer tracer, TraceContext traceContext, StatsFactory statsFactory) {
```
yurishkuro: I suggest adding another constructor that allows null as 3rd argument, and delegating to it from the existing one, for backwards compatibility

jaeger-jaxrs2/src/main/java/com/uber/jaeger/filters/jaxrs2/ClientFilter.java
```diff
@@ -79,6 +84,7 @@ public void filter(ClientRequestContext clientRequestContext, ClientResponseCont
             }
         } catch (Exception e) {
             logger.error("Client Filter Response:", e);
+            metrics.clientFilterExceptionResponse.inc(1);
```
yurishkuro: I don't understand where this exception can come from. `getProperty`, `getStatus` ?

jaeger-jaxrs2/src/main/java/com/uber/jaeger/filters/jaxrs2/ClientFilter.java
yurishkuro: actually, does this class need to be public? If it's always created via TraceUtils, it doesn't, which resolves the issue of constructor backwards compatibility

yurishkuro:
Decided this is not needed.

PR #27, opened 2016-08-13
Upgrade to latest OpenTracing API (v.0.13)

yurishkuro:
cc @oibe


jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -129,90 +90,50 @@ void reportSpan(Span span) {
...
+            return addReference(References.CHILD_OF, parent.context());
         }
 
         @Override
-        public io.opentracing.Tracer.SpanBuilder withOperationName(String operationName) {
-            this.operationName = operationName;
+        public io.opentracing.Tracer.SpanBuilder addReference(String referenceType, io.opentracing.SpanContext referencedContext) {
+            if (firstParent == null && (
+                    referenceType == References.CHILD_OF || referenceType == References.FOLLOWS_FROM)) {
+                this.firstParent = (SpanContext) referencedContext;
```
badiib: why do you need a cast here? isn't referencedContext already a SpanContext?

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -129,90 +90,50 @@ void reportSpan(Span span) {
...
         }
+        return extractor.extract(carrier);
     }
 
-    private abstract class AbstractSpanBuilder implements io.opentracing.Tracer.SpanBuilder {
+    private class SpanBuilder implements io.opentracing.Tracer.SpanBuilder {
 
         private String operationName = null;
         private long start;
+        private SpanContext firstParent;
```
badiib: what is a first parent? is there a second parent?

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -129,90 +90,50 @@ void reportSpan(Span span) {
...
+        public io.opentracing.Tracer.SpanBuilder asChildOf(io.opentracing.Span parent) {
+            return addReference(References.CHILD_OF, parent.context());
         }
 
         @Override
-        public io.opentracing.Tracer.SpanBuilder withOperationName(String operationName) {
-            this.operationName = operationName;
+        public io.opentracing.Tracer.SpanBuilder addReference(String referenceType, io.opentracing.SpanContext referencedContext) {
+            if (firstParent == null && (
+                    referenceType == References.CHILD_OF || referenceType == References.FOLLOWS_FROM)) {
```
badiib: a CHILD_OF makes sense, but why would FOLLOWS_FROM denote a parent?

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
yurishkuro: this is how the OT API is designed, when you getting a Span (or SpanContext) from the Tracer, you're getting a top-level interface, and subsequent operations inside the tracer must downcast it to the original impl. The only way to avoid that is to use generics, which will require all end-user code to be templatized on the type of tracer/span/spanCtx, which is not feasible.

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
yurishkuro: fair enough, I copied it from another project, but will rename to `parent`. We do not support multiple parents at this time anyway.

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
yurishkuro: That's a bit of a gray area. FOLLOWS_FROM in the spec still implied "the same trace", therefore it's a child of the previous span as far as our spanId/parentId are concerned.

badiib:
approved, i don't see any other issues.

oibe:
lgtm

PR #34, opened 2016-08-26
Adds jaeger-zipkin which makes Jaeger compatible with Zipkin tracing

adriancole:
Before, Jaeger used the same thrift definition as Zipkin. However, the
propagation and reporting layers are custom. This makes it possible to
reuse Jaeger in existing Zipkin systems.


adriancole:
cc @kristofa @michaelsembwever @bensigelman Fyi

I made the tests such that they test against Brave

jaeger-zipkin/src/main/java/com/uber/jaeger/propagation/b3/B3TextMapCodec.java
```diff
@@ -0,0 +1,116 @@
...
+      }
+    }
+
+    // Uber's parent ID is not nullable
+    if (parentID == null) {
+      parentID = traceID;
+    }
+
+    if (traceID != null && spanID != null) {
+      return contextFromString(String.format("%x:%x:%x:%x", traceID, spanID, parentID, flags));
```
adriancole: this is kindof gross, but the there's no way to construct the context naturally from outside its package.

jaeger-zipkin/src/test/java/com/uber/jaeger/propagation/b3/B3TextMapCodecTest.java
```diff
@@ -0,0 +1,170 @@
...
+import com.uber.jaeger.SpanContext;
+import io.opentracing.propagation.TextMap;
+import org.junit.Test;
+
+import java.net.URI;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import static org.junit.Assert.assertEquals;
```
adriancole: ps I guessed at the formatting.. seems to be intellij default

jaeger-zipkin/src/test/java/com/uber/jaeger/senders/zipkin/ZipkinSenderTest.java
```diff
@@ -0,0 +1,145 @@
...
+import org.junit.Test;
+import zipkin.junit.ZipkinRule;
+import zipkin.reporter.MessageEncoder;
+import zipkin.reporter.urlconnection.URLConnectionSender;
+
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+
+public class ZipkinSenderTest {
```
adriancole: this test was ported from UDPSenderTest

jaeger-zipkin/src/main/java/com/uber/jaeger/propagation/b3/B3TextMapCodec.java
yurishkuro: we can add public constructor

jaeger-zipkin/src/main/java/com/uber/jaeger/propagation/b3/B3TextMapCodec.java
```diff
@@ -0,0 +1,116 @@
...
+
+  /** Returns null when this is a root span. */
+  static Long nullableParentId(SpanContext context) {
+    return root(context) ? null : context.getParentID();
+  }
+
+  /** Returns true if this is the root span. */
+  static boolean root(SpanContext context) {
+    return context.getParentID() == context.getTraceID()
+        && context.getParentID() == context.getSpanID();
```
yurishkuro: The convention we used in Jaeger is the root has parent id == 0. There's also another invariant that traceId == spanID for the root span, but that's more of an optimization, not to be relied upon.

jaeger-zipkin/src/test/java/com/uber/jaeger/senders/zipkin/ZipkinSenderTest.java
yurishkuro: @adriancole having done it, do you think there'e enough commonality to have abase class for tests like this?  The expectations on the Sender somewhat involved, we might be able to encode them in an abstract test.

jaeger-zipkin/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Sending data to Zipkin
+Zipkin supports transports including Http and Kafka. You can configure Jaeger to send to a Zipkin server with
+`ZipkinSender`.
+
+For example:
+```java
+reporter = new RemoteReporter(ZipkinSender.create("http://localhost:9411/api/v1/spans"));
+tracer = new Tracer.Builder(serviceName, reporter, sampler)
+                   ...
```
yurishkuro: Q: because this is linked to Brave, does it mean we can reuse its Kafka transport?

yurishkuro:
@adriancole this is awesome! I made a few comments, main one about parent ID. Maybe the codecs need to adapt between Jaeger's representation of parent=0 for root span and Zipkin's model.

bensigelman:
(@adriancole thanks for doing this!)

jaeger-zipkin/README.md
adriancole: This is actually linked to
https://github.com/openzipkin/zipkin-java-reporter which brave will use in
the future. It has its own kafka reporter (and will soon have a scribe
one). The main goal of zipkin-java-reporter was to allow people to share
code (like for OT reporters), but without needing to bring in dependencies
on a different instrumentation project.

adriancole:
Just fixed the parent ID thing. Thanks for the quick review, Yuri. I think
it is possible to share test code between the zipkin-sender and the
previous one.

If you're up for it, I can give it a go post-merge.. would want to mildly
change the sender interface so that they are similar enough to be
compatible in tests.

jaeger-zipkin/src/test/java/com/uber/jaeger/senders/zipkin/ZipkinSenderTest.java
adriancole: made a general note.. I think there's a couple ways to do this.. maybe I can do it without affecting any of the main uber interfaces (ie by making an abstract method on the test case). I'll own at least trying this as a follow-up!

yurishkuro:
LGTM, merging

adriancole:
thanks!

PR #38, opened 2016-09-02
Add an event listener for spans

vprithvi:
@yurishkuro

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/SpanListener.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/SpanListener.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright (c) 2016, Uber Technologies, Inc
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package com.uber.jaeger;
+
+/**
+ * Listens to all spans that are being reported. Mainly intended to facilitate testing.
+ */
+public interface SpanListener {
+    void listen(Span span);
+}
--- a/jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
@@ -47,6 +47,7 @@ public class Tracer implements io.opentracing.Tracer {
     private final PropagationRegistry registry;
     private final String serviceName;
     private final Metrics metrics;
+    private SpanListener spanListener;
     private int ip;
 
     private Tracer(String serviceName, Reporter reporter, Sampler sampler, PropagationRegistry registry, Metrics metrics) {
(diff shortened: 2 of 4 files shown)


jaeger-core/src/main/java/com/uber/jaeger/SpanListener.java
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package com.uber.jaeger;
```
yurishkuro: I think we should move it to `events` package, as there will be other classes

jaeger-core/src/main/java/com/uber/jaeger/SpanListener.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Listens to all spans that are being reported. Mainly intended to facilitate testing.
+ */
+public interface SpanListener {
```
yurishkuro: I would call this EventListener, I am not certain it will always be "spans"

jaeger-core/src/main/java/com/uber/jaeger/SpanListener.java
```diff
@@ -0,0 +1,29 @@
...
+    void listen(Span span);
```
yurishkuro: s/listen/spanReported/

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -47,6 +47,7 @@
     private final PropagationRegistry registry;
     private final String serviceName;
     private final Metrics metrics;
+    private SpanListener spanListener;
```
yurishkuro: * Tracer should contain an EventListenerRegistry instance, always non-empty
* registry itself can implement EventListener so that tracer can always call `this.eventListenerRegistry.spanReported()` etc.
* tracer should have addEventListener method that will delegate to registry.addEventListener
* registry implements all listener methods by delegating to an internal list of listeners one at a time

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -82,6 +83,13 @@ Reporter getReporter() {
     void reportSpan(Span span) {
         reporter.report(span);
         metrics.spansFinished.inc(1);
+        if (spanListener != null) {
+            spanListener.listen(span);
+        }
+    }
+
+    public void setSpanListener(SpanListener listener) {
```
yurishkuro: s/../addEventListener/

jaeger-core/src/main/java/com/uber/jaeger/SpanListener.java
yurishkuro: there should be an EventListenerAdapter class that implements all EventListener methods as no-op. Your code in weaver template should extend the adapter and override spanReported method. Then if we add more methods to the interface in the future, code that extends the Adapter does not need to change.

adriancole:
Curious why not just make a wrapping reporter instead?

yurishkuro:
@adriancole 
> Curious why not just make a wrapping reporter instead?

 1. I would like to extend the interface to more events and potentially re-implement metrics emission as a listener rather than sprinkle the code with it
 1. in some internal unit tests we wanted to verify that a production configuration is "tracing enabled", which prevents us from replacing the reporter, but we can still add the listener.

adriancole:
your choice, but I'd wait for the api do evolve into something more than a
second reporter interface before adding it. the unit test concern seems
pretty weak to me.

adriancole:
ps I don't mean any offense by this.. just saying that you can evolve
your design by using a wrapping reporter that delegates into your
work-in-progress listener.

it seems to me the only thing you'd have to do is change how you do
your own unit tests :) ex how you decide something is production-ready
or not.

one of the advantages, is that it doesn't thrash upstream while you
figure out the design you actually need. Meanwhile, jaeger gets to
stay simpler, longer.

at the root of all of this is the change culture jaeger wants, too. ex
I tend towards least interfaces possible, which puts more burden on
developers to actually get multiple people to buy-into a change.
https://github.com/OpenFeign/feign/blob/master/HACKING.md

Mine isn't a universally acceptable idea, ex I think that it is valid
also to just merge and see.. especially since jaeger already has a
disclaimer on the readme saying "don't use this" :)

long story short: make a call! just nudging to consider the tech debt
aspects of adding apis.

jaeger-core/src/main/java/com/uber/jaeger/SpanListener.java
badiib: @yurishkuro if SpanListener should be EventListener, do you think "spanReported" is still the proper method name?

vprithvi:
@adriancole It turned out that we can use reflection to get the data we need for unit tests. I'm going to close this PR. Thanks for sharing the article!

adriancole:
:thumbsup:

PR #40, opened 2016-09-06
Record sampling strategy as span tags

yurishkuro:


jaeger-core/src/main/java/com/uber/jaeger/samplers/ConstSampler.java
```diff
@@ -21,11 +21,25 @@
...
 package com.uber.jaeger.samplers;
 
+import com.uber.jaeger.Constants;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
 public class ConstSampler implements Sampler {
+    public static final String TYPE = "const";
```
badiib: Is this available from the interface's perspective?

badiib:
LGTM, we can punt on making the type available on the interface if it ever becomes necessary

PR #47, opened 2016-09-09
Default peer service name to blank as it is required in Thrift Endpoint

### LINKED ISSUE #45: It is possible to report invalid spans (endpoint.serviceName == null)
In zipkin's thrift, Endpoint.serviceName is not an optional field. Conventionally, some set this to empty, but not serializing it is a mistake. I noticed this when using the jersey client, which might hint there's a bug there (as I did specify a serviceName).

This should be enforced in unit tests or similar as otherwise the only way to know is after it reaches zipkin.

yurishkuro:
Fixes #45

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/Span.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Span.java
@@ -72,7 +72,7 @@ public class Span implements io.opentracing.Span {
     public Endpoint getPeer() {
         synchronized (this) {
             if (peer == null) {
-                peer = new Endpoint();
+                peer = new Endpoint(0, (short) 0, "");
             }
         }
 
--- a/jaeger-core/src/main/java/com/uber/jaeger/Span.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Span.java
@@ -21,14 +21,14 @@
  */
 package com.uber.jaeger;
 
+import com.twitter.zipkin.thriftjava.Endpoint;
+import com.uber.jaeger.utils.Utils;
+import io.opentracing.tag.Tags;
+
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


jaeger-core/src/main/java/com/uber/jaeger/Span.java
```diff
@@ -72,7 +72,7 @@ public Tracer getTracer() {
     public Endpoint getPeer() {
         synchronized (this) {
             if (peer == null) {
-                peer = new Endpoint();
+                peer = new Endpoint(0, (short) 0, "");
```
adriancole: might want to return null.. ex null endpoint is ok, it is null serviceName that's the problem. an empty endpoint w/o an ip or serviceName or port, is probably worse than null.

adriancole:
PS verified this fixes it!

badiib:
LGTM pending the null conversation.

jaeger-core/src/main/java/com/uber/jaeger/Span.java
```diff
@@ -21,15 +21,15 @@
  */
 package com.uber.jaeger;
 
+import com.twitter.zipkin.thriftjava.Endpoint;
+import com.uber.jaeger.utils.Utils;
+import io.opentracing.tag.Tags;
```
yurishkuro: intelliJ kept insisting, so I let it

yurishkuro:
OK I figured it out. It was really a bug: reusing getPeer method as both lazy initializer and getter. I split them into plain getter and `getOrMakePeer`, the latter is only used by the methods that actually set the values on the peer, so the defaulting makes a lot more sense.

yurishkuro:
released as 0.8.1

adriancole:
thanks

PR #49, opened 2016-09-09
Exclude test classes from published jars

### LINKED ISSUE #48: Test classes are in release jars
For example `jaeger-core-0.8.0.jar` contains the class `com.uber.jaeger.SpanTest` which seems a bit weird to me. Is this intended? I'd expect the test classes to be published in `jaeger-core-0.8.0-tests.jar` if they need to be published at all. See also https://discuss.gradle.org/t/how-to-reference-a-test-jar-from-a-project-in-a-multi-project-build/5715.

yurishkuro:
Fixes #48


black-adder:
lgtm

jaeger-zipkin/src/main/java/com/uber/jaeger/senders/zipkin/ZipkinSender.java
```diff
@@ -51,6 +51,7 @@
   /**
    * @param endpoint The POST URL for zipkin's <a href="http://zipkin.io/zipkin-api/#/">v1 api</a>, usually
    *                 "http://zipkinhost:9411/api/v1/spans"
+   * @return new sender
```
badiib: is this comment accurate?

PR #50, opened 2016-09-12
Introduce Clock interface for time measurements

yurishkuro:
- Use nano-time for measuring duration
- Fix the tests, remove powermock dependency

- [ ] add more tests for duration calculation

### DIFF
--- a/jaeger-context/src/test/java/com/uber/jaeger/context/CallableTest.java
+++ b/jaeger-context/src/test/java/com/uber/jaeger/context/CallableTest.java
@@ -19,11 +19,8 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package com.uber.jaeger.filters.jaxrs2;
+package com.uber.jaeger.context;
 
-import com.uber.jaeger.Tracer;
-import com.uber.jaeger.reporters.InMemoryReporter;
-import com.uber.jaeger.samplers.ConstSampler;
 import io.opentracing.Span;
 import org.junit.Before;
 import org.junit.Test;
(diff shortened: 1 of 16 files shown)


yurishkuro:
@adriancole we discussed in the past how this lib used `(millis*1000 + nanos/1000)`, I now changed it to be "more better".

jaeger-core/src/main/java/com/uber/jaeger/utils/SystemClock.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+public class SystemClock implements Clock {
+
+    @Override
+    public long currentTimeMicros() {
+        return System.currentTimeMillis() * 1000;
+    }
+
+    @Override
+    public long currentTimeNanos() {
```
adriancole: I'd probably not use the word "Time" here, eventhough nanoTime does. maybe tick or nanoTick? https://google.github.io/guava/releases/19.0/api/docs/com/google/common/base/Ticker.html

adriancole: reason is that people might be confused that it relates to the other "time" in this interface which it doesn't

adriancole:
naming bikeshed aside, looks good.

jaeger-context/src/test/java/com/uber/jaeger/context/CallableTest.java
```diff
@@ -19,11 +19,8 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package com.uber.jaeger.filters.jaxrs2;
+package com.uber.jaeger.context;
```
yurishkuro: @oibe these tests weren't in the proper tests dir, so not sure if they were even exercised previously (intelliJ didn't understand them as tests)

jaeger-context/src/test/java/com/uber/jaeger/context/TracedExecutorServiceTest.java
```diff
@@ -41,22 +38,22 @@
...
-import static org.powermock.api.mockito.PowerMockito.when;
+import static org.mockito.Mockito.when;
 
 public class TracedExecutorServiceTest {
 
     TimeUnit TIME_UNIT = TimeUnit.MILLISECONDS;
     TracedExecutorService tracedExecutorService;
     ExecutorService wrappedExecutorService;
-    Tracer tracer = new Tracer.Builder("test-executor-service", new InMemoryReporter(), new ConstSampler(true)).build();
+    //Tracer tracer = new Tracer.Builder("test-executor-service", new InMemoryReporter(), new ConstSampler(true)).build();
```
yurishkuro: removing this

oibe:
lgtm

jaeger-context/src/test/java/com/uber/jaeger/context/CallableTest.java
oibe: Noted.

PR #51, opened 2016-09-13
Fix checkstyle and use google style guide

vprithvi:
This uses the google style guide and will fail the build when there are any checkstyle errors. Warning are ignored. 
See https://github.com/uber/jaeger-client-java/issues/30

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ clean:
 
 .PHONY: test
 test:
-	./gradlew test
+	./gradlew check
 
 .PHONY: release
 release: clean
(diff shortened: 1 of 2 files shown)


vprithvi:
@yurishkuro

yurishkuro:
I had a look at the build output, but didn't see even warnings from the checkstyle plugin. Is it not running in Travis?

yurishkuro:
Thanks, LGTM.  Let's merge this. The next thing before we close the issue is to try to get the google auto-formatter in.

PR #53, opened 2016-09-16
Support debug traces via HTTP header "jaeger-debug-id"

yurishkuro:
Recognize HTTP header or a TextMap key `jaeger-debug-id` which, if found in the carrier, forces the trace to be sampled as "debug" trace. The value of the header is recorded as the tag on the root span, so that the trace can be found in the UI using this value as a correlation ID.

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/SpanContext.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/SpanContext.java
@@ -154,7 +154,7 @@ public class SpanContext implements io.opentracing.SpanContext {
      * Create a new dummy SpanContext as a container for debugID string.
      * This is used when "jaeger-debug-id" header is passed in the request
      * headers and forces the trace to be sampled as debug trace, and the
-     * value of header recorded as a span tag to server as searchable
+     * value of header recorded as a span tag to serve as a searchable
      * correlation ID.
      *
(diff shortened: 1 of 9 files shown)


jaeger-core/src/main/java/com/uber/jaeger/SpanContext.java
```diff
@@ -123,14 +125,49 @@ public static SpanContext contextFromString(String value) throws MalformedTracer
...
+     */
+    boolean isDebugIDContainerOnly() {
+        return traceID == 0 && debugID != null;
+    }
+
+    /**
+     * Create a new dummy SpanContext as a container for debugID string.
+     * This is used when "jaeger-debug-id" header is passed in the request
+     * headers and forces the trace to be sampled as debug trace, and the
+     * value of header recorded as a span tag to server as searchable
```
black-adder: "is recorded"

oibe:
lgtm

jaeger-core/src/main/java/com/uber/jaeger/SpanContext.java
yurishkuro: [it] forces .. trace to be sampled .. and value .. recorded

yurishkuro: but s/server/serve/

PR #55, opened 2016-09-19
Use the verify task in google-formatter instead of checkstyle

vprithvi:
- The checkstyle checks weren't 100pc in sync with the formatter
  checks, so we solely rely on the formatter's checks
- The build now fails if code isn't formatted correctly

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,7 +14,6 @@ ext.apacheThriftVersion = '0.9.3'
 ext.jerseyVersion = '2.22.2'
 ext.slf4jVersion = '1.7.16'
 ext.jacksonVersion = '2.7.4'
-ext.checkstyleVersion = '7.1.1'
 
 ext.junitVersion = '4.12'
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -14,7 +14,6 @@ ext.apacheThriftVersion = '0.9.3'
 ext.jerseyVersion = '2.22.2'
 ext.slf4jVersion = '1.7.16'
 ext.jacksonVersion = '2.7.4'
-ext.checkstyleVersion = '7.1.1'
```
yurishkuro: I thought checkstyle checks not the formatting, but the actual code mistakes

vprithvi: I don't think checkstyle is that intelligent. Findbugs is more similar to `go vet`
From wikipedia, 
> The performed checks mainly limit themselves to the presentation and don't analyze content, and do not confirm the correctness or completeness of the program.

PR #57, opened 2016-09-20
Refactor configuration

vprithvi:
- The lazily instantiated tracer is now available in jaeger config
  instead of the jaxrs2 package
- StatsFactory is no longer required to get a TracerBuilder from the config
- Extract the tracer into a field where possible
- Add the ability get a NoopTracer to Tracer.java


vprithvi:
@yurishkuro

jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
```diff
@@ -67,19 +78,31 @@ public Configuration(
...
     Reporter reporter = reporterConfig.getReporter(metrics);
     Sampler sampler = samplerConfig.createSampler(serviceName, metrics);
     return new Tracer.Builder(serviceName, reporter, sampler).withMetrics(metrics);
   }
 
   public Tracer getNoopTracer() {
-    Reporter reporter = new NoopReporter();
-    Sampler sampler = new ConstSampler(false);
-    return new Tracer.Builder(serviceName, reporter, sampler).build();
+    return Tracer.getNoopTracer(serviceName);
```
yurishkuro: I'd prefer to keep this in the config. What's the reason for creating this method on the main Tracer?

jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/JaegerFeature.java
```diff
@@ -41,10 +42,10 @@ public boolean configure(FeatureContext featureContext) {
 
     switch (featureContext.getConfiguration().getRuntimeType()) {
       case SERVER:
-        featureContext.register(TracingUtils.serverFilter(jaegerConfig));
+        featureContext.register(TracingUtils.serverFilter(tracer));
```
yurishkuro: makes sense, unless here's a late binding issue - cc @oibe

jaeger-jaxrs2/src/main/java/com/uber/jaeger/filters/jaxrs2/Configuration.java
```diff
@@ -1,80 +0,0 @@
...
-
-import com.uber.jaeger.Tracer;
-import com.uber.jaeger.metrics.NullStatsReporter;
-import com.uber.jaeger.metrics.StatsFactory;
-import com.uber.jaeger.metrics.StatsFactoryImpl;
-
-import javax.ws.rs.client.ClientRequestContext;
-import javax.ws.rs.container.ContainerRequestContext;
-
-public class Configuration extends com.uber.jaeger.Configuration {
```
yurishkuro: why remove this?  I assume Dropwizard is not the only jaxrs framework .

jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
vprithvi: It's an artifact from a previous idea. I'll update.

jaeger-jaxrs2/src/main/java/com/uber/jaeger/filters/jaxrs2/Configuration.java
vprithvi: I agree that there are other jaxrs frameworks. I fail to see anything jaxrs specific in this configuration, and felt that it was adding unnecessary complexity. 
Anyone who uses this should be able to use jaeger.Configuration

yurishkuro: the  primary reason to keep this is for backwards compatibility. I suggest you restore the methods in TracingUtil that take this Config, but mark them deprecated.

vprithvi: Usually, I'd agree with you. In this particular case `TracingUtil` is the only class that uses this configuration, and is internal to `jaeger-java-client`.  Nobody within uber uses this configuration file separately, and I don't think that there is anybody outside using it either. 
Is there something else I'm missing?

jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/JaegerFeature.java
vprithvi: Spoke with @oibe who said that this looks fine

jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/Configuration.java
```diff
@@ -21,24 +21,33 @@
...
+    this.disable = disable;
   }
 
-  public Configuration setMetricRegistry(MetricRegistry registry) {
-    super.setStatsFactory(new StatsFactory(registry));
-    return this;
+  @Override
+  public Tracer getTracer() {
+    if (disable) {
+      return super.getNoopTracer();
```
yurishkuro: I don't think this makes sense. We're still using real Jaeger tracer, meaning it would still do context injection/extraction, still propagate baggage, etc. That's not at all what using a no-op tracer means in other languages. The opentracing-java repo now has a noop tracer implementation, let's use that (only make it a dependency of the dropwizard module).

vprithvi: I see, I was simply reverting it to what it used to do. I've updated the PR to use opentracing's nooptracer

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -29,7 +29,9 @@
 import com.uber.jaeger.propagation.Extractor;
 import com.uber.jaeger.propagation.Injector;
 import com.uber.jaeger.propagation.TextMapCodec;
+import com.uber.jaeger.reporters.NoopReporter;
 import com.uber.jaeger.reporters.Reporter;
+import com.uber.jaeger.samplers.ConstSampler;
```
yurishkuro: imports?

PR #59, opened 2016-09-21
Add errorprone plugin

vprithvi:
Moves the generated java sources into jaeger-generated gradle subproject and uses the errorprone compiler for all other subprojects. 
Also fixes bugs reported by errorprone. 
Fixes https://github.com/uber/jaeger-client-java/issues/56


jaeger-jaxrs2/src/test/java/com/uber/jaeger/propagation/SpanInfo.java
```diff
@@ -50,7 +52,12 @@ public SpanInfo(Span span) {
     traceID = stringContext.substring(0, stringContext.indexOf(':'));
     baggage = span.getBaggageItem(FilterIntegrationTest.BAGGAGE_KEY);
     sampled =
-        (stringContext.substring(stringContext.lastIndexOf(':') + 1).getBytes()[0] & (byte) 1) == 1;
+        (stringContext
+                        .substring(stringContext.lastIndexOf(':') + 1)
+                        .getBytes(StandardCharsets.UTF_8)[
+                    0]
+                & (byte) 1)
+            == 1;
```
vprithvi: I think the not google formatted version is easier to read in this case...

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -47,6 +47,7 @@
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
```
adriancole: didn't realize jaeger was a java 8 library..

jaeger-core/src/main/java/com/uber/jaeger/samplers/RemoteControlledSampler.java
```diff
@@ -148,6 +148,17 @@ public boolean equals(Object sampler) {
   }
 
   @Override
+  public int hashCode() {
```
adriancole: presumably things like this are because of presence of equals methods. Are we expecting people to try and do equivalence on types like this? If not, an easier path is to remove equals

jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/Configuration.java
```diff
@@ -45,7 +45,7 @@ public Configuration(
   }
 
   @Override
-  public Tracer getTracer() {
+  synchronized public Tracer getTracer() {
```
adriancole: wow errorprone caught this? be interested in how it figured it out!

jaeger-jaxrs2/src/test/java/com/uber/jaeger/propagation/SpanInfo.java
adriancole: this is the problem with doing things like format as a mandatory step at check-in. makes me hate plugins like formatters and checkstyle. one way is to rewrite it into separate lines.

jaeger-core/build.gradle
```diff
@@ -86,12 +67,3 @@ jar {
         attributes('Implementation-Title': 'jaeger-core', 'Implementation-Version': project.version)
     }
 }
-
-task jaegerVersion << {
```
yurishkuro: I think this needs to remain in this jar. The property is read by the main tracer, it doesn't make sense for it to be in a thrift-only jar.

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
yurishkuro: it shouldn't be - iirc the gradle even says 1.7 explicitly

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -174,7 +175,8 @@ void reportSpan(Span span) {
     public io.opentracing.Tracer.SpanBuilder addReference(
         String referenceType, io.opentracing.SpanContext referencedContext) {
       if (parent == null
-          && (referenceType == References.CHILD_OF || referenceType == References.FOLLOWS_FROM)) {
+          && (Objects.equals(referenceType, References.CHILD_OF)
+              || Objects.equals(referenceType, References.FOLLOWS_FROM))) {
```
yurishkuro: the use of `==` was definitely a bug, would only work for enums, but the constants are Strings.

But it doesn't mean we need to use `Objects.equals()`, instead use `References.CHILD_OF.equals(...)`

jaeger-core/src/main/java/com/uber/jaeger/samplers/ConstSampler.java
```diff
@@ -65,9 +66,17 @@ public boolean equals(Object other) {
     return false;
   }
 
+  @Override
+  public int hashCode() {
+    int result = (decision ? 1 : 0);
+    result = 31 * result + (tags != null ? tags.hashCode() : 0);
+    return result;
+  }
```
yurishkuro: can we not disable whatever error the plugging was giving? We do not need hashcodes in samplers.

jaeger-core/src/main/java/com/uber/jaeger/samplers/RemoteControlledSampler.java
yurishkuro: if the checker cannot be disabled, I am inclined to agree, we can simply rename equals and define it on the Sampler interface to be required.

jaeger-generated/build.gradle
```diff
@@ -0,0 +1,50 @@
+apply plugin: 'org.jruyi.thrift'
+
+description = 'Core library for jaeger-client'
```
yurishkuro: wrong descr

build.gradle
```diff
@@ -71,3 +72,7 @@ subprojects {
 
     apply from: '../gradle/publish.gradle'
 }
+
+configure(subprojects.findAll {it.name != 'jaeger-generated'}) {
+   apply plugin: 'net.ltgt.errorprone'
```
yurishkuro: are we sure we cannot simply configure this plugin to not run against generated code? The docs

https://github.com/tbroyer/gradle-errorprone-plugin#advanced-usage

seem to suggest it's doable, but probably requires a lot of gradle-fu.

build.gradle
```diff
@@ -71,3 +72,7 @@ subprojects {
 
     apply from: '../gradle/publish.gradle'
 }
+
+configure(subprojects.findAll {it.name != 'jaeger-generated'}) {
```
yurishkuro: I'd prefer to name it more precisely, jaeger-thrift. If or when we have something else that's generated, we can revisit the naming.

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
vprithvi: @adriancole Objects is available on Java 7 https://docs.oracle.com/javase/7/docs/api/java/util/Objects.html

jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/Configuration.java
vprithvi: I think it is because this class inherits from `jaeger/Configuration.java` in which I had declared `getTracer()` to be synchronized

jaeger-core/build.gradle
vprithvi: My bad, will move it back

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
vprithvi: Why prefer one over the other?

jaeger-core/src/main/java/com/uber/jaeger/samplers/ConstSampler.java
vprithvi: Yeah, we can suppress the warnings as described in http://errorprone.info/bugpattern/EqualsHashCode

Could you elaborate on why we have `equals` in the first place?

jaeger-generated/build.gradle
vprithvi: Will update

build.gradle
vprithvi: Will rename this

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
yurishkuro: if it's in 7 then fine

jaeger-core/src/main/java/com/uber/jaeger/samplers/ConstSampler.java
yurishkuro: remote sampler needs them

build.gradle
vprithvi: I looked at that earlier, and while it offers more control over where in the lifecycle this is executed, it doesn't seem to support exclusions.
Exclusions are thorny because errorprone is a javac compiler enhancement and is run in place of the plain old java compiler. By splitting `java-generated` off into it's own subproject, we can simply compile it with the regular javac and pass down the compiled files as dependencies. 
Unless there is some way of informing errorprone enhanced javac that some files are generated and it should exclude them, I don't see how this can be solved elegantly.

vprithvi: Looking at the generated files, all of them have something like the following in them
`@Generated(value = "Autogenerated by Thrift Compiler (0.9.3)", date = "2016-09-21")` 
I'll look into whether we can make error prone respect this. Their [documentation](http://errorprone.info/docs/flags) doesn't show anything promising.

adriancole:
ahh whoops. bad memory!

build.gradle
vprithvi: Error prone has an [undocumented](https://github.com/google/error-prone/issues/463) flag to ignore generated code that has an `@Generated` annotation. 
Thrift 0.9.3 has a bug where the `@Generated` annotation isn't applied to classes that only contain enums. (e.g.  `com.twitter.zipkin.thriftjava.AnnotationType`)

jaeger-core/build.gradle
```diff
@@ -88,9 +69,9 @@ jar {
 }
 
 task jaegerVersion << {
-    def propFile = new File(projectDir,
-            'src/main/gen-java/com/uber/jaeger/jaeger.properties'
-    ).getAbsolutePath()
+    def dirPath = 'src/main/resources/com/uber/jaeger/'
+    new File(projectDir, dirPath).mkdirs()
+    def propFile = new File(projectDir, "${dirPath}jaeger.properties").getAbsolutePath()
```
yurishkuro: does it need to be in `.gitignore` if you're moving it from gen-java?

jaeger-jaxrs2/src/test/java/com/uber/jaeger/propagation/SpanInfo.java
```diff
@@ -49,8 +51,12 @@ public SpanInfo(Span span) {
     String stringContext = span.getContext().contextAsString();
     traceID = stringContext.substring(0, stringContext.indexOf(':'));
     baggage = span.getBaggageItem(FilterIntegrationTest.BAGGAGE_KEY);
-    sampled =
-        (stringContext.substring(stringContext.lastIndexOf(':') + 1).getBytes()[0] & (byte) 1) == 1;
+    byte byteAfterColon =
+        stringContext
+                .substring(stringContext.lastIndexOf(":") + 1)
+                .getBytes(StandardCharsets.UTF_8)[
+            0];
```
yurishkuro: this code doesn't make sense (I know you didn't write it, but let's fix it). The value after the last `:` is a hex-encoded byte, not a raw byte converted to char. So we should be simply parsing it as int 

```
String lastNumber = stringContext.substring(stringContext.lastIndexOf(":") + 1);
int flags = Integer.parseInt(lastNumber);
sampled = flags & 1 == 1;
```

or even simpler, can't we just cast `span.getContext()` to jaeger's SpanContext and use `isSampled()`?

jaeger-thrift/build.gradle
```diff
@@ -0,0 +1,41 @@
+apply plugin: 'org.jruyi.thrift'
+
+description = 'Core library for jaeger-client'
```
yurishkuro: wrong descr

jaeger-jaxrs2/src/test/java/com/uber/jaeger/propagation/SpanInfo.java
vprithvi: done

jaeger-thrift/build.gradle
vprithvi: done

jaeger-core/build.gradle
vprithvi: done

PR #65, opened 2016-09-29
Expose interceptor callbacks and custom operation name for extension classes

atanasbalevski:


atanasbalevski:
Hi,
the baggage naming was just arbitrary, you're right that the name should be better.
My motivation is to provide a way to add more information on the span and then transport it to the eventual child spans (on the other side of the wire) if needed.

The original motivation for this PR was to expose the operationName, which is hard to override otherwise (e.g. if i register another interceptor)

yurishkuro:
> My motivation is to provide a way to add more information on the span and then transport it to the eventual child spans (on the other side of the wire) if needed.

Baggage is not a facility that should be abused, and it's definitely not a mechanism for passing parameters to downstream calls. So what is the actual use case / example?

jaeger-apachehttpclient/src/main/java/com/uber/jaeger/httpclient/TracingRequestInterceptor.java
```diff
@@ -77,6 +77,8 @@ public void process(HttpRequest httpRequest, HttpContext httpContext)
         Tags.PEER_PORT.set(clientSpan, (short) host.getPort());
       }
 
+      onSpanStart(clientSpan, httpRequest);
```
yurishkuro: I would pass httpcontext as well.  And we can have a similar beforeSpanFinished() method in the response interceptor, just to be symmetrical (in case users want to add more info like error codes or what not).

atanasbalevski:
the example is if someone wants to add the Tags.COMPONENT, or set/override any of the Tags.* that were already set.

as i mentioned, these are OK to be modified by another RequestInterceptor, however my main goal for the PR is to get the operationName extensible

yurishkuro:
Do you mind adding some documentation?
  * javadoc comments for the new methods explaining when they are called and giving an example of the extra data each method can add to the span (e.g. where would you get operation name?)
  * a Custom Extensions section to the README showing how extensions can be done

Otherwise LGTM.

atanasbalevski:
(yay)
Thanks!

PR #71, opened 2016-10-05
Fix typo in configuration and add convenience setter for metricRegistry

vprithvi:
- dropwizard configuration users can easily set a dropwizard metricRegistry


jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
```diff
@@ -98,10 +98,15 @@ public Configuration(
     return tracer;
   }
 
+  @Deprecated
```
yurishkuro: add comment explaining what should be used instead

vprithvi: done

PR #74, opened 2016-10-10
allow runnables without current span to run

rubbish:
Otherwise if a Runnable is submitted without a current span in the TraceContext, it will blow up with a empty stack exception.

This was necessary as I wanted to use a global executor service that was decorated with `TracedExecutorService`. We haven't yet been able to switch over everything to necessarily have a span, so I made the `TracedExecutorService` and friends handle that more gracefully.

I could instead push this further up into the `TracedExecutorService` instead of mucking with the Runnable and Callable. 

Thoughts?


yurishkuro:
lgtm

vprithvi:
lgtm as well

PR #75, opened 2016-10-12
set host on all zipkin annotations for queries

rubbish:
I found this while trying to figure out why zipkin's annotation queries weren't working. It seems that zipkin uses the serviceName in the endpoint as a search parameter: https://github.com/openzipkin/zipkin/blob/master/zipkin-storage/elasticsearch/src/main/java/zipkin/storage/elasticsearch/ElasticsearchSpanStore.java#L115 

This is from https://github.com/openzipkin/zipkin/pull/1292

So I changed the zipkin span converter to always send along the host on all the binary and non binary annotations. After that it, everything works quite swimmingly.

### DIFF
--- a/jaeger-zipkin/src/main/java/com/uber/jaeger/senders/zipkin/ZipkinSender.java
+++ b/jaeger-zipkin/src/main/java/com/uber/jaeger/senders/zipkin/ZipkinSender.java
@@ -167,7 +167,7 @@ public final class ZipkinSender implements Sender {
   // serviceName/host is needed on annotations so zipkin can query
   // see https://github.com/uber/jaeger-client-java/pull/75 for more info
   Span backFillHostOnAnnotations(Span span) {
-    Endpoint host = getLocalComponentEndpoint(span);
+    Endpoint host = getServiceEndpoint(span);
 
     if (host != null) {
       for (BinaryAnnotation binaryAnnotation : span.getBinary_annotations()) {
@@ -186,10 +186,10 @@ public final class ZipkinSender implements Sender {
     return span;
   }
 
-  private Endpoint getLocalComponentEndpoint(Span span) {
-    for (BinaryAnnotation binaryAnnotation : span.getBinary_annotations()) {
-      if (Constants.CORE_ANNOTATIONS.contains(binaryAnnotation.getKey()) && binaryAnnotation.isSetHost()) {
-        return binaryAnnotation.getHost();
+  private Endpoint getServiceEndpoint(Span span) {
+    for (Annotation annotation : span.getAnnotations()) {
+      if (Constants.CORE_ANNOTATIONS.contains(annotation.getValue()) && annotation.isSetHost()) {
+        return annotation.getHost();
       }
     }
 
--- a/jaeger-zipkin/src/main/java/com/uber/jaeger/senders/zipkin/ZipkinSender.java
+++ b/jaeger-zipkin/src/main/java/com/uber/jaeger/senders/zipkin/ZipkinSender.java
@@ -21,7 +21,7 @@
  */
 package com.uber.jaeger.senders.zipkin;
 
-import com.twitter.zipkin.thriftjava.Span;
+import com.twitter.zipkin.thriftjava.*;
 import com.uber.jaeger.exceptions.SenderException;
 import com.uber.jaeger.senders.Sender;
 
@@ -94,7 +94,7 @@ public final class ZipkinSender implements Sender {
    */
   @Override
   public int append(Span span) throws SenderException {
-    byte[] next = encoder.encode(span);
+    byte[] next = encoder.encode(backFillHostOnAnnotations(span));
     int messageSizeOfNextSpan = delegate.messageSizeInBytes(Collections.singletonList(next));
     // don't enqueue something larger than we can drain
     if (Integer.compare(messageSizeOfNextSpan, delegate.messageMaxBytes()) > 0) {
@@ -160,5 +160,32 @@ public final class ZipkinSender implements Sender {
       throw new SenderException("Failed to close " + delegate, e, n);
     }
     return n;
+  }
+
+  // serviceName/host is needed on annotations so zipkin can query
+  // see https://github.com/uber/jaeger-client-java/pull/75 for more info
+  Span backFillHostOnAnnotations(Span span) {
+    Endpoint host = null;
+    for (BinaryAnnotation binaryAnnotation : span.getBinary_annotations()) {
+      if (zipkincoreConstants.LOCAL_COMPONENT.equals(binaryAnnotation.getKey())) {
+        host = binaryAnnotation.getHost();
+      }
+    }
+
+    if (host != null) {
+      for (BinaryAnnotation binaryAnnotation : span.getBinary_annotations()) {
+        if (!binaryAnnotation.isSetHost()) {
+          binaryAnnotation.setHost(host);
+        }
+      }
+
+      for (Annotation annotation : span.getAnnotations()) {
+        if (!annotation.isSetHost()) {
+          annotation.setHost(host);
+        }
+      }
+    }
+
+    return span;
   }
 }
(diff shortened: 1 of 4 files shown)


jaeger-core/src/main/java/com/uber/jaeger/reporters/protocols/ThriftSpanConverter.java
```diff
@@ -74,7 +74,7 @@
     List<LogData> logs = span.getLogs();
     if (logs != null) {
       for (LogData logData : logs) {
-        annotations.add(new Annotation(logData.getTime(), logData.getMessage()));
+        annotations.add(new Annotation(logData.getTime(), logData.getMessage()).setHost(host));
```
yurishkuro: I do not believe this is necessary. Can you demonstrate the instrumentation code that results in non-searchable annotations? AFAIK the (old Scala) zipkin backend will extract service name from *any* annotations found on the span, and use that service name for indexing all other annotations. We already provide service name in the core annotations, _and_ in the `lc` (local component) annotation for non-rpc spans. Including service name in *all* annotations that come from the span logging API is only going to bloat the span size.

cc @adriancole

jaeger-core/src/main/java/com/uber/jaeger/reporters/protocols/ThriftSpanConverter.java
```diff
@@ -119,13 +119,13 @@
         // Every value is converted to string because zipkin search doesn't
         // work well with ints, and bytes.
         Object tagValue = tags.get(tagKey);
-        binaryAnnotations.add(buildBinaryAnnotation(tagKey, tagValue));
+        binaryAnnotations.add(buildBinaryAnnotation(tagKey, tagValue, host));
```
yurishkuro: I am 100% certain this is not needed for tags.

jaeger-core/src/main/java/com/uber/jaeger/reporters/protocols/ThriftSpanConverter.java
```diff
@@ -119,13 +119,13 @@
...
       }
     }
     return binaryAnnotations;
   }
 
-  private static BinaryAnnotation buildBinaryAnnotation(String tagKey, Object tagValue) {
+  private static BinaryAnnotation buildBinaryAnnotation(String tagKey, Object tagValue, Endpoint host) {
```
yurishkuro: I am 100% certain this is not needed for tags.

jaeger-core/src/main/java/com/uber/jaeger/reporters/protocols/ThriftSpanConverter.java
rubbish: This might be something that only the zipkin elasticsearch span store is running into. I'm going to try and verify against other storages. I totally agree that it'd bloat span sizes which is why I was a little hesitant to add it.

adriancole:
Annotation and binary annotations are only queryable when they have an
endpoint attached. This is because indexing is primarily by service name.
The indexing scheme in cassandra hasn't changed since it was scala.

jaeger-core/src/main/java/com/uber/jaeger/reporters/protocols/ThriftSpanConverter.java
adriancole: 100%?

adriancole:
The behavior of backfilling endpoint before reporting a span has existed for a very long time in finagle, which was the first tracer. This is for both annotations and binary annotations

https://github.com/twitter/finagle/blob/develop/finagle-zipkin-core/src/main/scala/com/twitter/finagle/zipkin/core/Span.scala#L61

queries without a service name aren't fully supported. Hence, it is not actually possible for the UI to omit the serviceName.

yurishkuro:
I think this is a difference between Zipkin backend and Jaeger backend. This is what we have (omitting fluff)

```go
func (s *spanStorage) indexSpanByAnnotations(session *gocql.Session, span *zipkincore.Span) error {
	serviceName := strings.ToLower(getEmittingServiceName(span))
...
	for _, anno := range span.GetAnnotations() {
...
			baseKey := serviceName + IndexDelimiter + anno.GetValue()
			// Diverging from Zipkin Scala Collector, which uses the timestamp of the "min" annotation (via full sort),
			// while we're just using the latestTs
			q := query.Bind([]byte(baseKey), bucketNo, latestTs, traceID)
			err2 := execAndReport(q, s.metrics.annotationsIndex)
```

the "emitting service" is determined by core annotations or `lc`, and then that service name applies to all tags and logs.

adriancole:
This would work for spans that contain a single-host. This is mostly the
case even in RPC spans when a collector is acting on data sent directly
from instrumentation (because ex a multi-host span is sent for each host).
It wouldn't work for those copying traces via POST or those who have a
pipeline that aggregates traces together before storing. It wouldn't work
because we'd have ambiguity in deciding which host to apply to which event.

We've often got into trouble guessing data on behalf of the reporter. Most
troubling was timestamps, and this has all but been removed at the cost of
lots of hours. I think it would be a good choice to have jaeger's
ZipkinReporter act like other reporters and resolve locally vs attempt to
make zipkin work like jaeger knowing it has to deal with multi-host spans
still. When we get to single-host spans, something like what was mentioned
could make a lot of sense.

rubbish:
So with all the discussion, I'm back to believing this is necessary for the zipkin integration to work correctly. 

Going to reopen it.

adriancole:
I'm guessing that folks here will prefer to send "sparse spans" by default,
and instead have the zipkin sender component backfill them. It would
accomplish the same thing, but not require sending more data to uber's
custom collectors.

ex here
https://github.com/uber/jaeger-client-java/blob/master/jaeger-zipkin/src/main/java/com/uber/jaeger/senders/zipkin/ZipkinSender.java#L97

rubbish:
@adriancole I get what you're saying. I'll make the ZipkinSender backfill instead of the span converter.

adriancole:
>
> @adriancole <https://github.com/adriancole> I get what you're saying.
> I'll make the ZipkinSender backfill instead of the span converter.
>
yep that or any other class in the zipkin module.

rubbish:
@yurishkuro @adriancole I've made the changes on the ZipkinSender to back fill annotations. How does it look?

jaeger-zipkin/src/main/java/com/uber/jaeger/senders/zipkin/ZipkinSender.java
```diff
@@ -161,4 +161,31 @@ public int close() throws SenderException {
     }
     return n;
   }
+
+  // serviceName/host is needed on annotations so zipkin can query
+  // see https://github.com/uber/jaeger-client-java/pull/75 for more info
+  Span backFillHostOnAnnotations(Span span) {
+    Endpoint host = null;
+    for (BinaryAnnotation binaryAnnotation : span.getBinary_annotations()) {
```
yurishkuro: core annotations are the more traditional source for Host/Service, they should be tried before the LC. Could also refactor the search into a method that returns as soon as host is found, instead of continuing looping.

rubbish: fixed in 8322bab206766137ec2bfbe7262946240b369656

jaeger-zipkin/src/main/java/com/uber/jaeger/senders/zipkin/ZipkinSender.java
```diff
@@ -161,4 +163,42 @@ public int close() throws SenderException {
...
+          annotation.setHost(host);
+        }
+      }
+    }
+
+    return span;
+  }
+
+  private Endpoint getLocalComponentEndpoint(Span span) {
+    for (BinaryAnnotation binaryAnnotation : span.getBinary_annotations()) {
```
yurishkuro: core annotations are in `span.annotations`, not binary.

jaeger-zipkin/src/main/java/com/uber/jaeger/senders/zipkin/ZipkinSender.java
```diff
@@ -161,4 +163,42 @@ public int close() throws SenderException {
...
+        if (!annotation.isSetHost()) {
+          annotation.setHost(host);
+        }
+      }
+    }
+
+    return span;
+  }
+
+  private Endpoint getLocalComponentEndpoint(Span span) {
```
yurishkuro: s/../getServiceEndpoint/ ?

PR #81, opened 2016-10-24
Add initialization logging

vprithvi:
- Add toString to Tracer and related objects
- Log when tracer is initialized using configuration

Fixes https://github.com/uber/jaeger-client-java/issues/79

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
@@ -145,9 +145,33 @@ public class Tracer implements io.opentracing.Tracer {
   /**
    * Shuts down the {@link Reporter} and {@link Sampler}
    */
-  public void close(){
+  public void close() {
     reporter.close();
     sampler.close();
+  }
+
+  @Override
+  public String toString() {
+    return "{\"_class\":\"Tracer\", "
+        + "\"VERSION\":"
+        + (VERSION == null ? "null" : "\"" + VERSION + "\"")
+        + ", "
+        + "\"serviceName\":"
+        + (serviceName == null ? "null" : "\"" + serviceName + "\"")
+        + ", "
+        + "\"reporter\":"
+        + (reporter == null ? "null" : "\"" + reporter + "\"")
+        + ", "
+        + "\"sampler\":"
+        + (sampler == null ? "null" : "\"" + sampler + "\"")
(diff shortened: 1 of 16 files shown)


black-adder:
do we want to cover these?

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -145,11 +145,35 @@ void reportSpan(Span span) {
...
    */
-  public void close(){
+  public void close() {
     reporter.close();
     sampler.close();
   }
 
+  @Override
+  public String toString() {
+    return "{\"_class\":\"Tracer\", "
```
badiib: is it worth using stringbuilder here

jaeger-core/src/main/java/com/uber/jaeger/reporters/InMemoryReporter.java
```diff
@@ -54,4 +55,12 @@ public void clear() {
       spans.clear();
     }
   }
+
+  @Override
+  public String toString() {
+    return "{\"_class\":\"InMemoryReporter\", "
```
badiib: same comment as above. same for all the toString methods below.

vprithvi:
@black-adder I'm not a fan of writing tests for naive toString methods because they are informational, and are auto-generated by the IDE. 
If we did not have a restriction on adding dependencies, we could use something like https://projectlombok.org/features/ToString.html which generates these methods during compile time.

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
vprithvi: It's not advantageous in this case because the compiler would convert the `+` operator to a `Stringbuilder` internally. See http://stackoverflow.com/a/11408526/481102 for a more detailed explanation

yurishkuro:
@vprithvi I am not against using lombok as long as it does not add runtime jar dependency

badiib:
LGTM then. It's best to have coverage imo, but I'm not a stickler for it.

vprithvi:
@yurishkuro Not sure about the coverage numbers, few of the older PRs don't have that either. I agree with your points about writing toString by hand. 
We can include lombok as a compile only dependency when using gradle > 2.12. 
I'm going to scrap this PR and work on using lombok instead.

PR #82, opened 2016-10-24
Add init logging to tracer

vprithvi:
- Add Lombok and toString methods to Tracer and related objects
- Move Tracer.VERSION to a field so that it's picked up by the
  autogenerated toString method
- TUDPTransport: writeButter -> WriteBuffer

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,10 @@ chmod a+x .git/hooks/pre-commit
 You can also setup the [IntelliJ plugin](https://plugins.jetbrains.com/plugin/8527)
 to reformat code from within your IDE
 
+### Lombok
+This project uses [Lombok](https://projectlombok.org/) to reduce boilerplate. You can setup
+ the [IntelliJ plugin](https://plugins.jetbrains.com/plugin/6317) to add IDE support. 
+
 ## License
   
   [The MIT License](LICENSE).
--- a/jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
@@ -49,11 +49,13 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Properties;
+import lombok.ToString;
 
+@ToString(exclude = {"registry", "clock", "metrics"})
 public class Tracer implements io.opentracing.Tracer {
-  public static final String VERSION = loadVersion();
   private final static Logger logger = LoggerFactory.getLogger(Tracer.class);
 
+  private final String version;
   private final String serviceName;
   private final Reporter reporter;
   private final Sampler sampler;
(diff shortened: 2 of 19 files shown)


README.md
```diff
@@ -136,6 +136,10 @@ chmod a+x .git/hooks/pre-commit
 You can also setup the [IntelliJ plugin](https://plugins.jetbrains.com/plugin/8527)
 to reformat code from within your IDE
 
+### Lombok
+This project uses [Lombok](https://projectlombok.org/) to reduce boilerplate. You can setup
+ the [IntelliJ plugin](https://plugins.jetbrains.com/plugin/6317) to add IDE support.
```
adriancole: ps a lot of spring projects explicitly stopped using lombok due to complaints about IDEs etc. That doesn't mean shouldn't be done, just saying that in the end many spring cloud projects undid lombok.

jaeger-core/src/main/java/com/uber/jaeger/reporters/protocols/TUDPTransport.java
```diff
@@ -27,18 +27,20 @@
...
+@ToString(exclude = {"writeBuffer"})
 public class TUDPTransport extends TTransport implements AutoCloseable {
   public static final int MAX_PACKET_SIZE = 65000;
 
   public final DatagramSocket socket;
   public byte[] receiveBuf;
   public int receiveOffSet = -1;
   public int receiveLength = 0;
-  public ByteBuffer writeButter;
+  public ByteBuffer writeBuffer;
```
yurishkuro: lol

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -49,11 +49,13 @@
 import java.util.Map;
 import java.util.Objects;
 import java.util.Properties;
+import lombok.ToString;
 
+@ToString(exclude = {"registry", "clock", "metrics"})
 public class Tracer implements io.opentracing.Tracer {
-  public static final String VERSION = loadVersion();
```
yurishkuro: what's the reason for this?  Version is static information.

vprithvi: This information depends on the runtime class path, and I didn't feel that it's sufficiently different from other things that depend on initialization like the `serviceName` and `ip`. I could change it back if you feel strongly about it.

README.md
vprithvi: I didn't know that, thanks for the perspective. Things do get gnarly in the IDE if one doesn't have the plugin installed (for IntelliJ), or the jar setup (for Eclipse); but once the one time setup is done, it's usually pretty smooth sailing. 
Do you recommend any alternatives?

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
yurishkuro: nah, keep it

adriancole:
>
> I didn't know that, thanks for the perspective. Things do get gnarly in
> the IDE if one doesn't have the plugin installed (for IntelliJ), or the jar
> setup (for Eclipse); but once the one time setup is done, it's usually
> pretty smooth sailing.
> Do you recommend any alternatives?
>
I'd say the major alternative is generating toString (via IDE or something
like AutoValue) only when required, or manually creating nice toString's
(summary as opposed to every field regardless of importance).

PR #83, opened 2016-10-26
Publish a shadow jar with thrift 0.9.2

vprithvi:
- publishes an additional jar that packages `org.apache.thrift:libthrift:0.9.2` under the classifier `thrift92`
- relocates `org.apache.thrift` to `org.shadow.apache.thrift92` in the shaded jar

Fixes https://github.com/uber/jaeger-client-java/issues/80

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,26 @@ For e.g, to depend on the core jaeger library, you'd include the following
 </dependency>
 ```
 
+###Thrift version conflicts
+Jaeger client uses `org.apache.thrift:libthrift:0.9.2`. If your project depends on a different
+version of `libthrift`, it is recommended that you use the shaded `jaeger-thrift` jar we publish
+which packages it's own `libthrift`.
+
+To depend on the shaded jar, add the following to your maven build.
+Note that this is only supported for a jaeger version >= 0.15.0
+```xml
+<dependencyManagement>
+  <dependencies>
+    <dependency>
+      <groupId>com.uber.jaeger</groupId>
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -39,6 +39,26 @@ For e.g, to depend on the core jaeger library, you'd include the following
...
+
+To depend on the shaded jar, add the following to your maven build.
+Note that this is only supported for a jaeger version >= 0.15.0
+```xml
+<dependencyManagement>
+  <dependencies>
+    <dependency>
+      <groupId>com.uber.jaeger</groupId>
+      <artifactId>jaeger-thrift</artifactId>
+      <classifier>thrift92</classifier>
```
vprithvi: The classifier adds in information on whether the shaded libthrift 0.9.2 is included or not. 
Omitting the classifier will result in using the default artifact that does not contain the bundled libthrift. (It simply contains the generated thrift files). 

In general, people never directly add jaeger-thrift because it is a transitive dependency of jaeger-core. 

See http://maven.apache.org/pom.html for more information on the classifier mechanism

> The classifier allows to distinguish artifacts that were built from the same POM but differ in their content. It is some optional and arbitrary string that - if present - is appended to the artifact name just after the version number.
> As a motivation for this element, consider for example a project that offers an artifact targeting JRE 1.5 but at the same time also an artifact that still supports JRE 1.4. The first artifact could be equipped with the classifier jdk15 and the second one with jdk14 such that clients can choose which one to use.
>

PR #91, opened 2016-11-17
Add Builder to TextMapCodec (https://github.com/uber/jaeger-client-java/issues/88)

### LINKED ISSUE #88: Anyway to provide a different SPAN_CONTEXT_KEY?
Is there a way to replace SPAN_CONTEXT_KEY to something like "x-trace-id" or "jaeger-trace-id" instead of ""uber-trace-id"?

The following TextMapCodec doesn't seem to be pluggable.

public class TextMapCodec implements Injector<TextMap>, Extractor<TextMap> { 
     /** Key used to store serialized span context representation */ 
     private static final String SPAN_CONTEXT_KEY = "uber-trace-id";

hzariv:
Add Builder to TextMapCodec - fixes #88

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/propagation/TextMapCodec.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/propagation/TextMapCodec.java
@@ -40,14 +40,14 @@ public class TextMapCodec implements Injector<TextMap>, Extractor<TextMap> {
 
   private static final PrefixedKeys keys = new PrefixedKeys();
 
-  private String contextKey = SPAN_CONTEXT_KEY;
+  private final String contextKey;
 
-  private String baggagePrefix = BAGGAGE_KEY_PREFIX;
+  private final String baggagePrefix;
 
-  private boolean urlEncoding;
+  private final boolean urlEncoding;
 
   public TextMapCodec(boolean urlEncoding) {
-    this.urlEncoding = urlEncoding;
+    this(builder().withURLEncoding(urlEncoding));
   }
 
   private TextMapCodec(Builder builder) {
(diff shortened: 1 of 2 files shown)


jaeger-core/src/main/java/com/uber/jaeger/propagation/TextMapCodec.java
```diff
@@ -40,16 +40,22 @@
 
   private static final PrefixedKeys keys = new PrefixedKeys();
 
-  private final String contextKey = SPAN_CONTEXT_KEY;
+  private String contextKey = SPAN_CONTEXT_KEY;
```
hzariv: Something like this would work:

````       
      private final String contextKey;

	  private final String baggagePrefix;

	  private final boolean urlEncoding;

	  public TextMapCodec(boolean urlEncoding) {
	    this.urlEncoding = urlEncoding;
	    this.contextKey = SPAN_CONTEXT_KEY;
	    this.baggagePrefix = BAGGAGE_KEY_PREFIX;
	  }
```

yurishkuro: yes, either that, or 
```java
public TextMapCodec(boolean urlEncoding) {
    this(Builder().withURLEncoding(urlEncoding));
}
```

yurishkuro: ideally defaulting should only happen in one place, in the builder

hzariv: fixed

yurishkuro:
@hzariv thanks for the change! 

@vprithvi  will be doing a new release soon.

PR #95, opened 2016-11-21
Automatically add span names for jersey instrumentation

vprithvi:
For the dropwizard jaeger feature implementation, generate the
default operation name for a given span based on the @Path()
annotations set on the calling resource.

### DIFF
--- a/jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/JerseyServerFilter.java
+++ b/jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/JerseyServerFilter.java
@@ -75,12 +75,8 @@ public class JerseyServerFilter extends ServerFilter {
           List<UriTemplate> templates = uriRoutingContext.getMatchedTemplates();
           String operationName = "";
           for (UriTemplate uriTemplate : templates) {
-            try {
-              String template = (String) normalizedTemplate.get(uriTemplate);
-              operationName = template + operationName;
-            } catch (IllegalAccessException e) {
-              log.error("No access to %s", normalizedTemplate, e);
-            }
+            String template = (String) normalizedTemplate.get(uriTemplate);
+            operationName = template + operationName;
           }
           return String.format("%s - %s", containerRequest.getMethod(), operationName);
         }
--- a/jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/JerseyServerFilter.java
+++ b/jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/JerseyServerFilter.java
@@ -82,9 +82,7 @@ public class JerseyServerFilter extends ServerFilter {
               log.error("No access to %s", normalizedTemplate, e);
             }
           }
-          String s = String.format("%s - %s", containerRequest.getMethod(), operationName);
-          System.out.println(s);
-          return s;
+          return String.format("%s - %s", containerRequest.getMethod(), operationName);
         }
       }
     } catch (Exception e) {
--- a/jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/JerseyServerFilter.java
+++ b/jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/JerseyServerFilter.java
@@ -99,11 +99,9 @@ public class JerseyServerFilter extends ServerFilter {
    * @return the operation name
    */
   private String getResourceMethodOperationName(ContainerRequestContext containerRequestContext) {
-    String operationName;Method method = resourceInfo.getResourceMethod();
-    operationName =
-        String.format("%s - %s:%s", containerRequestContext.getMethod(),
-                      method.getDeclaringClass().getCanonicalName(), method.getName());
-    return operationName;
+    Method method = resourceInfo.getResourceMethod();
+    return String.format("%s - %s:%s", containerRequestContext.getMethod(),
+                         method.getDeclaringClass().getCanonicalName(), method.getName());
   }
 
   /**
--- a/jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/JerseyServerFilter.java
+++ b/jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/JerseyServerFilter.java
@@ -46,6 +46,9 @@ import lombok.extern.slf4j.Slf4j;
  * A {@link ServerFilter} that provides a Jersey specific {@link #getOperationName} implementation.
  * When using this Filter with Jersey, the full @Path template hierarchy is used as the
  * operation name.
+ *
+ * WARNING: This class relies on an private implementation detail
+ * {@link UriTemplate#normalizedTemplate} to provide functionality.
  */
 @Slf4j
 @Provider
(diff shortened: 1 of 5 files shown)


jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/JerseyServerFilter.java
```diff
@@ -0,0 +1,89 @@
...
+            String template = (String) normalizedTemplate.get(uriTemplate);
+            operationName = template + operationName;
+          }
+          return String.format("%s - %s", containerRequest.getMethod(), operationName);
+        }
+      }
+    } catch (Exception e) {
+      log.error("Unable to get operation name", e);
+    }
+    return containerRequestContext.getMethod();
```
yurishkuro: maybe call `return super. getOperationName()` instead

jaeger-dropwizard/src/test/java/com/uber/jaeger/dropwizard/JerseyServerFilterTest.java
```diff
@@ -0,0 +1,97 @@
...
+import javax.ws.rs.core.Application;
+import javax.ws.rs.core.Response;
+
+public class JerseyServerFilterTest extends JerseyTest {
+
+  private InMemoryReporter reporter;
+
+  protected Application configure() {
+    Configuration config = new Configuration("world service", false,
+                                             new com.uber.jaeger.Configuration.SamplerConfiguration(ConstSampler.TYPE,
```
vprithvi: Done

jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/JerseyServerFilter.java
```diff
@@ -0,0 +1,89 @@
...
+  @Override
+  protected String getOperationName(ContainerRequestContext containerRequestContext) {
+    try {
+      if (normalizedTemplate != null && containerRequestContext instanceof ContainerRequest) {
+        ContainerRequest containerRequest = (ContainerRequest) containerRequestContext;
+        ExtendedUriInfo uriInfo = containerRequest.getUriInfo();
+        if (uriInfo instanceof UriRoutingContext) {
+          UriRoutingContext uriRoutingContext = (UriRoutingContext) uriInfo;
+          List<UriTemplate> templates = uriRoutingContext.getMatchedTemplates();
+          String operationName = "";
```
vprithvi: I used the `+`  for readability because we usually have only two strings to concatenate.

jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/JerseyServerFilter.java
```diff
@@ -0,0 +1,117 @@
...
+  /**
+   * Retrieves an operation name that contains the http method and value of the
+   * {@link Path} used by a Jersey server implementation.
+   * @return the operation name
+   */
+  @Override
+  protected String getOperationName(ContainerRequestContext containerRequestContext) {
+    CacheKey key  = CacheKey.of(resourceInfo.getResourceMethod(), containerRequestContext.getMethod());
+    if(methodToPathCache.containsKey(key)){
+      return methodToPathCache.get(key);
```
yurishkuro: nit: this requires computing hash code twice. I prefer `v = map.get(); if (v!=null) return v;`

jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/JerseyServerFilter.java
```diff
@@ -0,0 +1,117 @@
...
+    try {
+      if (normalizedTemplate != null && containerRequestContext instanceof ContainerRequest) {
+        ContainerRequest containerRequest = (ContainerRequest) containerRequestContext;
+        ExtendedUriInfo uriInfo = containerRequest.getUriInfo();
+        if (uriInfo instanceof UriRoutingContext) {
+          UriRoutingContext uriRoutingContext = (UriRoutingContext) uriInfo;
+          List<UriTemplate> templates = uriRoutingContext.getMatchedTemplates();
+          String path = "";
+          for (UriTemplate uriTemplate : templates) {
+            String template = (String) normalizedTemplate.get(uriTemplate);
```
yurishkuro: can we instead look up the annotation via reflection (first on method, then on class) and read the route from it? It would be cleaner than depending on a private implementation detail.

jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/JerseyServerFilter.java
```diff
@@ -0,0 +1,117 @@
...
+            path = template + path;
+          }
+          String operationName = String.format("%s - %s", containerRequest.getMethod(), path);
+          methodToPathCache.put(key, operationName);
```
yurishkuro: I suggest refactoring this method to have cache.put only once, e.g.
```
String operationName = cache.get();
if (operationName != null) { return operationName; }
try {
    compute operationName
} except {
    log
}
if (operationName == null) {
   operationName = getMethod();
}
cache.put(operationName)
return operationName;
```

jaeger-jaxrs2/src/main/java/com/uber/jaeger/filters/jaxrs2/ServerFilter.java
```diff
@@ -44,17 +46,21 @@
...
   @Override
   public void filter(ContainerRequestContext containerRequestContext) throws IOException {
+    //TODO: Add a @DisableTracing annotation and use the resourceInfo to check if it's applied
     try {
       Tracer.SpanBuilder builder =
           tracer
-              .buildSpan(containerRequestContext.getMethod())
+              .buildSpan(getOperationName(containerRequestContext))
               .withTag(Tags.SPAN_KIND.getKey(), Tags.SPAN_KIND_SERVER)
               //TODO(oibe) figure out how to get remote addr from grizzly HttpServletRequest
```
yurishkuro: please add an issue for this

jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/JerseyServerFilter.java
vprithvi: I looked into it earlier, it's easy to get the annotation on the method, and then on the class that the method is found. 
However, this isn't sufficient when people use [Jersey Subresources](https://jersey.java.net/documentation/latest/jaxrs-resources.html#d0e2496) which may be nested arbitrarily deep. Sub resources also don't form a type hierarchy, so there is no easy way to get them using public apis.

jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/JerseyServerFilter.java
vprithvi: Good catch. Also, because it is a ConcurrentHashMap, there is a possibility that something removed the key after the `containsKey` operation. 
I'll change this.

jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/JerseyServerFilter.java
vprithvi: I'll change this, but the benefit isn't obvious to me, could you elaborate?

jaeger-jaxrs2/src/main/java/com/uber/jaeger/filters/jaxrs2/ServerFilter.java
vprithvi: Created https://github.com/uber/jaeger-client-java/issues/97

jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/JerseyServerFilter.java
```diff
@@ -0,0 +1,117 @@
...
+
+  private Map<CacheKey, String> methodToPathCache = new ConcurrentHashMap<>();
+  private Field normalizedTemplate;
+
+  public JerseyServerFilter(Tracer tracer,
+                            TraceContext traceContext) {
+    super(tracer, traceContext);
+
+    try {
+      normalizedTemplate = UriTemplate.class.getDeclaredField("normalizedTemplate");
```
yurishkuro: add a WARNING that this is using undocumented private implementation detail.

jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/JerseyServerFilter.java
```diff
@@ -0,0 +1,146 @@
...
+    methodToPathCache.put(key, operationName);
+    return operationName;
+  }
+
+  /**
+   * Retrieves an operation name that uses the resource method name of the matched jersey resource
+   * @return the operation name
+   */
+  private String getResourceMethodOperationName(ContainerRequestContext containerRequestContext) {
+    Method method = resourceInfo.getResourceMethod();
```
yurishkuro: here's what I don't understand. Where does the `resourceInfo` property come from? It looks like an inherited member variable, which implies there's going to be a new instance of this filter class for every method, if not for every http call, which makes the cache non-reusable.

vprithvi: The resourceInfo is an proxy of the `ResourceInfo` class that is injected by Jersey's HK2 injection framework.  It is inherited from [here](https://github.com/uber/jaeger-client-java/pull/95/files/eca86e5c18a5fe2d015671fe536e2c78b99d1912#diff-2fb9d84565a5970cc3700f4658c9db9dR49)
The proxy is request scoped and can handle multiple requests simultaneously. The jersey docs are light on details, but allude to this [here](https://jersey.java.net/nonav/documentation/latest/user-guide.html#d0e2713)

vprithvi: I added a test [here](https://github.com/uber/jaeger-client-java/pull/95/commits/5b8913ecac0596590f09a642704384048f323693#diff-52f72d513043802c8172cd3d15ac891cR106) that demonstrates that the cache holds more than one value.

jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/JerseyServerFilter.java
```diff
@@ -0,0 +1,148 @@
...
+    return String.format("%s - %s:%s", containerRequestContext.getMethod(),
```
yurishkuro: don't use spaces in the operation name

jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/JerseyServerFilter.java
```diff
@@ -0,0 +1,148 @@
...
+        ExtendedUriInfo uriInfo = containerRequest.getUriInfo();
+        if (uriInfo instanceof UriRoutingContext) {
+          UriRoutingContext uriRoutingContext = (UriRoutingContext) uriInfo;
+          List<UriTemplate> templates = uriRoutingContext.getMatchedTemplates();
+          String path = "";
+          for (UriTemplate uriTemplate : templates) {
+            String template = (String) normalizedTemplate.get(uriTemplate);
+            path = template + path;
+          }
+          operationName = String.format("%s - %s", containerRequest.getMethod(), path);
```
yurishkuro: don't use spaces in the operation name

PR #100, opened 2016-11-30
Tighten Span#isRPC check

vprithvi:
- Span#isRPC is set specifically for server or client spans. Other values for `span.kind` don't affect `isRPC`.


vprithvi:
@yurishkuro Not really, previously we set `isRPC` to true simply when a `span.kind` was present. Now we tighten the check to only set it to true when span.kind is `server` or `client`. I updated the commit message to clarify this.

yurishkuro:
Ah, ok, good catch.

yurishkuro:
But the PR description is still confusing - you're changing functionality, being "more readable" is much less irrelevant.

vprithvi:
Agree, shortened the message

PR #101, opened 2016-12-01
Make Tracer#isRPCServer robust

vprithvi:
- Instead of comparing that the specific Tags.SPAN_KIND_SERVER object
  is used, simply check whether the field contains "server"

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
@@ -158,7 +158,8 @@ public class Tracer implements io.opentracing.Tracer {
     sampler.close();
   }
 
-  private class SpanBuilder implements io.opentracing.Tracer.SpanBuilder {
+  //Visible for testing
+  class SpanBuilder implements io.opentracing.Tracer.SpanBuilder {
 
     private String operationName = null;
     private long startTimeMicroseconds;
@@ -259,7 +260,8 @@ public class Tracer implements io.opentracing.Tracer {
           null);
     }
 
-    private boolean isRPCServer() {
+    //Visible for testing
+    boolean isRPCServer() {
       return Tags.SPAN_KIND_SERVER.equals(tags.get(Tags.SPAN_KIND.getKey()));
     }
 
--- a/jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
@@ -260,7 +260,7 @@ public class Tracer implements io.opentracing.Tracer {
     }
 
     private boolean isRPCServer() {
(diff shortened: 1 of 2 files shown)


vprithvi:
@yurishkuro done

jaeger-core/src/test/java/com/uber/jaeger/TracerTest.java
```diff
@@ -136,6 +138,22 @@ public void testServiceNameNotEmptyNull() {
...
+    Tracer.SpanBuilder spanBuilder = (Tracer.SpanBuilder) tracer.buildSpan("ndnd");
+    spanBuilder.withTag(Tags.SPAN_KIND.getKey(), "server");
+
+    assertTrue(spanBuilder.isRPCServer());
+  }
+
+  @Test
+  public void testBuilderIsNotServerRPC() {
+    Tracer.SpanBuilder spanBuilder = (Tracer.SpanBuilder) tracer.buildSpan("Lrrr");
+    spanBuilder.withTag(Tags.SPAN_KIND.getKey(), "peachy");
```
adriancole: well good to know some people's servers are peachy.. wait.. I guess they aren't servers if they are peachy? :)

vprithvi: peachy spans make not a peachy server!

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -158,7 +158,8 @@ public void close() {
     sampler.close();
   }
 
-  private class SpanBuilder implements io.opentracing.Tracer.SpanBuilder {
+  //Visible for testing
```
yurishkuro: why can't we test this via methods on the resulting Span?

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -259,8 +260,9 @@ private SpanContext createChildContext() {
           null);
     }
 
-    private boolean isRPCServer() {
-      return tags.get(Tags.SPAN_KIND.getKey()) == Tags.SPAN_KIND_SERVER;
+    //Visible for testing
```
yurishkuro: why can't we test this via methods on the resulting Span?

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
vprithvi: Because it is no longer a unit test. `SpanBuilder#isRPCServer` is implemented in `SpanBuilder` and not in `Span`. 

`Span` has it's own separate implementation of when it's representing an rpc or client span. I feel that we should document and refactor these classes at some point.

PR #102, opened 2016-12-01
Fix client/server span classification

vprithvi:
The `Span` object has custom logic to set `Span#isRpc` and `Span#isClient` based
on the tags passed into `Span#setTag`.
Creating a `Span` using `SpanBuilder` bypasses this logic causing `Span#isRPC` to be
false. This causes ThriftSpanConverter to use `lc` annotations instead of `sr/ss`

This commit sanitizes input from `SpanBuilder` so that setting tags on the
`SpanBuilder` results in the same behavior as setting a tag on the `Span`.

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/Span.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Span.java
@@ -75,7 +75,9 @@ public class Span implements io.opentracing.Span {
   }
 
   public long getDuration() {
-    return durationMicroseconds;
+    synchronized (this) {
+      return durationMicroseconds;
+    }
   }
 
   public Tracer getTracer() {
--- a/jaeger-core/src/main/java/com/uber/jaeger/Span.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Span.java
@@ -185,17 +185,17 @@ public class Span implements io.opentracing.Span {
   }
 
   @Override
-  public Span setTag(String key, String value) {
+  public synchronized Span setTag(String key, String value) {
     return setTagAsObject(key, value);
   }
 
   @Override
-  public Span setTag(String key, boolean value) {
+  public synchronized Span setTag(String key, boolean value) {
     return setTagAsObject(key, value);
   }
 
   @Override
-  public Span setTag(String key, Number value) {
+  public synchronized Span setTag(String key, Number value) {
     return setTagAsObject(key, value);
   }
 
@@ -241,7 +241,6 @@ public class Span implements io.opentracing.Span {
    * See {@link #handleSpecialTag(String, Object)}
    */
   private Span setTagAsObject(String key, Object value) {
-    synchronized (this) {
       if (key.equals(Tags.SAMPLING_PRIORITY.getKey()) && (value instanceof Number)) {
         int priority = ((Number) value).intValue();
         byte newFlags;
@@ -259,7 +258,6 @@ public class Span implements io.opentracing.Span {
           tags.put(key, value);
         }
       }
-    }
 
     return this;
   }
--- a/jaeger-core/src/main/java/com/uber/jaeger/Span.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Span.java
@@ -59,7 +59,7 @@ public class Span implements io.opentracing.Span {
     this.startTimeMicroseconds = startTimeMicroseconds;
     this.startTimeNanoTicks = startTimeNanoTicks;
     this.computeDurationViaNanoTicks = computeDurationViaNanoTicks;
-    this.tags = tags;
+    sanitizeAndSetTags(tags);
   }
 
   public String getLocalComponent() {
@@ -199,6 +199,10 @@ public class Span implements io.opentracing.Span {
     return setTagAsObject(key, value);
   }
 
+  /**
+   * Sets various fields on the {@link Span} when certain {@link Tags} are encountered
+   * @return true iff a special tag is handled
+   */
   private boolean handleSpecialTag(String key, Object value) {
     // TODO use a map of handlers for special tags, instead of if/then
     if (key.equals(Tags.COMPONENT.getKey()) && value instanceof String) {
@@ -231,6 +235,11 @@ public class Span implements io.opentracing.Span {
     return false;
   }
 
+  /**
+   * Set a key:value tag on a span ONLY when the tag isn't special.
+   *
+   * See {@link #handleSpecialTag(String, Object)}
+   */
   private Span setTagAsObject(String key, Object value) {
     synchronized (this) {
       if (key.equals(Tags.SAMPLING_PRIORITY.getKey()) && (value instanceof Number)) {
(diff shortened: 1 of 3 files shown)


jaeger-core/src/test/java/com/uber/jaeger/TracerTest.java
```diff
@@ -66,6 +68,26 @@ public void testBuildSpan() {
   }
 
   @Test
+  public void testBuildServerSpan() {
+    Span span = (com.uber.jaeger.Span) tracer.buildSpan("flexo")
+                                             .withTag(Tags.SPAN_KIND.getKey(),
+                                                      Tags.SPAN_KIND_SERVER).start();
```
yurishkuro: odd formatting

jaeger-core/src/main/java/com/uber/jaeger/Span.java
```diff
@@ -259,6 +264,16 @@ private Span setTagAsObject(String key, Object value) {
     return this;
   }
 
+  /**
+   * uses {@link #setTagAsObject} to only insert tags that don't have special significance
+   */
+  private void sanitizeAndSetTags(Map<String, Object> tags) {
+    this.tags = new HashMap<>();
+    for (Map.Entry<String, Object> tag : tags.entrySet()) {
+      setTagAsObject(tag.getKey(), tag.getValue());
```
badiib: a little worried about performance here - setTagAsObject locks every time. if sanitizeAndSetTags is only called in the constructor, I don't think we need to worry about syncing.

jaeger-core/src/test/java/com/uber/jaeger/TracerTest.java
```diff
@@ -66,6 +68,26 @@ public void testBuildSpan() {
...
+
+    assertTrue(span.isRPC());
+    assertFalse(span.isRPCClient());
+  }
+
+  @Test
+  public void testBuildClientSpan() {
```
yurishkuro: this test seems identical to the one above

jaeger-core/src/main/java/com/uber/jaeger/Span.java
```diff
@@ -259,6 +264,16 @@ private Span setTagAsObject(String key, Object value) {
     return this;
   }
 
+  /**
+   * uses {@link #setTagAsObject} to only insert tags that don't have special significance
+   */
+  private void sanitizeAndSetTags(Map<String, Object> tags) {
```
yurishkuro: "sanitize" implies validation and potentially discarding or replacing tags, we do not do that here.

jaeger-core/src/main/java/com/uber/jaeger/Span.java
vprithvi: This is only called in the constructor

jaeger-core/src/test/java/com/uber/jaeger/TracerTest.java
vprithvi: Good catch, I'll update

jaeger-core/src/main/java/com/uber/jaeger/Span.java
vprithvi: That incorrect, we check and discard tags. 
For e.g., one can never set a `span.kind:server` on the `Span` object. It is always discarded after setting the `isRPC` field to true. Other tags that are discarded are PEER_PORT, PEER_HOST, PEER_SERVICE, etc.

jaeger-core/src/test/java/com/uber/jaeger/TracerTest.java
vprithvi: Will update

jaeger-core/src/main/java/com/uber/jaeger/Span.java
vprithvi: Agree, updated.

jaeger-core/src/main/java/com/uber/jaeger/Span.java
yurishkuro: > one can never set a span.kind:server on the Span object. It is always discarded after setting the isRPC field to true.

This method is called from the constructor, so "after" is not applicable. And we do not discard any tags; they may not be stored as tags explicitly, but they are certainly "stored" one way or another. Nothing is being "sanitized" here.

I would just inline this code in the constructor, since it's never reused.

jaeger-core/src/test/java/com/uber/jaeger/TracerTest.java
```diff
@@ -66,6 +68,26 @@ public void testBuildSpan() {
   }
 
   @Test
+  public void testBuildServerSpan() {
+    Span span = (com.uber.jaeger.Span) tracer.buildSpan("flexo")
+                                             .withTag(Tags.SPAN_KIND.getKey(), Tags.SPAN_KIND_SERVER)
```
yurishkuro: can you put `tracer.buildSpan("flexo")` on the 2nd line to avoid the insane indenting by the formatter? Or at least put `.buildSpan()` on the second line.

jaeger-core/src/main/java/com/uber/jaeger/Span.java
yurishkuro: Also, once we upgrade this to jaeger.thrift, all this special handling goes away.

vprithvi: >And we do not discard any tags; they may not be stored as tags explicitly, but they are certainly "stored" one way or another. Nothing is being "sanitized" here.

I agree that the data contained in the tags are stored one way or the other. 
The tags themselves are discarded and cannot be retrieved by `Span#getTags`. Generally, if something is not discarded, I expect something set using the setter to be retrievable using a getter. 

In the class, setting tags using `#setTag` does not follow this behavior. 

At any rate, I don't see the point of arguing about semantics here. I'll inline this code.

vprithvi: >And we do not discard any tags; they may not be stored as tags explicitly, but they are certainly "stored" one way or another. Nothing is being "sanitized" here.

I agree that the data contained in the tags are stored one way or the other. 
The tags themselves are discarded and cannot be retrieved by `Span#getTags`. Generally, if something is not discarded, I expect something set using the setter to be retrievable using a getter. 

In the class, setting tags using `#setTag` does not follow this behavior. 

At any rate, I don't see the point of arguing about semantics here. I'll inline this code.

jaeger-core/src/test/java/com/uber/jaeger/TracerTest.java
vprithvi: This is what it looks like on my IDE, I think it is reasonable indenting. Putting `.buildSpan` on the next line would simply move the block to the left by a few characters. Could you elaborate on why the current indenting is insane? Perhaps github is rendering things differently?
image

vprithvi:
Accidentally staged more hunks than I meant to, undoing.

PR #103, opened 2016-12-05
Define json models for SamplingStrategyResponse and use them

vprithvi:
- Use our own json models instead of depending on thrift generated classes.
  This will help up when we move to manual thrift generation
- Use Google's gson instead of jackson-databind
- Add tests for HttpSamplingManager

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,7 +14,6 @@ ext.guavaVersion = '18.0'
 ext.apacheThriftVersion = '0.9.2'
 ext.jerseyVersion = '2.22.2'
 ext.slf4jVersion = '1.7.16'
-ext.jacksonVersion = '2.7.4'
 ext.apacheHttpComponentsVersion = '4.1.2'
 ext.gsonVersion = '2.8.0'
 
--- a/build.gradle
+++ b/build.gradle
@@ -16,6 +16,7 @@ ext.jerseyVersion = '2.22.2'
 ext.slf4jVersion = '1.7.16'
 ext.jacksonVersion = '2.7.4'
 ext.apacheHttpComponentsVersion = '4.1.2'
+ext.gsonVersion = '2.8.0'
 
 ext.junitVersion = '4.12'
 ext.mockitoVersion = '2.0.2-beta'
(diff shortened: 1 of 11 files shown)


jaeger-core/src/main/java/com/uber/jaeger/samplers/http/ProbabilisticSamplingStrategy.java
```diff
@@ -0,0 +1,29 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package com.uber.jaeger.samplers.http;
+
+import lombok.Value;
```
vprithvi: @Value automatically sets the class to final, and fields to be private final
See https://projectlombok.org/features/Value.html

black-adder:
LGTM although I'm not a fan of "samplers/http/" as a directory name for json models.

build.gradle
```diff
@@ -16,6 +16,7 @@ ext.jerseyVersion = '2.22.2'
 ext.slf4jVersion = '1.7.16'
 ext.jacksonVersion = '2.7.4'
```
yurishkuro: do we still need this then?

vprithvi: Yes, it's used by crossdock

yurishkuro: if *only* by crossdock, let's move it there, so it's less confusing at the top level.

vprithvi: Yeah, it seems crossdock has it's own version. It's also in use by jaeger-jaxrs2, but it seems that it's unused. 
I deleted it.

yurishkuro:
LGTM

PR #109, opened 2016-12-13
Support per operation Sampling

vprithvi:
- Add a PerOperationSampler that maintains a sampling probability for every operation


vprithvi:
I thought we agreed that dependent PRs were painful.

Are you recommending that I rebase some of the initial commits (Adding sampling tags) to a different branch, create a PR and wait for it to be approved, and then move the remaining commits over?

yurishkuro:
They are not as painful as expecting a quality review from 600+ lines diff.
Yes, create another branch that only has sampling tags change, and make a PR for it. It should get approved a lot quicker than the change with the actual adaptive logic.

vprithvi:
sgtm, will do

PR #110, opened 2016-12-13
Flatten directories for coveralls

vprithvi:
Coveralls might require directories to be flattened to generate accurate coverage reports.


vprithvi:
whoa @coveralls...
Still seems to be broken in that https://coveralls.io/builds/9267814 shows 465 source files but only lists 3...

PR #111, opened 2016-12-14
Refactor Sampler interface

vprithvi:
In preparation for per operation sampling, 

- `Sampler#isSampled` takes in an operation name
- `Sampler#isSampled` returns a `SamplingStatus` which contains tags
- `Sampler#getTags` is removed

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/samplers/ConstSampler.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/samplers/ConstSampler.java
@@ -51,13 +51,8 @@ public class ConstSampler implements Sampler {
    * @return A boolean that says whether to sample.
    */
   @Override
-  public SamplingStatus getSamplingStatus(String operation, long id) {
+  public SamplingStatus sample(String operation, long id) {
     return SamplingStatus.of(decision, tags);
-  }
-
-  // Visible for testing
-  Map<String, Object> getTags() {
-    return this.tags;
   }
 
   @Override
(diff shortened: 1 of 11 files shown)


jaeger-core/src/main/java/com/uber/jaeger/samplers/ConstSampler.java
```diff
@@ -51,12 +51,12 @@ public ConstSampler(boolean decision) {
...
-  public boolean isSampled(long id) {
-    return decision;
+  public SamplingStatus getSamplingStatus(String operation, long id) {
+    return SamplingStatus.of(decision, tags);
   }
 
-  @Override
-  public Map<String, Object> getTags() {
+  // Visible for testing
+  Map<String, Object> getTags() {
```
yurishkuro: why do we still need this?  Tests can get the tags via getSamplingStatus

jaeger-core/src/main/java/com/uber/jaeger/samplers/SamplingStatus.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+package com.uber.jaeger.samplers;
+
+import java.util.Map;
+import lombok.Value;
+
+@Value(staticConstructor = "of")
+public class SamplingStatus {
+  boolean isSampled;
+  Map<String, Object> tags;
```
yurishkuro: shouldn't these be private final?

vprithvi: Yeah, they are, because we are using the `@value` annotation. See https://projectlombok.org/features/Value.html for docs.

PR #112, opened 2016-12-14
Support per operation sampling

vprithvi:
Add a PerOperationSampler that maintains a sampling probability for every operation

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/samplers/GuaranteedThroughputSampler.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/samplers/GuaranteedThroughputSampler.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright (c) 2016, Uber Technologies, Inc
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package com.uber.jaeger.samplers;
+
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
+
+/**
+ * {@link GuaranteedThroughputSampler} is a {@link Sampler} that guarantees a throughput by using
+ * a {@link ProbabilisticSampler} and {@link RateLimitingSampler} in tandem.
+ *
+ * The RateLimitingSampler is used to establish a lowerBound so that every operation is sampled
+ * at least once in the time interval defined by the lowerBound.
+ */
+@EqualsAndHashCode
+@ToString
+public class GuaranteedThroughputSampler implements Sampler {
+  private final ProbabilisticSampler probabilisticSampler;
+  private final RateLimitingSampler lowerBoundSampler;
+
+  public GuaranteedThroughputSampler(double samplingRate, double lowerBound) {
+    this(new ProbabilisticSampler(samplingRate), new RateLimitingSampler(lowerBound));
+  }
+
+  // Visible for testing
+  GuaranteedThroughputSampler(ProbabilisticSampler probabilisticSampler,
+                              RateLimitingSampler rateLimitingSampler) {
+    this.probabilisticSampler = probabilisticSampler;
+    this.lowerBoundSampler = rateLimitingSampler;
+  }
+
+
+  /**
+   * Calls {@link Sampler#sample(String, long)} (String, long)} on both samplers, returning true for
+   * {@link SamplingStatus#isSampled} if either samplers set #isSampled to true.
+   * The tags corresponding to the sampler that returned true are set on {@link SamplingStatus#tags}
+   * If both samplers return true, tags for {@link ProbabilisticSampler} is given priority.
+   * @param operation The operation name, which is ignored by this sampler
+   * @param id The traceId on the span
+   */
+  @Override
+  public SamplingStatus sample(String operation, long id) {
+    SamplingStatus probabilisticSamplingStatus = probabilisticSampler.sample(operation, id);
+    SamplingStatus lowerBoundSamplingStatus = lowerBoundSampler.sample(operation, id);
+
+    if (probabilisticSamplingStatus.isSampled()) {
+      return probabilisticSamplingStatus;
+    }
+
+    return lowerBoundSamplingStatus;
+  }
+
+  public void update(double samplingRate, double lowerBound){
+    probabilisticSampler.update(samplingRate);
+    lowerBoundSampler.update(lowerBound);
+  }
+
+  @Override
+  public void close() {
+    probabilisticSampler.close();
+    lowerBoundSampler.close();
+  }
+}
--- a/jaeger-core/src/main/java/com/uber/jaeger/samplers/GuaranteedThroughputSampler.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/samplers/GuaranteedThroughputSampler.java
@@ -21,7 +21,10 @@
  */
 package com.uber.jaeger.samplers;
 
-import lombok.Value;
+import lombok.AccessLevel;
+import lombok.AllArgsConstructor;
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
 
 /**
  * {@link GuaranteedThroughputSampler} is a {@link Sampler} that guarantees a throughput by using
@@ -30,22 +33,16 @@ import lombok.Value;
  * The RateLimitingSampler is used to establish a lowerBound so that every operation is sampled
  * at least once in the time interval defined by the lowerBound.
  */
-@Value
-public class GuaranteedThroughputSampler implements Sampler {
-  ProbabilisticSampler probabilisticSampler;
-  RateLimitingSampler lowerBoundSampler;
+@ToString
+@EqualsAndHashCode
+@AllArgsConstructor (access = AccessLevel.PACKAGE) // Visible for testing
+public final class GuaranteedThroughputSampler implements Sampler {
+  private final ProbabilisticSampler probabilisticSampler;
+  private final RateLimitingSampler lowerBoundSampler;
 
   GuaranteedThroughputSampler(double samplingRate, double lowerBound) {
     this(new ProbabilisticSampler(samplingRate), new RateLimitingSampler(lowerBound));
   }
-
-  // Visible for testing
-  GuaranteedThroughputSampler(ProbabilisticSampler probabilisticSampler,
-                              RateLimitingSampler rateLimitingSampler) {
-    this.probabilisticSampler = probabilisticSampler;
-    this.lowerBoundSampler = rateLimitingSampler;
-  }
-
 
   /**
    * Calls {@link Sampler#sample(String, long)} (String, long)} on both samplers, returning true for
--- a/jaeger-core/src/main/java/com/uber/jaeger/samplers/GuaranteedThroughputSampler.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/samplers/GuaranteedThroughputSampler.java
@@ -69,9 +69,11 @@ public class GuaranteedThroughputSampler implements Sampler {
     return lowerBoundSamplingStatus;
   }
 
-  public void update(double samplingRate, double lowerBound){
-    probabilisticSampler.update(samplingRate);
-    lowerBoundSampler.update(lowerBound);
+  public boolean update(double samplingRate, double lowerBound){
+    boolean isUpdated;
+    isUpdated = probabilisticSampler.update(samplingRate);
+    isUpdated |= lowerBoundSampler.update(lowerBound);
+    return isUpdated;
   }
 
   @Override
--- a/jaeger-core/src/main/java/com/uber/jaeger/samplers/GuaranteedThroughputSampler.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/samplers/GuaranteedThroughputSampler.java
@@ -21,8 +21,7 @@
  */
 package com.uber.jaeger.samplers;
 
-import lombok.EqualsAndHashCode;
-import lombok.ToString;
+import lombok.Value;
 
 /**
  * {@link GuaranteedThroughputSampler} is a {@link Sampler} that guarantees a throughput by using
@@ -31,13 +30,12 @@ import lombok.ToString;
  * The RateLimitingSampler is used to establish a lowerBound so that every operation is sampled
  * at least once in the time interval defined by the lowerBound.
  */
-@EqualsAndHashCode
-@ToString
+@Value
 public class GuaranteedThroughputSampler implements Sampler {
-  private final ProbabilisticSampler probabilisticSampler;
-  private final RateLimitingSampler lowerBoundSampler;
+  ProbabilisticSampler probabilisticSampler;
+  RateLimitingSampler lowerBoundSampler;
 
-  public GuaranteedThroughputSampler(double samplingRate, double lowerBound) {
+  GuaranteedThroughputSampler(double samplingRate, double lowerBound) {
     this(new ProbabilisticSampler(samplingRate), new RateLimitingSampler(lowerBound));
   }
 
@@ -67,13 +65,6 @@ public class GuaranteedThroughputSampler implements Sampler {
     }
 
     return lowerBoundSamplingStatus;
-  }
-
-  public boolean update(double samplingRate, double lowerBound){
-    boolean isUpdated;
-    isUpdated = probabilisticSampler.update(samplingRate);
-    isUpdated |= lowerBoundSampler.update(lowerBound);
-    return isUpdated;
   }
 
   @Override
--- a/jaeger-core/src/main/java/com/uber/jaeger/samplers/PerOperationSampler.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/samplers/PerOperationSampler.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright (c) 2016, Uber Technologies, Inc
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package com.uber.jaeger.samplers;
+
+import com.uber.jaeger.samplers.http.OperationSamplingParameters;
+import com.uber.jaeger.samplers.http.PerOperationSamplingParameters;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * Computes {@link #sample(String, long)} using the name of the operation, and maintains a specific
+ * {@link GuaranteedThroughputSampler} instance for each operation.
+ */
+@Slf4j
+public class PerOperationSampler implements Sampler {
+  private final ConcurrentHashMap<String, GuaranteedThroughputSampler> operationNameToSampler;
+  private final int maxOperations;
+  private final ProbabilisticSampler defaultSampler;
+
+  public PerOperationSampler(int maxOperations, OperationSamplingParameters strategies) {
+    this(maxOperations,
+         new ConcurrentHashMap<String, GuaranteedThroughputSampler>(),
+         new ProbabilisticSampler(strategies.getDefaultSamplingProbability()));
+    update(strategies);
+  }
+
+  // Visible for testing
+  PerOperationSampler(int maxOperations,
+                      ConcurrentHashMap<String, GuaranteedThroughputSampler> operationNameToSampler,
+                      ProbabilisticSampler defaultSampler) {
+
+    this.operationNameToSampler = operationNameToSampler;
+    this.maxOperations = maxOperations;
+    this.defaultSampler = defaultSampler;
+  }
+
+  // Note that this isn't synchronized because ProbabilisticSampler#update and
+  // GuaranteedThroughputSampler#update are synchronized
+  public void update(OperationSamplingParameters strategies){
+    double lowerBound = strategies.getDefaultLowerBoundTracesPerSecond();
+    double defaultSamplingRate = strategies.getDefaultSamplingProbability();
+
+    defaultSampler.update(defaultSamplingRate);
+
+    for (PerOperationSamplingParameters strategy : strategies.getPerOperationStrategies()) {
+      String operation = strategy.getOperation();
+      double samplingRate = strategy.getProbabilisticSampling().getSamplingRate();
+
+      GuaranteedThroughputSampler guaranteedThroughputSampler = operationNameToSampler.get(operation);
+
+      if (guaranteedThroughputSampler != null) {
+        guaranteedThroughputSampler.update(samplingRate, lowerBound);
+      } else if (operationNameToSampler.size() < maxOperations) {
+        operationNameToSampler.putIfAbsent(operation,
+                                           new GuaranteedThroughputSampler(samplingRate,
+                                                                           lowerBound));
+      } else {
+        log.info("Exceeded the maximum number of operations({}) for per operations sampling",
+                 maxOperations);
+      }
+    }
+
+  }
+
+  @Override
+  public SamplingStatus sample(String operation, long id) {
+    GuaranteedThroughputSampler sampler = operationNameToSampler.get(operation);
+    if (sampler != null) {
+      return sampler.sample(operation, id);
+    }
+
+    return defaultSampler.sample(operation, id);
+  }
+
+  @Override
+  public void close() {
+    defaultSampler.close();
+    for (Map.Entry<String, GuaranteedThroughputSampler> sampler : operationNameToSampler.entrySet()) {
+      sampler.getValue().close();
+    }
+
+  }
+}
--- a/jaeger-core/src/main/java/com/uber/jaeger/samplers/PerOperationSampler.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/samplers/PerOperationSampler.java
@@ -26,6 +26,9 @@ import com.uber.jaeger.samplers.http.PerOperationSamplingParameters;
 
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import lombok.AccessLevel;
+import lombok.Data;
+import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 
 /**
@@ -33,10 +36,12 @@ import lombok.extern.slf4j.Slf4j;
  * {@link GuaranteedThroughputSampler} instance for each operation.
  */
 @Slf4j
+@Data
+@Getter(AccessLevel.NONE)
 public class PerOperationSampler implements Sampler {
   private final ConcurrentHashMap<String, GuaranteedThroughputSampler> operationNameToSampler;
   private final int maxOperations;
-  private final ProbabilisticSampler defaultSampler;
+  private volatile ProbabilisticSampler defaultSampler;
 
   public PerOperationSampler(int maxOperations, OperationSamplingParameters strategies) {
     this(maxOperations,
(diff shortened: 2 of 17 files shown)


vprithvi:
I think code coverage is safe to ignore for now. 

image

jaeger-core/src/main/java/com/uber/jaeger/samplers/PerOperationSampler.java
```diff
@@ -0,0 +1,104 @@
...
+      } else if (operationNameToSampler.size() < maxOperations) {
+        operationNameToSampler.putIfAbsent(operation,
+                                           new GuaranteedThroughputSampler(samplingRate,
+                                                                           lowerBound));
+      } else {
+        log.info("Exceeded the maximum number of operations({}) for per operations sampling",
+                 maxOperations);
+      }
+    }
+
```
black-adder: not a fan of hanging newlines

jaeger-core/src/main/java/com/uber/jaeger/samplers/ProbabilisticSampler.java
```diff
@@ -33,25 +33,38 @@
...
-    this.positiveSamplingBoundary = (long) (((1L << 63) - 1) * samplingRate);
-    this.negativeSamplingBoundary = (long) ((1L << 63) * samplingRate);
 
-    Map<String, Object> tags = new HashMap<>();
+    HashMap<String, Object> tags = new HashMap<>();
     tags.put(Constants.SAMPLER_TYPE_TAG_KEY, TYPE);
     tags.put(Constants.SAMPLER_PARAM_TAG_KEY, samplingRate);
     this.tags = Collections.unmodifiableMap(tags);
+
+    positiveSamplingBoundary = (long) (((1L << 63) - 1) * samplingRate);
```
black-adder: was there a reason for this change? I actually prefer it with "this."

jaeger-core/src/main/java/com/uber/jaeger/samplers/RemoteControlledSampler.java
```diff
@@ -104,6 +105,11 @@ private void updateSampler() {
 
   private Sampler extractSampler(SamplingStrategyResponse response)
       throws UnknownSamplingStrategyException {
+    if(response.getOperationSampling() != null) {
```
black-adder: space between if and (

it's inconsistent with the rest of the code

jaeger-core/src/main/java/com/uber/jaeger/samplers/RemoteControlledSampler.java
```diff
@@ -104,6 +105,11 @@ private void updateSampler() {
...
+      return new PerOperationSampler(maxOperations, response.getOperationSampling());
```
black-adder: it looks like we're recreating the whole PerOperationSampler every minute which is not what we want to do. We want to call the Update function instead if the PerOperationSampler already exists.

vprithvi: Good catch

jaeger-core/src/main/java/com/uber/jaeger/samplers/ProbabilisticSampler.java
vprithvi: Because it is unnecessary cruft in this use case; it's usually use to disambiguate scope, which doesn't need to happen here.

jaeger-core/src/main/java/com/uber/jaeger/samplers/ProbabilisticSampler.java
```diff
@@ -33,25 +33,38 @@
...
 
-  private final long positiveSamplingBoundary;
-  private final long negativeSamplingBoundary;
-  private final double samplingRate;
-  private final Map<String, Object> tags;
+  private final double DELTA = 0.00001;
+  private long positiveSamplingBoundary;
+  private long negativeSamplingBoundary;
+  private double samplingRate;
+  private Map<String, Object> tags;
```
yurishkuro: I prefer to keep these final. You can change the update() method to return a new instance.

vprithvi: I am concerned about garbage collection when a service has many endpoints; updating these in place would perform much better than creating a new object on each update for every endpoint.

yurishkuro: we're talking about updates that happen once a minute. GC is not a concern at all at this rate.

vprithvi: Yeah, but it is wasteful to create and throw away up to 2000 probabilistic samplers every minute when we don't need to. That being said, I like immutability because it simplifies certain code. In this case, making all of the Samplers immutable will lead to easier to understand code.
Do you think we should do that at the expense of gc (which even at 2k per minute would only be noticed by very constrained applications)?

vprithvi: @yurishkuro Could you elaborate more on why you prefer these to be final?

Having these be mutable means that we have have fields in `GuaranteedThroughputSampler` and `PerOperationSampler` be final. They also lend themselves to elegant tests because we only need to verify that these samplers correctly call the `ProbabilisticSampler#update`.

I don't think having `update` always return a new instance is a good idea because they can simply use the constructor to create it. I feel that in that case, any custom behavior like short circuiting creation should lie with the caller. I particularly don't like the idea of having a `getOrCreate` endpoint because I feel its very prone to errors.

jaeger-core/src/main/java/com/uber/jaeger/samplers/RemoteControlledSampler.java
vprithvi: Done

jaeger-core/src/main/java/com/uber/jaeger/samplers/PerOperationSampler.java
vprithvi: Removed

jaeger-core/src/main/java/com/uber/jaeger/samplers/RemoteControlledSampler.java
```diff
@@ -87,38 +89,35 @@ private void updateSampler() {
...
+      log.info("No strategy present in response.");
     }
 
-    if (!this.sampler.equals(newSampler)) {
-      synchronized (this) {
-        this.sampler = newSampler;
-        metrics.samplerUpdated.inc(1);
-      }
+    if (oldSampler != sampler) {
+      metrics.samplerUpdated.inc(1);
```
vprithvi: @yurishkuro This metric updates whenever there is a change in the parameters for a Sampler since the last minute. As such, I think that if we care about it, it should be emitted from a collector. 
We already collect metrics when the sampler update fails, which is far more informative for debugging. What are your thoughts on removing this metric from the client?

jaeger-core/src/main/java/com/uber/jaeger/samplers/ProbabilisticSampler.java
yurishkuro: Immutable structures are generally easier to reason about, and easier to get right, especially w.r.t. thread safety. The PerOperationSampler is inherently mutable, I am not sure what you see as being `final` there. The GuaranteedThroughputSampler is still simple enough to be immutable.

The `update` method is just one way of building an API. What you're really doing there is asking "if these were your new parameters, do you need to update?"

I still don't see the GC concerns. First, 2000 endpoints is unrealistic number - if any service has that many, they have bigger problems. But even with that number, the actual rate of change is significantly lower than once a minute, because the sampling rates are supposed to stabilize quickly.

vprithvi: Why is the `PerOperationSampler` inherently mutable? 
The current interface on `PerOperationSampler` supports the `update` operation, but why not make it Immutable as well? 
I think my argument is now about inconsistencies between samplers. I would like for all of them to by immutable.

jaeger-core/src/main/java/com/uber/jaeger/samplers/RemoteControlledSampler.java
yurishkuro: I think we need to keep these. Collector has no idea when clients actually update their samplers, it only knows when it calculates them. This metric allows you to monitor the behavior of the client itself, see how much churn in the samplers it is getting, etc.

jaeger-core/src/main/java/com/uber/jaeger/samplers/RemoteControlledSampler.java
```diff
@@ -86,33 +89,35 @@ private void updateSampler() {
...
+    Sampler oldSampler = sampler;
+
+    if (response.getOperationSampling() != null) {
+      OperationSamplingParameters samplingParameters = response.getOperationSampling();
+      if (sampler instanceof PerOperationSampler) {
+        if (((PerOperationSampler) sampler).update(samplingParameters)) {
+          metrics.samplerUpdated.inc(1);
+        }
+      }
+      sampler = new PerOperationSampler(maxOperations, response.getOperationSampling());
```
black-adder: this should be in an else stmt

jaeger-core/src/main/java/com/uber/jaeger/samplers/RemoteControlledSampler.java
```diff
@@ -86,33 +89,35 @@ private void updateSampler() {
...
-        this.sampler = newSampler;
-        metrics.samplerUpdated.inc(1);
-      }
+    if (oldSampler != sampler) {
+      metrics.samplerUpdated.inc(1);
     }
+
   }
 
   private Sampler extractSampler(SamplingStrategyResponse response)
```
black-adder: is this function used anywhere now?

jaeger-core/src/main/java/com/uber/jaeger/samplers/PerOperationSampler.java
```diff
@@ -57,36 +62,42 @@ public PerOperationSampler(int maxOperations, OperationSamplingParameters strate
...
-        isUpdated |= guaranteedThroughputSampler.update(samplingRate, lowerBound);
-      } else if (operationNameToSampler.size() < maxOperations) {
-        isUpdated = (operationNameToSampler
-                        .putIfAbsent(operation,
-                                     new GuaranteedThroughputSampler(samplingRate,lowerBound)))
-                        == null;
-      } else {
-        log.info("Exceeded the maximum number of operations({}) for per operations sampling",
-                 maxOperations);
+      GuaranteedThroughputSampler sampler = new GuaranteedThroughputSampler(samplingRate, lowerBound);
```
black-adder: So looking at this it seems like we're recreating the guaranteedthroughputsampler every minute? The issue is that the bucket approach that we take for the lowerbound sampler will be broken if we recreate it every minute.

vprithvi: Yeah, we are recreating it every minute, but why does it break the lower bound?

black-adder: the way the ratelimit is implemented is that when the sampler is created, it starts off with 0 credits which is incremented once a tick until it goes over 1. The issue is that the bucket is reset to 0 every minute and never has chance to grow so we'll end up with never emitting any lowerbound sampled traces. The default lowerbound rate limit is 1/10 minutes. I haven't checked the Java codebase but this is how it is in all other languages.

jaeger-core/src/main/java/com/uber/jaeger/samplers/PerOperationSampler.java
```diff
@@ -0,0 +1,120 @@
...
+        operationNameToSampler.replace(operation, sampler);
+        isUpdated = true;
+      }
+    }
+
+    return isUpdated;
+  }
+
+  @Override
+  public SamplingStatus sample(String operation, long id) {
```
yurishkuro: this is not the same behavior as in Go - https://github.com/uber/jaeger-client-go/blob/master/sampler.go#L310

when we see new op names we should eagerly create GTSampler until max is reached.

vprithvi: Interesting, so sample has some side effects. 

I don't like the design of this object. The behavior of the go client is thus,

On `#updates`, rely on the agent to supply a list of `PerOperationSamplingStrategies` and create GTSamplers for all of them. [`#update`](https://github.com/uber/jaeger-client-go/blob/master/sampler.go#L343) doesn't do any checking of the size of the response. 

On `#sample`, check the size of the size of the samplers map, and add a `operation -> GTSampler` if it cannot find `operation` in the map. If the limit is reached, fallback on the `defaultSampler`, which is a probabilistic sampler.

As such, we are setting a default of a GTSampler for endpoints that we haven't seen yet when we have space in the map, but we use a default of a Probabilistic sampler otherwise. I don't see the benefit of doing this. 

Additionally, having a GTSampler as the default means that `sample` can be simpler and side effect free.

yurishkuro: > As such, we are setting a default of a GTSampler for endpoints that we haven't seen yet when we have space in the map, but we use a default of a Probabilistic sampler otherwise. I don't see the benefit of doing this.

This is one of the objective of the project, to use GTSampler for each endpoint to guarantee minimum rate. Default probabilistic sampler (a) is not going to provide that guarantee, and (b) is only used when service exceeds the limit on the number of operations we want to track, which should be never. GTSampler as default is not any better when applied to random endpoints, and is largely irrelevant because of (b) above.

vprithvi: Precisely my point, having a default of the GTSampler isn't any different from having a copy of a GTSampler for any endpoints that don't have probabilities computed at the collector. 

By having the GTSampler as a default, `sample` would no longer have any side effects, and the code will never fall back to using a `Probabilistic` sampler. As a benefit, the code would be simpler to reason about.

vprithvi: That's right, I'll update

jaeger-core/src/main/java/com/uber/jaeger/samplers/ProbabilisticSampler.java
yurishkuro: I agree with @black-adder - this increases the size of the diff for no reason. And I also prefer constructor to be consistent in its use of `this.` prefix - if it's used for something, better be used for everything.

vprithvi: I'll change this back; we don't have such consistency within constructors elsewhere in the codebase, and such consistency is not required by the style guide that we supply with this project.

jaeger-core/src/main/java/com/uber/jaeger/samplers/PerOperationSampler.java
```diff
@@ -0,0 +1,129 @@
...
+      sampler = new GuaranteedThroughputSampler(lowerBound, defaultSamplingProbability);
+      operationNameToSampler.put(operation, sampler);
+      return sampler.sample(operation, id);
+    }
+
+    return defaultSampler.sample(operation, id);
+  }
+
+  @Override
+  public void close() {
```
yurishkuro: synchronize

jaeger-core/src/main/java/com/uber/jaeger/samplers/PerOperationSampler.java
```diff
@@ -0,0 +1,129 @@
...
+    defaultSampler.close();
+    for (Map.Entry<String, GuaranteedThroughputSampler> sampler : operationNameToSampler.entrySet()) {
```
yurishkuro: can simplify: `operationNameToSampler.values()`

jaeger-core/src/main/java/com/uber/jaeger/samplers/ProbabilisticSampler.java
```diff
@@ -26,32 +26,38 @@
...
 
-  private final long positiveSamplingBoundary;
-  private final long negativeSamplingBoundary;
-  private final double samplingRate;
-  private final Map<String, Object> tags;
+  @Getter
+  double samplingRate;
+
+  long positiveSamplingBoundary;
+  long negativeSamplingBoundary;
```
yurishkuro: I still don't understand why we're not declaring these private final. I don't care if lombok makes them final automatically or not, it improves readability/understanding of the class by stating this explicitly.

jaeger-core/src/main/java/com/uber/jaeger/samplers/ProbabilisticSampler.java
```diff
@@ -26,32 +26,38 @@
...
-    this.negativeSamplingBoundary = (long) ((1L << 63) * samplingRate);
 
-    Map<String, Object> tags = new HashMap<>();
+    HashMap<String, Object> tags = new HashMap<>();
     tags.put(Constants.SAMPLER_TYPE_TAG_KEY, TYPE);
     tags.put(Constants.SAMPLER_PARAM_TAG_KEY, samplingRate);
     this.tags = Collections.unmodifiableMap(tags);
+
+    this.positiveSamplingBoundary = (long) (((1L << 63) - 1) * samplingRate);
+    this.negativeSamplingBoundary = (long) ((1L << 63) * samplingRate);
```
yurishkuro: One again, these 7 lines of change above make no difference except for making the diff bigger. Why introduce it and make us review a complete no-op change?

jaeger-core/src/main/java/com/uber/jaeger/samplers/RateLimitingSampler.java
```diff
@@ -27,22 +27,29 @@
...
 import java.util.Map;
+import lombok.AccessLevel;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
 import lombok.ToString;
+import lombok.Value;
 
-@SuppressWarnings("EqualsHashCode")
+@Value
 @ToString(exclude = "rateLimiter")
```
yurishkuro: can exclude tags too

jaeger-core/src/main/java/com/uber/jaeger/samplers/ProbabilisticSampler.java
vprithvi: I'll change them back, during the first iteration, they made sense, but now it looks like i'm simply moving things over for no reason.

jaeger-core/src/main/java/com/uber/jaeger/samplers/ProbabilisticSampler.java
```diff
@@ -26,32 +26,38 @@
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
-import lombok.ToString;
+import lombok.AccessLevel;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.Value;
```
yurishkuro: also, all this lombok-ification is again making the diff 600+ lines long. Why not a separate PR?

vprithvi: Will do.

jaeger-core/src/main/java/com/uber/jaeger/samplers/PerOperationSampler.java
vprithvi: done

jaeger-core/src/main/java/com/uber/jaeger/samplers/PerOperationSampler.java
vprithvi: done

jaeger-core/src/main/java/com/uber/jaeger/samplers/PerOperationSampler.java
```diff
@@ -0,0 +1,126 @@
...
+ */
+@Slf4j
+@Data
+@Getter(AccessLevel.NONE)
+public class PerOperationSampler implements Sampler {
+  private final HashMap<String, GuaranteedThroughputSampler> operationNameToSampler;
+  private final int maxOperations;
+  private ProbabilisticSampler defaultSampler;
+  private double lowerBound;
+  private double defaultSamplingProbability;
```
yurishkuro: Also, even though we didn't do it in other clients, I rather question why we can't have the default sampler as GuaranteedThroughputSampler as well - it doesn't generate excessive traffic, and is no worse that probabilistic sampler.

jaeger-core/src/main/java/com/uber/jaeger/samplers/PerOperationSampler.java
```diff
@@ -0,0 +1,126 @@
...
+
+  @Override
+  public synchronized SamplingStatus sample(String operation, long id) {
+    GuaranteedThroughputSampler sampler = operationNameToSampler.get(operation);
+    if (sampler != null) {
+      return sampler.sample(operation, id);
+    }
+
+    if (operationNameToSampler.size() < maxOperations) {
+      sampler = new GuaranteedThroughputSampler(lowerBound, defaultSamplingProbability);
```
yurishkuro: per comment above, if we make default sampler GuaranteedThroughputSampler, then here we can just do something like `sampler = defaultSampler.copy()`. This will make this class simpler.

jaeger-core/src/main/java/com/uber/jaeger/samplers/RemoteControlledSampler.java
```diff
@@ -86,36 +88,31 @@ private void updateSampler() {
...
       ProbabilisticSamplingStrategy strategy = response.getProbabilisticSampling();
-      return new ProbabilisticSampler(strategy.getSamplingRate());
-    }
-
-    if (SamplingStrategyType.RATE_LIMITING.equals(response.getStrategyType())) {
+      sampler = new ProbabilisticSampler(strategy.getSamplingRate());
+    } else if (response.getRateLimitingSampling() != null) {
       RateLimitingSamplingStrategy strategy = response.getRateLimitingSampling();
-      return new RateLimitingSampler(strategy.getMaxTracesPerSecond());
+      sampler = new RateLimitingSampler(strategy.getMaxTracesPerSecond());
```
yurishkuro: this resets the internal state of rate limiter. Only update when different.

jaeger-core/src/main/java/com/uber/jaeger/samplers/RemoteControlledSampler.java
```diff
@@ -86,36 +88,31 @@ private void updateSampler() {
...
+    } else {
+      metrics.samplerParsingFailure.inc(1);
+      log.info("No strategy present in response.");
     }
 
-    throw new UnknownSamplingStrategyException(
-        String.format("Unsupported sampling strategy type %s", response.getStrategyType()));
+    if (!oldSampler.equals(sampler)) {
```
yurishkuro: inefficient, especially for per-op sampler, since you already checked above it if's different or not.

jaeger-core/src/test/java/com/uber/jaeger/samplers/PerOperationSamplerTest.java
```diff
@@ -0,0 +1,156 @@
...
+  @Test
+  public void testCreateGuaranteedSamplerOnSample() {
+    String newOperation = "new operation";
+    undertest.sample(newOperation, traceId);
+    assertEquals(new GuaranteedThroughputSampler(0, 0), operationToSamplers.get(newOperation));
+  }
+
+  @Test
+  public void testPerOperationSampler() {
+    GuaranteedThroughputSampler guaranteedThroughputSampler = mock(GuaranteedThroughputSampler.class);
```
yurishkuro: nit: var name like `sampler` or `s1` would make for more readable code

jaeger-core/src/test/java/com/uber/jaeger/samplers/PerOperationSamplerTest.java
```diff
@@ -0,0 +1,156 @@
...
+    SamplingStatus samplingStatus = undertest.sample(operation, traceId);
+    assertTrue(samplingStatus.isSampled());
+    verify(defaultProbabilisticSampler).sample(operation, traceId);
+  }
+
+  @Test
+  public void testCreateGuaranteedSamplerOnSample() {
+    String newOperation = "new operation";
+    undertest.sample(newOperation, traceId);
+    assertEquals(new GuaranteedThroughputSampler(0, 0), operationToSamplers.get(newOperation));
```
yurishkuro: not sure I follow - why `0, 0` ?

jaeger-core/src/test/java/com/uber/jaeger/samplers/PerOperationSamplerTest.java
```diff
@@ -0,0 +1,156 @@
...
+  @Test
+  public void testFallbackToDefaultProbabilisticSampler() {
+    undertest = new PerOperationSampler(0, operationToSamplers, defaultProbabilisticSampler);
+    SamplingStatus samplingStatus = undertest.sample(operation, traceId);
+    assertTrue(samplingStatus.isSampled());
+    verify(defaultProbabilisticSampler).sample(operation, traceId);
+  }
+
+  @Test
+  public void testCreateGuaranteedSamplerOnSample() {
```
yurishkuro: s/OnSample/OnUnseenOperation/

jaeger-core/src/test/java/com/uber/jaeger/samplers/PerOperationSamplerTest.java
```diff
@@ -0,0 +1,156 @@
...
+    String newOperation = "new operation";
+    undertest.sample(newOperation, traceId);
+    assertEquals(new GuaranteedThroughputSampler(0, 0), operationToSamplers.get(newOperation));
+  }
+
+  @Test
+  public void testPerOperationSampler() {
```
yurishkuro: nit: naming - something about testing against known op name

jaeger-core/src/test/java/com/uber/jaeger/samplers/PerOperationSamplerTest.java
```diff
@@ -0,0 +1,156 @@
...
+    GuaranteedThroughputSampler guaranteedThroughputSampler = mock(GuaranteedThroughputSampler.class);
+    operationToSamplers.put(operation, guaranteedThroughputSampler);
+
+    when(guaranteedThroughputSampler.sample(operation, traceId)).thenReturn(SamplingStatus.of(true, new HashMap<String, Object>()));
+
+    SamplingStatus samplingStatus = undertest.sample(operation, traceId);
+    assertTrue(samplingStatus.isSampled());
+    verify(guaranteedThroughputSampler).sample(operation, traceId);
+    verifyNoMoreInteractions(defaultProbabilisticSampler);
```
yurishkuro: +1, but conversely a test is missing when you do fall back onto default sampler (when maxOps limit is reached)

jaeger-core/src/test/java/com/uber/jaeger/samplers/PerOperationSamplerTest.java
```diff
@@ -0,0 +1,156 @@
...
+    List<PerOperationSamplingParameters> parametersList = new ArrayList<>();
+    parametersList.add(perOperationSamplingParameters);
+
+    OperationSamplingParameters parameters =
+        new OperationSamplingParameters(DEFAULT_SAMPLING_PROBABILITY,
+                                        DEFAULT_LOWER_BOUND_TRACES_PER_SECOND, parametersList);
+
+    assertTrue(undertest.update(parameters));
+    //Checks that creating a new instance with the given parameters is the same as updating an existing instance
+    assertEquals(new PerOperationSampler(MAX_OPERATIONS, parameters), undertest);
```
yurishkuro: need a test where the update is a no-op (which is the prevalent scenario)

jaeger-core/src/test/java/com/uber/jaeger/samplers/RemoteControlledSamplerTest.java
```diff
@@ -73,7 +73,10 @@ public void testGetRateLimitingRemoteSamplingStrategy() throws Exception {
 
   private RemoteControlledSampler makeSamplerWith(Sampler underlyingSampler) throws Exception {
     HTTPSamplingManager manager = mock(HTTPSamplingManager.class);
-    SamplingStrategyResponse response = new SamplingStrategyResponse(SamplingStrategyType.PROBABILISTIC, null, null);
+    SamplingStrategyResponse response = new SamplingStrategyResponse(SamplingStrategyType.PROBABILISTIC,
```
yurishkuro: break to new line after `(` for better formatting

jaeger-core/src/main/java/com/uber/jaeger/samplers/RemoteControlledSampler.java
```diff
@@ -86,36 +88,31 @@ private void updateSampler() {
...
     }
 
-    Sampler newSampler;
-    try {
-      newSampler = extractSampler(response);
-    } catch (UnknownSamplingStrategyException e) {
-      metrics.samplerParsingFailure.inc(1);
-      return; // sampler updateGauge without a new sampler
-    }
+    Sampler oldSampler = sampler;
```
vprithvi: Could you please elaborate?
My understanding of the [volatile spec](http://docs.oracle.com/javase/specs/jls/se7/html/jls-8.html#jls-8.3.1.4) is that this should be thread safe.

jaeger-core/src/main/java/com/uber/jaeger/samplers/RemoteControlledSampler.java
```diff
@@ -22,29 +22,31 @@
...
-import com.uber.jaeger.samplers.http.SamplingStrategyType;
 
 import java.util.Timer;
 import java.util.TimerTask;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import lombok.ToString;
+import lombok.extern.slf4j.Slf4j;
 
 @SuppressWarnings("EqualsHashCode")
 @ToString(exclude = {"pollTimer", "lock"})
```
vprithvi: This is preexisting code, not part of this commit. 
It is used for initialization logging to what the sampler has been initialized with.

jaeger-core/src/main/java/com/uber/jaeger/samplers/PerOperationSampler.java
```diff
@@ -0,0 +1,126 @@
...
+import lombok.Data;
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * Computes {@link #sample(String, long)} using the name of the operation, and maintains a specific
+ * {@link GuaranteedThroughputSampler} instance for each operation.
+ */
+@Slf4j
+@Data
```
vprithvi: It's just shorthand, we use the @ToString, and @EqualsAndHashCode

jaeger-core/src/main/java/com/uber/jaeger/samplers/RemoteControlledSampler.java
vprithvi: will refactor

jaeger-core/src/main/java/com/uber/jaeger/samplers/PerOperationSampler.java
yurishkuro: so why not use them explicitly? `@Data` is more than just those two, and is confusing (just like `@Value` would be) for a class that is not a pure value object.

jaeger-core/src/main/java/com/uber/jaeger/samplers/RemoteControlledSampler.java
yurishkuro: What I mean is that you're reading `sampler` member variable multiple times and making decisions between those reads, so you have a race condition.

It may not be critical because update() should only be called internally by a single thread. Worth a comment somewhere.

jaeger-core/src/test/java/com/uber/jaeger/samplers/PerOperationSamplerTest.java
vprithvi: I already have this test. 
```
  @Test
  public void testFallbackToDefaultProbabilisticSampler() {
    undertest = new PerOperationSampler(0, operationToSamplers, defaultProbabilisticSampler);
    SamplingStatus samplingStatus = undertest.sample(operation, traceId);
    assertTrue(samplingStatus.isSampled());
    verify(defaultProbabilisticSampler).sample(operation, traceId);
  }
```

jaeger-core/src/test/java/com/uber/jaeger/samplers/PerOperationSamplerTest.java
vprithvi: No particular reason, I'll change them to something non default

jaeger-core/src/main/java/com/uber/jaeger/samplers/PerOperationSampler.java
vprithvi: I think we should have this discussion separately, I don't like it because when used that way, the GuaranteedTrhoughputSampler no longer fulfills it's guarantee.

jaeger-core/src/main/java/com/uber/jaeger/samplers/PerOperationSampler.java
vprithvi: Done.

jaeger-core/src/main/java/com/uber/jaeger/samplers/GuaranteedThroughputSampler.java
```diff
@@ -0,0 +1,75 @@
...
+import lombok.Value;
+
+/**
+ * {@link GuaranteedThroughputSampler} is a {@link Sampler} that guarantees a throughput by using
+ * a {@link ProbabilisticSampler} and {@link RateLimitingSampler} in tandem.
+ *
+ * The RateLimitingSampler is used to establish a lowerBound so that every operation is sampled
+ * at least once in the time interval defined by the lowerBound.
+ */
+@Value
```
vprithvi: Done

jaeger-core/src/test/java/com/uber/jaeger/samplers/RemoteControlledSamplerTest.java
vprithvi: Done

jaeger-core/src/main/java/com/uber/jaeger/samplers/RemoteControlledSampler.java
```diff
@@ -75,47 +77,58 @@ public ReentrantReadWriteLock getLock() {
     return lock;
   }
 
-  private void updateSampler() {
+  /**
+   * Updates {@link #sampler} to a new sampler when it is different.
+   */
+  private synchronized void updateSampler() {
```
yurishkuro: this is way too coarse, you're going to block creation of all traces in the process while you're making a network call to the agent

jaeger-core/src/test/java/com/uber/jaeger/samplers/PerOperationSamplerTest.java
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  @Test
+  public void testNoopUpdate(){
+    List<PerOperationSamplingParameters> parametersList = new ArrayList<>();
+    OperationSamplingParameters parameters = new OperationSamplingParameters(0, 0, parametersList);
```
yurishkuro: this doesn't really test the deep comparison feature

yurishkuro: also, why `0, 0`? Isn't `underTest` initialized with `DEFULT_xx` constants that are not 0s?

yurishkuro: lastly, I would want this test to actually verify that per-op samplers did not change, instead of relying on the method claiming that via return param.

PR #119, opened 2016-12-22
Rename strategy to parameters

vprithvi:
- Rename `ProbabilisticSamplingStrategy` to `ProbabilisticSamplingParameters`
- Rename `RateLimitingSamplingStrategy` to `RateLimitingSamplingParameters`

These classes are used for the json model, Parameters captures what they are better than Strategy.


yurishkuro:
Is it worth it? This way the classes are named differently than the IDL types.

vprithvi:
>Is it worth it? This way the classes are named differently than the IDL types.

I think it is.
With both `RateLimitingSamplingStrategy` and `RateLimitingSampler` in the code base, it's not easy to tell at first glance what each of these do whereas the word `parameter` makes it extremely clear that there is no logic in the class, and it simply used for holding parameters for a sampler.

PR #124, opened 2017-02-01
Run RemoteReporter's flush timer in daemon thread.

aristidisp:
This prevents the flush timer's thread from keeping a process alive after main() has returned.

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/reporters/RemoteReporter.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/reporters/RemoteReporter.java
@@ -31,7 +31,6 @@ import java.util.Timer;
 import java.util.TimerTask;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
-import java.util.concurrent.atomic.AtomicLong;
 import lombok.ToString;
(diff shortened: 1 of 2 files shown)


jaeger-core/src/main/java/com/uber/jaeger/reporters/RemoteReporter.java
```diff
@@ -59,12 +61,14 @@ public RemoteReporter(final Sender sender, int flushInterval, int maxQueueSize,
     commandQueue = new ArrayBlockingQueue<>(maxQueueSize);
 
     // start a thread to append spans
+    long serialNumber = nextSerialNumber.getAndIncrement();
```
yurishkuro: I am not sure why we need this. In a real application there would be only one thread with each name.

aristidisp: Seemed like a good idea to ensure thread name uniqueness, but you're right, it's not necessary.  This is my first pull request on GitHub: do I make the change in my repo then push, or do you do it here directly?

yurishkuro: yes, you make the change in the same branch as you did and just push that branch, the PR should be automatically updated.

PR #127, opened 2017-02-21
Fix leaky bucket bug

black-adder:
Fix bug where if the creditsPerSecond < itemCost, the leaky bucket never fills up. Added a maxBalance to set the upperbound of the balance.

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/utils/RateLimiter.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/utils/RateLimiter.java
@@ -34,7 +34,7 @@ public class RateLimiter {
 
   public RateLimiter(double creditsPerSecond, double maxBalance, Clock clock) {
     this.clock = clock;
-    this.balance = creditsPerSecond;
+    this.balance = maxBalance;
     this.maxBalance = maxBalance;
     this.creditsPerNanosecond = creditsPerSecond / 1.0e9;
   }
--- a/jaeger-core/src/main/java/com/uber/jaeger/utils/RateLimiter.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/utils/RateLimiter.java
@@ -22,20 +22,20 @@
 package com.uber.jaeger.utils;
 
 public class RateLimiter {
-  private final double creditsPerSecond;
   private final double creditsPerNanosecond;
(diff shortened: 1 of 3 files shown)


vprithvi:
Could you provide a description of what the bug was and how it is fixed in the commit?
Providing more information in commit messages makes `git log` significantly more useful

jaeger-core/src/main/java/com/uber/jaeger/utils/RateLimiter.java
```diff
@@ -22,20 +22,20 @@
...
   private long lastTick;
 
-  public RateLimiter(double creditsPerSecond) {
-    this(creditsPerSecond, new SystemClock());
+  public RateLimiter(double creditsPerSecond, double maxBalance) {
+    this(creditsPerSecond, maxBalance, new SystemClock());
   }
 
-  public RateLimiter(double creditsPerSecond, Clock clock) {
+  public RateLimiter(double creditsPerSecond, double maxBalance, Clock clock) {
```
vprithvi: I think this interface should be something like `public RateLimiter(double rate, Seconds timePeriod)`
This reduces the cognitive overhead of the code and allows a reader to easily determine how many requests per a given time period is allowed by the RateLimiter. This is also consistent with the behavior of rate limiter interface before your commit. 

I don't like the interface you propose because it exposes the internals of a leaky bucket rate limiter, and expects users to be familiar with how leaky bucket rate limiting works (as opposed to saying give me a rate limiter that lets in `x` requests in `y` seconds)

yurishkuro: this rate limiter exposes the notion of item cost to the user (and works with variable item costs), so I don't think it's unreasonable to ask people for the other two params which are related to the "balance" maintained by the algorithm.

vprithvi: I think it is unnecessary to force an user to compute `creditsPerSecond` and `maxBalance` to do the basic function of rate limiting.

As a glance, looking at something like `RateLimiter(creditsPerSecond=2, maxBalance=100)` doesn't immediately tell a reader that this is a RateLimiter that rate limits at around 100 requests per 50s when the item cost is 1.

PR #131, opened 2017-03-10
Bump opentracingVersion to latest

black-adder:


jaeger-core/src/test/java/com/uber/jaeger/SpanTest.java
```diff
@@ -21,9 +21,7 @@
  */
 package com.uber.jaeger;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.*;
```
black-adder: interesting, I didn't do this manually.

jaeger-core/src/main/java/com/uber/jaeger/Span.java
```diff
@@ -267,19 +267,48 @@ private Span setTagAsObject(String key, Object value) {
...
+
+  @Override
+  public Span log(long timestampMicroseconds, Map<String, ?> fields) {
+    synchronized (this) {
+      if (context.isSampled()) {
+        if (logs == null) {
+          this.logs = new ArrayList<>();
+        }
+        for (Map.Entry<String, ?> kv : fields.entrySet()) {
+          logs.add(new LogData(timestampMicroseconds, kv.getKey(), kv.getValue()));
```
black-adder: The intended use for jaeger log fields were that each log can be associated with multiple tags. Here, I'm creating a new log for each tag. In the JaegerThriftConverter I can add a special case for when the value is a map and generate multiple tags for the same log.

yurishkuro: -1. A log record is a timestamp + a set of fields.

jaeger-core/src/test/java/com/uber/jaeger/SpanTest.java
yurishkuro: you need to fix your IDE formatter. I think readme has links to IntelliJ settings

jaeger-core/src/main/java/com/uber/jaeger/Span.java
yurishkuro: also, we don't want to do the conversion of fields here, we just want to stash the map somewhere and only do conversion when the span is reported.

build.gradle
```diff
@@ -8,7 +8,7 @@ plugins {
 }
 
 
-ext.opentracingVersion = '0.15.0'
+ext.opentracingVersion = '0.20.9'
```
yurishkuro: there may be 0.20.10

jaeger-core/src/main/java/com/uber/jaeger/Span.java
```diff
@@ -267,19 +267,46 @@ private Span setTagAsObject(String key, Object value) {
...
+  }
+
+  @Override
+  public Span log(long timestampMicroseconds, Map<String, ?> fields) {
+    synchronized (this) {
+      if (context.isSampled()) {
+        if (logs == null) {
+          this.logs = new ArrayList<>();
+        }
+        logs.add(new LogData(timestampMicroseconds, "", fields));
```
yurishkuro: Self-inflicted inefficiency here, you reuse (String, Object) signature and then are forced to check for instanceof Map. Not to mention that there may be a payload that's a map.

I suggest simply having another constructor for LogData that stores `Map<String, ?>` fields proper.

jaeger-core/src/main/java/com/uber/jaeger/reporters/protocols/JaegerThriftSpanConverter.java
```diff
@@ -58,8 +58,12 @@
       for (LogData logData : logs) {
         Log jLog = new Log();
         jLog.setTimestamp(logData.getTime());
-        final Tag tag = buildTag(logData.getMessage(), logData.getPayload());
-        jLog.setFields(new ArrayList<Tag>() {{add(tag);}});
+        if (logData.getFields() != null) {
+          jLog.setFields(buildTags(logData.getFields()));
+        } else {
+          final Tag tag = buildTag(logData.getMessage(), logData.getPayload());
```
yurishkuro: I think this is incorrect. Should be `"event": message, "payload": payload` (and drop either if null)

jaeger-core/src/main/java/com/uber/jaeger/Span.java
```diff
@@ -267,19 +267,52 @@ private Span setTagAsObject(String key, Object value) {
...
+  }
+
+  @Override
   public Span log(String message, /* @Nullable */ Object payload) {
     return log(tracer.clock().currentTimeMicros(), message, payload);
   }
 
   @Override
-  public Span log(long instantMicroseconds, String message, /* @Nullable */ Object payload) {
+  public Span log(long timestampMicroseconds, String message, /* @Nullable */ Object payload) {
```
black-adder: given that this function is deprecated, I'm not going to do a check here to see if the payload is a map.

yurishkuro: I wasn't suggesting it. Payload as map should be stringified as any other object.

jaeger-core/src/main/java/com/uber/jaeger/LogData.java
```diff
@@ -21,26 +21,28 @@
...
-    return time;
-  }
-
-  public String getMessage() {
-    return message;
-  }
-
-  public Object getPayload() {
-    return payload;
+  LogData(long time, Map fields) {
```
felixbarny: Add generics?

yurishkuro:
@black-adder did you cover the new methods? why did coverage go down?

black-adder:
I'm gonna go ahead and land this, the code that isnt covered wasn't covered before but I moved it one line down and all hell broke loose. Will add followup PR to cover the rest.

PR #132, opened 2017-03-11
Java 6 compatibility

### LINKED ISSUE #129: Java 6 compatibility
Hi,

it would be nice if Jaeger was Java 6 compatible as I want to use it for stagemonitor, which currently is Java 6 compatible. Are there any reasons you don't want to support Java 6?

Cheers,
Felix

felixbarny:
Checked with IntelliJ:
Run Inspection by Name >
Usages of API which isn't available at the configured language level

jaeger-crossdock still requires Java 7 because it calls
InetAddress#getLoopbackAddress()

(closes #129)

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/utils/Java6CompatibleThreadLocalRandom.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/utils/Java6CompatibleThreadLocalRandom.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright (c) 2017, Uber Technologies, Inc
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package com.uber.jaeger.utils;
+
+import java.util.Random;
+import java.util.concurrent.ThreadLocalRandom;
+
+public final class Java6CompatibleThreadLocalRandom {
+
+  private static final boolean THREAD_LOCAL_RANDOM_PRESENT;
+
+  private static final String THREAD_LOCAL_RANDOM_CLASS_NAME =
+      "java.util.concurrent.ThreadLocalRandom";
+
+  static {
+    boolean present = true;
+    try {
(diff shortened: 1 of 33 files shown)


yurishkuro:
lgtm (with we could do reformatting separately).

it's ok to keep crossdock at 1.7, since it's just an integration test.

@vprithvi could you please review as well?

vprithvi:
Sure, I'll take a look later in the day

jaeger-core/src/main/java/com/uber/jaeger/utils/Java6CompatibleThreadLocalRandom.java
```diff
@@ -0,0 +1,67 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package com.uber.jaeger.utils;
+
+import java.util.Random;
+import java.util.concurrent.ThreadLocalRandom;
+
+public final class Java6CompatibleThreadLocalRandom {
```
vprithvi: Could you add tests for this?

felixbarny:
> LGTM, but I'm not convinced of the value of moving the entire project to Java 6. Wouldn't moving jaeger-core and jaeger-context satisfy your use case?

If jaeger-zipkin also was 1.6 it would satisfy *my* use case. But I think it's a bit confusing and inconsistent for developers and users to have a mix of java versions in published artifacts. Also, in the other artifacts I almost had to change nothing. All in all it would make jaeger accessible to a larger user base without much effort.

> This should be automated so that a future commit doesn't introduce such usages. I've heard good things about the AnimalSniffer plugin for this purpose.

I tried this plugin but didn't get it to work at first. Was a layer 8 problem ;)

felixbarny:
https://github.com/uber/jaeger-client-java/pull/131 also ist important for java 6 compatibility as the currently used ot-api version (0.15.0) is not java 6 compatible while the recent one is.

jaeger-thrift/build.gradle
```diff
@@ -7,8 +7,16 @@ description = 'Generated thrift code'
...
     compile group: 'org.apache.thrift', name: 'libthrift', version: apacheThriftVersion
+
+    signature 'org.codehaus.mojo.signature:java16:1.1@signature'
 }
 
+/**
+ * Note that you need to install thrift 0.9.2 on your system for this task to work.
+ *
+ * If you are using macOS, you can install this specific version via homebrew like so:
+ * `brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/56d8c1eba1e5ac30290dd0c486f4bba37f821e42/Formula/thrift.rb`
```
yurishkuro: Nice touch!

jaeger-zipkin/src/main/java/com/uber/jaeger/senders/zipkin/ZipkinSender.java
```diff
@@ -101,15 +101,15 @@ public int append(Span span) throws SenderException {
     byte[] next = encoder.encode(backFillHostOnAnnotations(span));
     int messageSizeOfNextSpan = delegate.messageSizeInBytes(Collections.singletonList(next));
     // don't enqueue something larger than we can drain
-    if (Integer.compare(messageSizeOfNextSpan, delegate.messageMaxBytes()) > 0) {
+    if (Integer.valueOf(messageSizeOfNextSpan).compareTo(delegate.messageMaxBytes()) > 0) {
```
yurishkuro: how is this different from `messageSizeOfNextSpan > delegate.messageMaxBytes()` ?

jaeger-thrift/build.gradle
felixbarny: That took me quite some time to figure out 😅

jaeger-core/src/main/java/com/uber/jaeger/utils/Java6CompatibleThreadLocalRandom.java
felixbarny: Hard to test this in a meaningful way. Especially since Gradle 3 does not run with Java 6.

jaeger-zipkin/src/main/java/com/uber/jaeger/senders/zipkin/ZipkinSender.java
felixbarny: you're right

felixbarny:
Conflicts are resolved now. I also tried to improve the `Java6CompatibleThreadLocalRandomTest`.

PR #134, opened 2017-03-29
Add GuaranteedThroughputSampler#update

black-adder:
Bringing the adaptive GuaranteedThroughputSampler code more into line with the other clients.

- [x] Tests

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/samplers/GuaranteedThroughputSampler.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/samplers/GuaranteedThroughputSampler.java
@@ -21,10 +21,16 @@
  */
 package com.uber.jaeger.samplers;
 
+import com.uber.jaeger.Constants;
+import com.uber.jaeger.samplers.http.OperationSamplingParameters;
+import com.uber.jaeger.samplers.http.PerOperationSamplingParameters;
 import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
 import lombok.EqualsAndHashCode;
 import lombok.ToString;
+
+import java.util.HashMap;
+import java.util.Map;
 
 /**
  * {@link GuaranteedThroughputSampler} is a {@link Sampler} that guarantees a throughput by using
@@ -37,11 +43,39 @@ import lombok.ToString;
 @EqualsAndHashCode
 @AllArgsConstructor (access = AccessLevel.PACKAGE) // Visible for testing
 public final class GuaranteedThroughputSampler implements Sampler {
-  private final ProbabilisticSampler probabilisticSampler;
-  private final RateLimitingSampler lowerBoundSampler;
+  public static final String TYPE = "lowerbound";
+
+  private ProbabilisticSampler probabilisticSampler;
+  private RateLimitingSampler lowerBoundSampler;
+  private Map<String, Object> tags;
 
   GuaranteedThroughputSampler(double samplingRate, double lowerBound) {
-    this(new ProbabilisticSampler(samplingRate), new RateLimitingSampler(lowerBound));
+    tags = new HashMap<String, Object>();
+    tags.put(Constants.SAMPLER_TYPE_TAG_KEY, TYPE);
+    tags.put(Constants.SAMPLER_PARAM_TAG_KEY, samplingRate);
+
+    this.probabilisticSampler = new ProbabilisticSampler(samplingRate);
+    this.lowerBoundSampler = new RateLimitingSampler(lowerBound);
+  }
+
+  /**
+   * Updates the probabilistic and lowerBound samplers
+   * @param samplingRate The sampling rate for probabilistic sampling
+   * @param lowerBound The lower bound limit for lower bound sampling
+   * @return true iff any samplers were updated
+   */
+  public synchronized boolean update(double samplingRate, double lowerBound) {
+    boolean isUpdated = false;
+    if (samplingRate != probabilisticSampler.getSamplingRate()) {
+      probabilisticSampler = new ProbabilisticSampler(samplingRate);
+      tags.put(Constants.SAMPLER_PARAM_TAG_KEY, samplingRate);
+      isUpdated = true;
+    }
+    if (lowerBound != lowerBoundSampler.getMaxTracesPerSecond()) {
+      lowerBoundSampler = new RateLimitingSampler(lowerBound);
+      isUpdated = true;
+    }
+    return isUpdated;
   }
 
   /**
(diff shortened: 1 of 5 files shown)


jaeger-core/src/test/java/com/uber/jaeger/samplers/PerOperationSamplerTest.java
```diff
@@ -101,24 +101,24 @@ public void testPerOperationSamplerWithKnownOperation() {
...
-
-    OperationSamplingParameters parameters =
-        new OperationSamplingParameters(DEFAULT_SAMPLING_PROBABILITY,
-                                        DEFAULT_LOWER_BOUND_TRACES_PER_SECOND, parametersList);
-
-    assertTrue(undertest.update(parameters));
-    //Checks that creating a new instance with the given parameters is the same as updating an existing instance
-    assertEquals(new PerOperationSampler(MAX_OPERATIONS, parameters), undertest);
-  }
+//  @Test
```
black-adder: tis a flesh wound, gonna fix this.

jaeger-core/src/main/java/com/uber/jaeger/samplers/GuaranteedThroughputSampler.java
```diff
@@ -37,11 +43,39 @@
 @EqualsAndHashCode
 @AllArgsConstructor (access = AccessLevel.PACKAGE) // Visible for testing
 public final class GuaranteedThroughputSampler implements Sampler {
-  private final ProbabilisticSampler probabilisticSampler;
-  private final RateLimitingSampler lowerBoundSampler;
+  public static final String TYPE = "lowerbound";
+
+  private ProbabilisticSampler probabilisticSampler;
```
vprithvi: I prefer immutable objects to mutable ones, and would prefer that these fields are final. 
What problem are we trying to solve by having an `update` method?

black-adder: before, we would create a new guaranteedthroughput sampler on each peroperationsamper.update call if the strategy changed slightly (which is highly likely). By adding this update function in guaranteedthroughput sampler, we can update the 2 different constituent samplers separately. Creating a new RateLimitingSampler every update is a big nono since the samplers would sample every minute.

jaeger-core/src/test/java/com/uber/jaeger/samplers/GuaranteedThroughputSamplerTest.java
```diff
@@ -63,4 +65,27 @@ public void testProbabilityTagsOverrideRateLimitingTags() throws Exception {
     assertEquals(tags.get(Constants.SAMPLER_TYPE_TAG_KEY), ProbabilisticSampler.TYPE);
     assertEquals(tags.get(Constants.SAMPLER_PARAM_TAG_KEY), 0.999);
   }
+
+  @Test
+  public void testUpdate() {
```
vprithvi: The tests would be a lot more readable if the test cases are separated into something like  `testUpdate_probabilisticSampler, testUpdate_rateLimitingSampler`, etc

jaeger-core/src/test/java/com/uber/jaeger/samplers/GuaranteedThroughputSamplerTest.java
```diff
@@ -48,8 +50,8 @@ public void testRateLimitingLowerBound() throws Exception {
     Assert.assertTrue(samplingStatus.isSampled());
     Map<String, Object> tags = samplingStatus.getTags();
 
-    assertEquals(tags.get(Constants.SAMPLER_TYPE_TAG_KEY), RateLimitingSampler.TYPE);
-    assertEquals(tags.get(Constants.SAMPLER_PARAM_TAG_KEY), 1.0);
+    assertEquals(tags.get(Constants.SAMPLER_TYPE_TAG_KEY), GuaranteedThroughputSampler.TYPE);
+    assertEquals(tags.get(Constants.SAMPLER_PARAM_TAG_KEY), 0.0001);
```
vprithvi: Why do we return the probabilistic sampler's sampling rate even when the rate limiting lower bound sampler is triggered?

jaeger-core/src/test/java/com/uber/jaeger/samplers/PerOperationSamplerTest.java
```diff
@@ -103,21 +103,21 @@ public void testPerOperationSamplerWithKnownOperation() {
 
   @Test
   public void testUpdate(){
-    GuaranteedThroughputSampler guaranteedThroughputSampler = mock(GuaranteedThroughputSampler.class);
+    GuaranteedThroughputSampler guaranteedThroughputSampler = new GuaranteedThroughputSampler(SAMPLING_RATE, DEFAULT_LOWER_BOUND_TRACES_PER_SECOND);
     operationToSamplers.put(OPERATION, guaranteedThroughputSampler);
 
     PerOperationSamplingParameters perOperationSamplingParameters =
-        new PerOperationSamplingParameters(OPERATION, new ProbabilisticSamplingStrategy(SAMPLING_RATE));
+        new PerOperationSamplingParameters(OPERATION, new ProbabilisticSamplingStrategy(0.001));
```
vprithvi: Could you use named constants instead?

jaeger-core/src/test/java/com/uber/jaeger/samplers/PerOperationSamplerTest.java
```diff
@@ -103,21 +103,21 @@ public void testPerOperationSamplerWithKnownOperation() {
...
 
     OperationSamplingParameters parameters =
         new OperationSamplingParameters(DEFAULT_SAMPLING_PROBABILITY,
-                                        DEFAULT_LOWER_BOUND_TRACES_PER_SECOND, parametersList);
+                                        1.0, parametersList);
 
     assertTrue(undertest.update(parameters));
-    //Checks that creating a new instance with the given parameters is the same as updating an existing instance
-    assertEquals(new PerOperationSampler(MAX_OPERATIONS, parameters), undertest);
+    assertEquals(new GuaranteedThroughputSampler(0.001, 1.0),
```
vprithvi: This test doesn't really test whether the `guaranteedThroughputSampler` is a new instance, or one that was updated by using the `update` method. 
I suggest using a mock for the `guaranteedThroughputSampler` and verifying that update was called with the right parameters.

jaeger-core/src/test/java/com/uber/jaeger/samplers/GuaranteedThroughputSamplerTest.java
black-adder: It's an artifact for when I first designed this. The lowerbound sampler needed to return the probability that **didn't** sample it so on the backend we know that we should bump this probability. However, given that the backend is the source of probability now, this isn't needed but I'm keeping it for consistency sake before I rip it out of all the clients and return adaptive sampling tags.

PR #138, opened 2017-04-16
Java6 fix

felixbarny:
Sorry guys, I screwed that one up :(

I'm not sure why but the JVM seems to load all classes which are used in a (static) method. That makes it not safe to just check the threadLocalRandomPresent flag and access ThreadLocalRandom right away. Maybe it has to do with some JIT magic? BTW making the flag final did not help either.

I noticed that in the integration tests for stagemonitor: https://travis-ci.org/stagemonitor/stagemonitor-integration-tests/jobs/222501692#L700. Tested the fix locally with Java 6.


felixbarny:
Hey @raphw, could you explain us the magic?

raphw:
The JVM is really free to load a class referenced within an executed method at any time it wants to. The concept of optional types is not know to the runtime. Reflection would probably be the best way to guard against eager loading but there is a runtime impact. As for your solution, it should work with HotSpot but it would generally be safer to move code to a dispatcher class which is discarded in case Java 7 is not available.

felixbarny:
Thanks a lot for you quick answer! 

> it would generally be safer to move code to a dispatcher class which is discarded in case Java 7 is not available.

How would that look like? Do you have an example for this approach which is used in some OS project?

adriancole:
>
>
> it would generally be safer to move code to a dispatcher class which is
> discarded in case Java 7 is not available.
>
> How would that look like? Do you have an example for this approach which
> is used in some OS project?
>
I've used a Platform class (idea stolen from okhttp), which has a Java6
variant once via reflection you know which to keep.

raphw:
Have a look at these dispatchers which do the same job in Byte Buddy: https://github.com/raphw/byte-buddy/blob/master/byte-buddy-dep/src/main/java/net/bytebuddy/dynamic/loading/ByteArrayClassLoader.java

felixbarny:
@yurishkuro I think this should be good enough. Or should I reiterate?

felixbarny:
Nice! Are you planning to do a bugfix release? Otherwise, when is 0.19.0 scheduled?

PR #139, opened 2017-04-18
WIP: Move to jaeger.thrift

pavolloffay:
@vprithvi @yurishkuro could you please review?

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/reporters/protocols/ThriftSpanConverter.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/reporters/protocols/ThriftSpanConverter.java
@@ -21,6 +21,12 @@
  */
 package com.uber.jaeger.reporters.protocols;
 
+import java.nio.charset.Charset;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
 import com.twitter.zipkin.thriftjava.Annotation;
 import com.twitter.zipkin.thriftjava.AnnotationType;
 import com.twitter.zipkin.thriftjava.BinaryAnnotation;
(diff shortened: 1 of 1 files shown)


pavolloffay:
CLA signed

yurishkuro:
Have a couple of issues with this change.
  1. Ideally, we want to switch the client to `jaeger.thrift` that supports kv logs in the data model natively. It's on our roadmap anyway (#140), as currently only the Node.js client uses `jaeger.thrift` data model.
  1. In lieu of the previous point, the next best thing we can do is to replicate the logic in the Go client  where log fields are encoded as a JSON string and stored as Annotation almost without data loss (only the field value types are lost). Then in the backend we already have logic to process such format correctly and reconstruct the log fields from JSON before storing in OT-compatible format.

jaeger-core/src/main/java/com/uber/jaeger/reporters/protocols/ThriftSpanConverter.java
```diff
@@ -76,7 +77,32 @@
...
+        if (logData.getPayload() != null) {
+          value = logData.getPayload().toString();
+        } else if (logData.getFields() != null && !logData.getFields().isEmpty()) {
+          // special-case the "event" field which is similar to the semantics of a zipkin annotation
+          Object event = logData.getFields().get("event");
+          if (event != null && logData.getFields().size() == 1) {
+            value  = event.toString();
+          } else {
+            StringBuilder result = new StringBuilder();
+            for (Iterator<? extends Map.Entry<String, ?>> i = logData.getFields().entrySet().iterator(); i.hasNext(); ) {
```
adriancole: might want to comment that this is the existing practice in zipkin-go-opentracing.. we do need to put together an OT mapping page on the website.. something bas and I spoke of in the past.

pavolloffay: This is definitely a good idea :+1:

pavolloffay:
This should be an intermediate fix. You persuade me to port this to jaeger.thrift.

pavolloffay:
Closing in favour of #139

pavolloffay:
https://github.com/uber/jaeger-client-java/pull/142 is moving zipkin's `SpanThriftConverter` to `jaeger-zipkin` it is probably going to be used only for sending data to zipkin.. We could apply this patch if there is an interest

PR #141, opened 2017-04-19
Extract jaeger-b3 from jaeger-zipkin w/o any non-core deps

### LINKED ISSUE #136: Add B3TextMapCodec to jaeger-core
As B3 context propagation is not Zipkin specific, it would be nice if B3TextMapCodec was included in jaeger-core.

adriancole:
copy/pasting a hex codec was far simpler than trying to hack the build.

Fixes #136


adriancole:
np

PR #145, opened 2017-04-22
Use travis trusty to speed up build

pavolloffay:


vprithvi:
I'm not convinced that this by itself speeds up builds. Do we need to do anything additional?

build-times

yurishkuro:
The default `precise` is going away, `trusty` will become the default [according to this](https://blog.travis-ci.com/2017-04-17-precise-EOL). So I think it makes sense to do this, I will merge.

pavolloffay:
closing in favour of #154

PR #148, opened 2017-04-25
Fix checkstyle

pavolloffay:
Checkstyle runs for every `./gradlew test`. I also removed `config/checkstyle/checkstyle.xml` it seems that it is not used.

resolves https://github.com/uber/jaeger-client-java/issues/147

### DIFF
--- a/jaeger-crossdock/src/test/java/com/uber/jaeger/crossdock/resources/behavior/http/EndToEndBehaviorResourceTest.java
+++ b/jaeger-crossdock/src/test/java/com/uber/jaeger/crossdock/resources/behavior/http/EndToEndBehaviorResourceTest.java
@@ -21,17 +21,16 @@
  */
 package com.uber.jaeger.crossdock.resources.behavior.http;
 
+import static org.junit.Assert.*;
+
 import com.uber.jaeger.Span;
 import com.uber.jaeger.crossdock.api.*;
 import com.uber.jaeger.reporters.InMemoryReporter;
 import com.uber.jaeger.samplers.ConstSampler;
 import io.opentracing.Tracer;
+import java.util.*;
 import org.junit.Before;
 import org.junit.Test;
-
-import java.util.*;
-
-import static org.junit.Assert.*;
 
 public class EndToEndBehaviorResourceTest {
   private EndToEndBehaviorResource resource;
@@ -58,8 +57,7 @@ public class EndToEndBehaviorResourceTest {
   public void testCreateTraces() throws Exception {
     Map<String, String> tags = new HashMap<String, String>();
     tags.put("key", "value");
-    CreateTracesRequest request =
-        new CreateTracesRequest("const", "operation", 2, tags);
+    CreateTracesRequest request = new CreateTracesRequest("const", "operation", 2, tags);
 
     resource.createTraces(request);
     validateSpans(reporter.getSpans(), request);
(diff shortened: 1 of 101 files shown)


jaeger-crossdock/src/test/java/com/uber/jaeger/crossdock/resources/behavior/http/EndToEndBehaviorResourceTest.java
```diff
@@ -21,18 +21,17 @@
...
 package com.uber.jaeger.crossdock.resources.behavior.http;
 
+import static org.junit.Assert.*;
+
 import com.uber.jaeger.Span;
 import com.uber.jaeger.crossdock.api.*;
 import com.uber.jaeger.reporters.InMemoryReporter;
 import com.uber.jaeger.samplers.ConstSampler;
 import io.opentracing.Tracer;
+import java.util.*;
```
vprithvi: This violates https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports

vprithvi: I'm not convinced that the `verification` step fails builds for style failures because it passed this build.

pavolloffay: > This violates https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports 

Then this is probably a bug in the plugin.

pavolloffay:
What about using https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/google_checks.xml directly through gradle checkstyle? We would include this file in repository, remove couple of unnecessary things (e.g. comment every public method).

pavolloffay:
It does not have an automatic format, but in my opinion that is not necessary.

vprithvi:
>What about using https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/google_checks.xml directly through gradle checkstyle? 

We were previously using checkstyle previously but chose the google formatter mainly because of it's formatting abilities and integration with IntelliJ/Eclipse. Let me pull down the diff and check why it's not catching the * import.

vprithvi:
An alternative would be to run the checkstyle plugin after the formatting plugin.
[google-java-formatter](https://github.com/google/google-java-format) seems to have no way of verifying style, while the [gradle plugin](https://github.com/sherter/google-java-format-gradle-plugin) does. I suspect what verify does is to simply run the formatter, and see if there is a diff between the formatted and unformatted style. I also don't expect a formatter to convert a wildcard import into a normal import. 
We should also look into [Spotless](https://github.com/diffplug/spotless/) to see if it meets our needs.

yurishkuro:
I can buy either suggestion, but after working with Go I do like having an auto-format, not just a style checker. Back at Morgan Stanley we used to have a formatter for Eclipse wired into the project setup, so that formatting has never been an issue.

pavolloffay:
```
./gradlew googleJavaFormat
./gradlew checkstyleMain
```

still find checkstyle issues. https://github.com/sherter/google-java-format-gradle-plugin is not reliable solution.

pavolloffay:
closing in favour of #149

PR #149, opened 2017-04-26
Google checkstyle with gradle checkstyle plugin

### LINKED ISSUE #147: Codify import ordering and enforce on build
It looks like the import order that we have doesn't really follow our style guide (google java style), nor is it enforced during the build.
This results in a lot of import statements moved around in diffs making reviewing them difficult.

pavolloffay:
Related to #148, resolves #147 

Use only standard gradle checkstyle plugin with google checkstyle definitions from https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/google_checks.xml. If we feel that some of the rules are too strict we can adjust/disable annoying ones. 

Note that auto formatting provided by `com.github.sherter.google-java-format` does not work very well it does not format code properly. 

This fails `./gradle test`. If we agree I will apply style changes to align current code base with google code style (there are tons of violation right now, imports, javadoc, no empty lines, naming conventions).

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/SpanContext.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/SpanContext.java
@@ -19,6 +19,7 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
+
 package com.uber.jaeger;
 
 import com.uber.jaeger.exceptions.EmptyTracerStateStringException;
@@ -32,33 +33,33 @@ public class SpanContext implements io.opentracing.SpanContext {
   protected static final byte flagSampled = 1;
   protected static final byte flagDebug = 2;
 
-  private final long traceID;
-  private final long spanID;
-  private final long parentID;
+  private final long traceId;
+  private final long spanId;
+  private final long parentId;
   private final byte flags;
   private final Map<String, String> baggage;
-  private final String debugID;
+  private final String debugId;
 
-  public SpanContext(long traceID, long spanID, long parentID, byte flags) {
-    this(traceID, spanID, parentID, flags, Collections.<String, String>emptyMap(), null);
+  public SpanContext(long traceId, long spanId, long parentId, byte flags) {
+    this(traceId, spanId, parentId, flags, Collections.<String, String>emptyMap(), null);
   }
 
   SpanContext(
-      long traceID,
-      long spanID,
-      long parentID,
+      long traceId,
+      long spanId,
+      long parentId,
       byte flags,
       Map<String, String> baggage,
-      String debugID) {
+      String debugId) {
     if (baggage == null) {
       throw new NullPointerException();
     }
-    this.traceID = traceID;
-    this.spanID = spanID;
-    this.parentID = parentID;
+    this.traceId = traceId;
+    this.spanId = spanId;
+    this.parentId = parentId;
     this.flags = flags;
     this.baggage = baggage;
-    this.debugID = debugID;
+    this.debugId = debugId;
   }
 
   @Override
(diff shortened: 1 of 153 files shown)


vprithvi:
Hey, I'm testing out the [Spotless](https://github.com/diffplug/spotless) plugin to see whether we can keep the formatter and checker unified. If Spotless doesn't work out, I'm all for what you suggest.

vprithvi:
I looked at the plugin. It provides a nice abstraction around any formatters that we might want to run, and uses the same form of checking as `com.github.sherter.google-java-format`. It also uses `google-java-formatter`, which seems to be based on a `eclipse-formatter`.
I don't see a good reason to use it. 

I mostly agree with your approach, in that we do the following on the gradle `check` task
- verifyGoogleFormat (if it handles any cases that checkstyle does not)
- checkstyle (with the google style guide)
- LicenseCheck

Having `verifyGoogleFormat` also ensures that the formatter output does not fail checkstyle.

pavolloffay:
@vprithvi I think I don't understand what you want. 

gradle checkstyle checks code better than `verifyGoogleFormat`  from `com.github.sherter.google-java-format`. We saw that `verifyGoogleFormat` is not reliable e.g. wildcard imports are not checked. Second having two code checks mechanisms does not seem right. 

LicenseCheck - this PR is only about code style, IIRC license check works well.

vprithvi:
I apologize for the delay and confusion. 
I don't see any cases that `verifyGoogleFormat` catches that check style does not, and agree with what you proposed earlier.

yurishkuro:
+1, let's go with this check.

pavolloffay:
Thanks, then I will proceed and apply code changes required to pass checkstyle. If there are too many changes I would like to split it into multiple PRs. First handle imports, and then javadoc etc.. 

It will be probably huge.

pavolloffay:
@yurishkuro @vprithvi could you please have a quick look. There will be breaking changes due to naming conventions.

jaeger-core/src/main/java/com/uber/jaeger/samplers/HttpSamplingManager.java
```diff
@@ -19,31 +19,29 @@
...
 import java.net.HttpURLConnection;
 import java.net.URL;
 import java.net.URLEncoder;
 import java.nio.charset.Charset;
-
 import lombok.ToString;
 
 @ToString
-public class HTTPSamplingManager implements SamplingManager {
+public class HttpSamplingManager implements SamplingManager {
```
pavolloffay: @yurishkuro @vprithvi example of breaking change.

pavolloffay:
@yurishkuro @vprithvi should I continue with this? Or do you want to scale it down to checkstyle for imports only?

yurishkuro:
I would like to finish setting this up, but it would be better to finish the thrift upgrade first, and then maybe do this in chunks, ie enable style checks incrementally instead of one enormous diff. Wdyt?

pavolloffay:
Yes, I want to do this in chunks.

Currently, I disabled checkstyle for tests and javadoc styles.  So this PRs handles only code style(imports, naming conventions, empty lines) in main.

pavolloffay:
This should be ready to go in. After this is merged we can handle checkstyle in tests and javadocs.

pavolloffay:
oh, I did wrong rebase, do not merge yet.

pavolloffay:
PR rebased.

jaeger-core/src/main/java/com/uber/jaeger/SpanContext.java
```diff
@@ -74,16 +75,16 @@ public String getBaggageItem(String key) {
     return this.baggage;
   }
 
-  public long getTraceID() {
-    return traceID;
+  public long getTraceId() {
```
yurishkuro: Can we keep the methods with uppercase ID as deprecated and disable stylecheck on them? We have internal libraries that I don't want to break by a non-essential renaming.

yurishkuro:
@pavolloffay I merged it to unblock you, could you please add back the all-caps ID methods to SpanContext as deprecated in another PR?

jaeger-core/src/main/java/com/uber/jaeger/SpanContext.java
pavolloffay: Sure, only these in `SpanContext` ?

yurishkuro: Those that had `s/ID/Id/` change. I think SpanContext was the only one important enough to keep backwards compatible.

pavolloffay:
@yurishkuro PR https://github.com/uber/jaeger-client-java/pull/157

PR #150, opened 2017-04-26
Bump gradle version to 3.5

pavolloffay:
Bump gradle wrapper version to 3.5

Files were generated using `./gradlew wrapper` (after moving wrapper plugin to global scope, it should not be applied for all projects).


build.gradle
```diff
@@ -74,7 +74,6 @@ subprojects {
         skipExistingHeaders true
         strictCheck true
         include "**/*.java"
-        excludes(["**/gen-java/**"])
```
yurishkuro: Is `excludes` not supported?

pavolloffay: It seems that it is supported https://github.com/hierynomus/license-gradle-plugin, however it does not work here. I have disabled license check for the whole jaeger-thrift module

PR #153, opened 2017-04-27
Add codecov ignores

vprithvi:
Also set precision to 2.


codecov.yml
```diff
@@ -0,0 +1,7 @@
+ignore:
+ - "jaeger-thrift"
+ - "jaeger-crossdock"
+ - "**/gen-java/**"
```
pavolloffay: IIRC this is only in `jaeger-thrift` which you are ignoring just above

PR #156, opened 2017-04-28
Fix immutable baggage when returning iterator from context

pavolloffay:


jaeger-core/src/main/java/com/uber/jaeger/SpanContext.java
```diff
@@ -64,7 +63,7 @@ public SpanContext(long traceID, long spanID, long parentID, byte flags) {
 
   @Override
   public Iterable<Map.Entry<String, String>> baggageItems() {
-    return baggage.entrySet();
+    return new HashMap<String, String>(baggage).entrySet();
```
yurishkuro: not quite sure where you're going with this. The main immutability of the context (and baggage by extension) comes from this line in the [setBaggageItem](https://github.com/uber/jaeger-client-java/blob/master/jaeger-core/src/main/java/com/uber/jaeger/Span.java#L135), i.e. the whole context is replaced when baggage is amended.

```java
this.context = this.context.withBaggageItem(key, value);
```

pavolloffay: Iterator in Java allows to remove an item from the underlying collection.

yurishkuro: ok, I am not too concerned with this method, it's rarely used anyway. The potential future improvement would be to use immutable map in the constructor.

yurishkuro:
can't merge due to a conflict

pavolloffay:
rebased

PR #157, opened 2017-05-03
SpanContext add deprecated **ID methods

pavolloffay:
Related too https://github.com/uber/jaeger-client-java/pull/149#issuecomment-298712037

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/SpanContext.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/SpanContext.java
@@ -177,4 +177,28 @@ public class SpanContext implements io.opentracing.SpanContext {
   String getDebugId() {
     return debugId;
   }
+
+  /**
+   * @deprecated use {@link SpanContext#getTraceId()} instead.
+   */
+  @Deprecated
+  public long getTraceID() {
+    return traceId;
+  }
+
+  /**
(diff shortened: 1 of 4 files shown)


pavolloffay:
crossdock is passing locally. Could you please restart CI?

jaeger-core/src/main/java/com/uber/jaeger/SpanContext.java
```diff
@@ -177,4 +177,44 @@ public static SpanContext withDebugId(String debugId) {
...
+   */
+  @Deprecated
+  public long getParentID() {
+    return parentId;
+  }
+
+  /**
+   * @deprecated use {@link SpanContext#getDebugId()} instead.
+   */
+  @Deprecated
```
yurishkuro: these last two are not needed, external dependencies are only on the trace/span id related fields

pavolloffay:
PR updated.

pavolloffay:
CI failed on a weird error.

adriancole:
So what impact does this have to folks using B3? Is there another way to
get the parent id? Or are you dropping support for B3? cc @felixbarny

yurishkuro:
@adriancole no impact on B3. The non-deprecated methods just have been renamed to ID->Id per Google style. I asked Pavol to add the xxID methods back for backwards compatibility.

adriancole:
cool. thanks for clarifying

PR #158, opened 2017-05-03
Checkstyle fix for tests

pavolloffay:


jaeger-crossdock/src/test/java/com/uber/jaeger/crossdock/resources/behavior/http/TraceBehaviorResourceTest.java
```diff
@@ -149,7 +149,7 @@ public void testJoinTraceHttp() throws Exception {
     TraceResponse traceResponse = resp.readEntity(TraceResponse.class);
 
     assertNotNull(traceResponse.getDownstream());
-    validateTraceResponse(traceResponse, expectedTraceId, expectedBaggage, 2);
+    validateTraceResponse(traceResponse, String.format("%x", span.context().getTraceId()), expectedBaggage, 2);
```
badiib: why not use expectedTraceId? It's used in two places.

pavolloffay: I don't see it used in two places. Only this one. Otherwise it would not compile.

Original code was breaking this rule: http://checkstyle.sourceforge.net/config_coding.html#VariableDeclarationUsageDistance

PR #161, opened 2017-05-03
Multiple references clarification

pavolloffay:
This should clarify how multiple references should be processed. Previous discussion: https://github.com/uber/jaeger-client-java/pull/142#discussion_r113798215

Jaeger thrift span model holds a list of references but also `long parenId`. This `parentId` should be used for `childOf` reference to save payload as in 99% of situations there is only one childOf reference. However, this adds a little bit of confusion to the model. 

E.g:
* there is one 'childOf' refs. `parentId` is used.
* there are multiple `childOf` refs. Which one should be used as `parentId`? Should be one treated in a spacial way as a preferred parent?

I propose that `parentId` should be used only for situations when there is only one `childOf` reference. in all other situations pass all references in the list of references to keep it semantically clear.

I would like to also clarify sampling, debug and traceId when there are multiple references (potentially from diffrent traces).

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/Span.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Span.java
@@ -57,7 +57,7 @@ public class Span implements io.opentracing.Span {
     this.startTimeNanoTicks = startTimeNanoTicks;
     this.computeDurationViaNanoTicks = computeDurationViaNanoTicks;
     this.tags = new HashMap<String, Object>();
-    this.references = new ArrayList<Reference>(references);
+    this.references = references != null ? new ArrayList<Reference>(references) : null;
 
     for (Map.Entry<String, Object> tag : tags.entrySet()) {
       setTagAsObject(tag.getKey(), tag.getValue());
@@ -79,9 +79,10 @@ public class Span implements io.opentracing.Span {
   }
 
   public List<Reference> getReferences() {
-    synchronized (this) {
-      return Collections.unmodifiableList(references);
+    if (references == null) {
+      return Collections.emptyList();
     }
+    return Collections.unmodifiableList(references);
   }
 
   public Map<String, Object> getTags() {
--- a/jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
@@ -171,9 +171,12 @@ public class Tracer implements io.opentracing.Tracer {
 
     private String operationName = null;
     private long startTimeMicroseconds;
-    private final List<Reference> references = new ArrayList<Reference>();
+    /**
+     * In 99% situations there is only one parent (childOf), so we do not want to allocate
+     * a collection of references.
+     */
+    private List<Reference> references = Collections.emptyList();
     private final Map<String, Object> tags = new HashMap<String, Object>();
-    private final Map<String, String> baggage = new HashMap<String, String>();
 
     SpanBuilder(String operationName) {
       this.operationName = operationName;
@@ -192,11 +195,25 @@ public class Tracer implements io.opentracing.Tracer {
     @Override
     public io.opentracing.Tracer.SpanBuilder addReference(
         String referenceType, io.opentracing.SpanContext referencedContext) {
-      if (referencedContext instanceof SpanContext
-          && (Utils.equals(referenceType, References.CHILD_OF)
-                      || Utils.equals(referenceType, References .FOLLOWS_FROM))) {
-        this.references.add(new Reference((SpanContext) referencedContext, referenceType));
-        this.baggage.putAll(((SpanContext)referencedContext).baggage());
+
+      if (!(referencedContext instanceof SpanContext)) {
+        return this;
+      }
+
+      // Jaeger thrift currently does not support other reference types
+      if (!References.CHILD_OF.equals(referenceType)
+          && !References.FOLLOWS_FROM.equals(referenceType)) {
+        return this;
+      }
+
+      if (references.isEmpty()) {
+        // Optimization for 99% situations, when there is only one parent
+        references = Collections.singletonList(new Reference((SpanContext) referencedContext, referenceType));
+      } else {
+        if (references.size() == 1) {
+          references = new ArrayList<Reference>(references);
+        }
+        references.add(new Reference((SpanContext) referencedContext, referenceType));
       }
 
       return this;
@@ -249,9 +266,31 @@ public class Tracer implements io.opentracing.Tracer {
       return new SpanContext(id, id, 0, flags);
     }
 
-    private SpanContext createChildContext(SpanContext preferredParent) {
+    private Map<String, String> createChildBaggage() {
+      Map<String, String> baggage = null;
+
+      // optimization for 99% use cases, when there is only one parent
+      if (references.size() == 1) {
+        return references.get(0).getSpanContext().baggage();
+      }
+
+      for (Reference reference: references) {
+        if (reference.getSpanContext().baggage() != null) {
+          if (baggage == null) {
+            baggage = new HashMap<String, String>();
+          }
+          baggage.putAll(reference.getSpanContext().baggage());
+        }
+      }
+
+      return baggage;
+    }
+
+    private SpanContext createChildContext() {
+      SpanContext preferredReference = preferredReference();
+
       if (isRpcServer()) {
-        if (preferredParent.isSampled()) {
+        if (isSampled()) {
           metrics.tracesJoinedSampled.inc(1);
         } else {
           metrics.tracesJoinedNotSampled.inc(1);
@@ -259,15 +298,17 @@ public class Tracer implements io.opentracing.Tracer {
 
         // Zipkin server compatibility
         if (zipkinSharedRpcSpan) {
-          return preferredParent;
+          return preferredReference;
         }
       }
+
       return new SpanContext(
-          preferredParent.getTraceId(),
+          preferredReference.getTraceId(),
           Utils.uniqueId(),
-          preferredParent.getSpanId(),
-          preferredParent.getFlags(),
-          this.baggage,
+          preferredReference.getSpanId(),
+          // should we do OR across passed references?
+          preferredReference.getFlags(),
+          createChildBaggage(),
           null);
     }
 
@@ -276,25 +317,48 @@ public class Tracer implements io.opentracing.Tracer {
       return Tags.SPAN_KIND_SERVER.equals(tags.get(Tags.SPAN_KIND.getKey()));
     }
 
-    @Override
-    public io.opentracing.Span start() {
-      // find preferredParent, it is first childOf
-      Reference preferredParent = references.isEmpty() ? null : references.get(0);
+    private SpanContext preferredReference() {
+      Reference preferredReference = references.get(0);
       for (Reference reference: references) {
+        // childOf takes precedence as a preferred parent
         if (References.CHILD_OF.equals(reference.getType())
-            && !References.CHILD_OF.equals(preferredParent.getType())) {
-          preferredParent = reference;
+            && !References.CHILD_OF.equals(preferredReference.getType())) {
+          preferredReference = reference;
           break;
         }
       }
+      return preferredReference.getSpanContext();
+    }
 
+    private boolean isSampled() {
+      if (references != null) {
+        for (Reference reference : references) {
+          if (reference.getSpanContext().isSampled()) {
+            return true;
+          }
+        }
+      }
+      return false;
+    }
+
+    private String debugId() {
+      if (references.size() == 1 && references.get(0).getSpanContext().isDebugIdContainerOnly()) {
+        return references.get(0).getSpanContext().getDebugId();
+      }
+      return null;
+    }
+
+    @Override
+    public io.opentracing.Span start() {
       SpanContext context;
-      if (preferredParent == null) {
+
+      String debugId = debugId();
+      if (references.isEmpty()) {
         context = createNewContext(null);
-      } else if (preferredParent.getSpanContext().isDebugIdContainerOnly()) {
-        context = createNewContext(preferredParent.getSpanContext().getDebugId());
+      } else if (debugId != null) {
+        context = createNewContext(debugId);
       } else {
-        context = createChildContext(preferredParent.getSpanContext());
+        context = createChildContext();
       }
 
       long startTimeNanoTicks = 0;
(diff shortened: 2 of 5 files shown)


jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -197,11 +202,31 @@ public void close() {
...
-        this.references.add(new Reference((SpanContext) referencedContext, referenceType));
-        this.baggage.putAll(((SpanContext)referencedContext).baggage());
+
+      if (!(referencedContext instanceof SpanContext)) {
+        return this;
+      }
+
+      // TODO UI currently does not support arbitrary references
+      if (Utils.equals(referenceType, References.CHILD_OF)
+          || Utils.equals(referenceType, References .FOLLOWS_FROM)) {
```
yurishkuro: why Util.equals instead of ` References.CHILD_OF.equals()`?

yurishkuro: nit: could negate the condition and `return this`, to decrease the nesting of the main code

jaeger-core/src/main/java/com/uber/jaeger/reporters/protocols/JaegerThriftSpanConverter.java
```diff
@@ -48,7 +48,7 @@ private JaegerThriftSpanConverter() {}
             context.getTraceId(),
             0, // TraceIdHigh is currently not supported
             context.getSpanId(),
-            context.getParentId(),
+            span.getParentContext() == null ? 0 : span.getParentContext().getSpanId(),
```
yurishkuro: not sure I follow this one. The SpanContext class still has the parentID field, and it's going to be used by injectors, so why do we need to change this? And if we don't, then it looks like we don't need  `span.getParentContext` to be public

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -281,25 +315,64 @@ boolean isRpcServer() {
...
+      }
+      return false;
+    }
+
+    private String debugId() {
+      if (parentContext != null && parentContext.isDebugIdContainerOnly()) {
+        return parentContext.getDebugId();
+      }
+
+      if (references != null) {
```
yurishkuro: I think this is over-complicating the use case. The "debug id" (and in the future baggage) is the ability of the tracer to accept some info from the inbound request when the request does not have an actual trace context in it, but the user is supplying debug-id or baggage as a break-glass measure. I guess it's theoretically possible that these break-glass snippets can come to instrumentation that uses FollowsFrom ref type for inbound span, for example a Kafka consumer, but perhaps we can think of that scenario when it actually becomes real. For now child-of is the only case where break-glass data can sneak through.

jaeger-core/src/main/java/com/uber/jaeger/Span.java
```diff
@@ -37,6 +37,7 @@
   private final Map<String, Object> tags;
   private long durationMicroseconds; // span durationMicroseconds
   private String operationName;
+  private final SpanContext parentContext;
```
yurishkuro: is it possible to keep this new field out of the Span and limit the impact on the SpanBuilder only?

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
pavolloffay: Utils.equals was used before, I thought it's preferred here.

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
pavolloffay: :+1: good to know, I wasn't sure how exactly it is used.

jaeger-core/src/main/java/com/uber/jaeger/reporters/protocols/JaegerThriftSpanConverter.java
pavolloffay: See the description of this PR. 

I would like to use `parentId` in thrift span only if there is one parent (`childOf`). In all other cases use a list of references. You are right I can set `parentId` to null if references are null (see changed code).

pavolloffay: The ideal would be if `parentId` could be set to null (currently it is `long`)

jaeger-core/src/main/java/com/uber/jaeger/Span.java
pavolloffay: It was possible, if references are null then it will use parentId from context.

pavolloffay:
Could we do the next release once this in?

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -276,25 +311,56 @@ boolean isRpcServer() {
...
-          preferredParent = reference;
+            && !References.CHILD_OF.equals(preferredReference.getType())) {
+          preferredReference = reference;
           break;
         }
       }
+      return preferredReference.getSpanContext();
+    }
+
+    private boolean isSampled() {
```
yurishkuro: again, all this complexity seems to be coming from unnecessary dichotomy between `parentContext` and `references`. If we define `parentContext` to be
  * the first CHILD_OF reference, or
  * the first reference of any type, or
  * null
then we never need to look at `references` for all other values like isSampled, getDebugId, etc.

The ONLY change we would need is
```
   if parentContext != null {
      excludePrimaryChildOf = references.filter(r => r.refContext() != parentContext || r.refType != CHILD_OF)
      if !excludePrimaryChildOf.isEmpty {
          span.references = excludePrimaryChildOf
      }
   } else {
      span.references = references
   }
```

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -192,11 +197,32 @@ public void close() {
...
+      // TODO UI currently does not support arbitrary references
+      if (!References.CHILD_OF.equals(referenceType)
+          && !References.FOLLOWS_FROM.equals(referenceType)) {
+        return this;
+      }
+
+      if (parentContext == null && references == null
+          && References.CHILD_OF.equals(referenceType)) {
+        parentContext = (SpanContext) referencedContext;
+      } else {
```
pavolloffay: This is not sufficient for situations when there is only one parent, so the baggage can be reused.

pavolloffay: Actually we don't have to allocate baggage at this method, SpanBuilder does not provide an access to baggage, therefore I will move it to start method. It should be cleaner.

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
pavolloffay: > the first reference of any type, or.

Then we loose reference type. It should be used only for childOf

jaeger-core/src/main/java/com/uber/jaeger/reporters/protocols/JaegerThriftSpanConverter.java
```diff
@@ -48,7 +48,7 @@ private JaegerThriftSpanConverter() {}
             context.getTraceId(),
             0, // TraceIdHigh is currently not supported
             context.getSpanId(),
-            context.getParentId(),
+            span.getReferences() == null ? 0 : span.context().getParentId(),
```
pavolloffay: > shouldn't this be simply context.getParentId() == 0 ? null : context.getParentId()?

`parentId` in thrift is long not an object. I'm proposing to set it to 0 when there are multiple parents.

pavolloffay: See tests in jaeger thrift

pavolloffay:
I did a major simplification in https://github.com/uber/jaeger-client-java/pull/161/commits/d912aa29a45d3b774932430abd03414b2e5622ad. In my eyes it's a little bit cryptic but it does not create a collection when there is only one parent.

jaeger-core/src/main/java/com/uber/jaeger/SpanContext.java
```diff
@@ -35,6 +35,7 @@
 
   private final long traceId;
   private final long spanId;
+  // for zipkin compatibility
```
yurishkuro: this feels out of place. If we want to elaborate on the need for parentId, I'd move it to the class comment and make it more descriptive.

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -249,9 +265,35 @@ private SpanContext createNewContext(String debugId) {
...
+          }
+          baggage.putAll(reference.getSpanContext().baggage());
+        }
+      }
+
+      return baggage;
+    }
+
+    private SpanContext createChildContext() {
+      if (references.isEmpty()) {
```
yurishkuro: nit: I think we can do away with this check, as well as the similar check in preferredParent(), because both of these methods are only called by start() when `!references.isEmpty()`. And if by accident it gets called otherwise, it will throw exception from `references.get(0);` in preferredParent() anyway.

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -262,12 +304,15 @@ private SpanContext createChildContext(SpanContext preferredParent) {
           return preferredParent;
         }
       }
+
       return new SpanContext(
           preferredParent.getTraceId(),
           Utils.uniqueId(),
+          // for zipkin compatibility
```
yurishkuro: Comment is not accurate. We capture parent ID not because of zipkin, but as storage optimization for 99% use case of a plain parent-child relationship. Similar to my comment above, I think the explanation is best placed in the SpanContext class comment.

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -192,11 +195,24 @@ public void close() {
...
-          && (Utils.equals(referenceType, References.CHILD_OF)
-                      || Utils.equals(referenceType, References .FOLLOWS_FROM))) {
-        this.references.add(new Reference((SpanContext) referencedContext, referenceType));
-        this.baggage.putAll(((SpanContext)referencedContext).baggage());
+
+      if (!(referencedContext instanceof SpanContext)) {
+        return this;
+      }
+
+      // TODO UI currently does not support arbitrary references
```
yurishkuro: do we need this TODO? It's not actionable, and It's not just the UI, the Thrift model doesn't support other ref types: https://github.com/uber/jaeger-idl/blob/master/thrift/jaeger.thrift#L43

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -192,11 +195,24 @@ public void close() {
...
+      if (!References.CHILD_OF.equals(referenceType)
+          && !References.FOLLOWS_FROM.equals(referenceType)) {
+        return this;
+      }
+
+      if (references.isEmpty()) {
+        references = Collections.singletonList(new Reference((SpanContext) referencedContext, referenceType));
```
yurishkuro: I'd add a comment that this optimizes for 99% use case of a single parent.

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -276,25 +321,51 @@ boolean isRpcServer() {
...
+        return null;
+      }
+
+      Reference preferredReference = references.get(0);
       for (Reference reference: references) {
         if (References.CHILD_OF.equals(reference.getType())
-            && !References.CHILD_OF.equals(preferredParent.getType())) {
-          preferredParent = reference;
+            && !References.CHILD_OF.equals(preferredReference.getType())) {
+          preferredReference = reference;
```
yurishkuro: nit: add comment "CHILD_OF reference takes precedence as preferred parent"

jaeger-core/src/test/java/com/uber/jaeger/reporters/protocols/JaegerThriftSpanConverterTest.java
```diff
@@ -162,4 +167,60 @@ public void testTruncateString() {
...
+    Span parent = tracer.buildSpan("foo").start();
+
+    Span child = tracer.buildSpan("foo")
+        .asChildOf(parent)
+        .start();
+
+    com.uber.jaeger.thriftjava.Span span = JaegerThriftSpanConverter.convertSpan((com.uber.jaeger.Span) child);
+
+    assertEquals(((com.uber.jaeger.SpanContext)child.context()).getParentId(), span.getParentSpanId());
+    assertEquals(0, span.getReferences().size());
```
yurishkuro: nice! 👍

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -308,8 +379,8 @@ boolean isRpcServer() {
         }
       }
 
-      // TODO add tracer tags to zipkin first span in process
-      if (zipkinSharedRpcSpan && (preferredParent == null || isRpcServer())) {
+      // TODO, this adds tracer tags to zipkin first span in a process
```
yurishkuro: what specifically do we want TODO?

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
pavolloffay: Well, this was based on the previous comment that UI does not support user defined reference types. Will remove.

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
pavolloffay: https://github.com/uber/jaeger-client-java/pull/142#discussion_r112581943

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
pavolloffay: Agree, this comment should be in injector/extractor.

pavolloffay:
PR updated

pavolloffay:
rebased

pavolloffay:
@yurishkuro could you please do a release with these changes?

yurishkuro:
let's wait until after the OSCON workshop, since it changes the default port and our instructions for running Docker image only mention exposing zipkin.thrift port.

We need to prep the changelog for the release.

PR #168, opened 2017-05-04
Expand exception logs

pavolloffay:


jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -389,6 +397,17 @@ public Builder withZipkinSharedRpcSpan() {
       return this;
     }
 
+    /**
+     * If the exception is logged and 'message', 'stack' or 'error.kind' is missing. Tracer will
```
pavolloffay: `message`, `error.kind` could be done automatically,  and `stack` could be optional for perf reasons.

jaeger-core/src/main/java/com/uber/jaeger/Span.java
```diff
@@ -259,4 +271,45 @@ public Span log(long timestampMicroseconds, String message, /* @Nullable */ Obje
...
   }
+
+  /**
+   * Creates logs related to logged exception
+   *
+   * @param fields map containing exception logs which are not present in fields
+   * @return logged fields
+   */
+  private static Map<String, Object> exceptionLogs(Map<String, ?> fields) {
+    Object ex = fields.get("error.object");
```
black-adder: are "error.object", "stack", "message", and "error.kind" defined in opentracing Java? If not, should we make them public consts for jaeger-client-java?

jaeger-core/src/main/java/com/uber/jaeger/Span.java
```diff
@@ -223,12 +225,22 @@ public Span log(long timestampMicroseconds, Map<String, ?> fields) {
         if (logs == null) {
           this.logs = new ArrayList<LogData>();
         }
+
+        if (tracer.expandExceptionLogs()) {
+          Map<String, Object> errorFields = exceptionLogs(fields);
+          if (!errorFields.isEmpty()) {
+            errorFields.putAll(fields);
+          }
+          fields = errorFields;
```
black-adder: it looks like if the fields don't contain "error.object", we throw out all the fields the user passed in?

jaeger-core/src/main/java/com/uber/jaeger/Span.java
```diff
@@ -223,12 +225,22 @@ public Span log(long timestampMicroseconds, Map<String, ?> fields) {
...
+        }
+
         logs.add(new LogData(timestampMicroseconds, fields));
       }
       return this;
     }
   }
 
+
```
black-adder: nit: double newline uncessecary

jaeger-core/src/main/java/com/uber/jaeger/Span.java
pavolloffay: good catch, this should be inside, the if above

jaeger-core/src/main/java/com/uber/jaeger/Span.java
pavolloffay: They are defined in the spec, but not in API artifact, I will open PR to ot-java.

pavolloffay:
PR updated, added some tests

jaeger-core/src/main/java/com/uber/jaeger/Span.java
```diff
@@ -223,6 +225,15 @@ public Span log(long timestampMicroseconds, Map<String, ?> fields) {
         if (logs == null) {
           this.logs = new ArrayList<LogData>();
         }
+
+        if (tracer.expandExceptionLogs()) {
+          Map<String, Object> errorFields = exceptionLogs(fields);
+          if (!errorFields.isEmpty()) {
+            errorFields.putAll(fields);
+            fields = errorFields;
```
black-adder: I'm not overly familiar with certain Java things, but given that we do `errorFields.putAll(fields)` above, is it still necessary for `fields = errorFields;`?

black-adder:
LGTM, I wonder if we should wait until the tags are specified in ot-java before landing this though.

jaeger-core/src/main/java/com/uber/jaeger/Span.java
pavolloffay: It is necessary because of line 237.

black-adder: ohhh, putAll puts fields into errorFields and not the other way round (duhhh).

Given that, this means if the user specifically added a "error.object", "stack", "message", and "error.kind", that will overwrite the k,v in errorFields? sounds right.

pavolloffay: Tracer does not change explicitly logged items ( "error.object", "stack", "message" ..). It only derives new fields if are not logged.

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
yurishkuro: Nit: comma after "missing"

jpkrohling:
Is this still relevant ?

pavolloffay:
Yes I will rebase it

pavolloffay:
Rebased, it can be reviewed. The last review comment was to wait for `Fields` to be defined in OT. They have been added in 0.31.0

jaeger-core/src/main/java/com/uber/jaeger/Span.java
```diff
@@ -228,6 +231,13 @@ public Span log(long timestampMicroseconds, Map<String, ?> fields) {
         return this;
       }
       if (context.isSampled()) {
+        if (tracer.isExpandExceptionLogs()) {
+          Map<String, Object> errorFields = exceptionLogs(fields);
```
yurishkuro: minor point, but couldn't we move all the logic into exceptionLogs() so that here we only have
```
fields = exceptionLogs(fields);
```
?

Internally it could do `Map<String, Object> errorFields = new HashMap<String, Object>(fields);`

pavolloffay: I have changed it to  ^^. I will merge or green

pavolloffay:
Coverage is odd 
screenshot of compare jaegertracing_jaeger-client-java

PR #177, opened 2017-05-11
[jaeger-core:test] testRemoteReporterReport failing randomly

jkandasa:
* Test case `testRemoteReporterReport` failing randomly due to incorrect thread sleep. I have added actual `flushInterval` to sleep before to validate span entry on `sender` object.
* Added fix on  `com.uber.jaeger.reporters.RemoteReporterTest`: Incorrect `assertEquals` parameters makes confusion. Current test failure reports as `java.lang.AssertionError: expected:<0> but was:<1>`.


jaeger-core/src/test/java/com/uber/jaeger/reporters/RemoteReporterTest.java
```diff
@@ -62,10 +62,11 @@ public void setUp() throws Exception {
   public void testRemoteReporterReport() throws Exception {
     Span span = (Span) tracer.buildSpan("raza").start();
     reporter.report(span);
-    Thread.sleep(5);
+    // do sleep until automatic flush happens on 'reporter'
+    Thread.sleep(flushInterval + 20); // added 20ms on top of 'flushInterval' to avoid corner cases
```
pavolloffay: There shouldn't be sleep statements at all. This does not scale well. 

You know what the expected result is, therefore you can use something like https://github.com/awaitility/awaitility

pavolloffay:
Here are all the sleep statements
```
jaeger-core/src/test/java/com/uber/jaeger/reporters/RemoteReporterTest.java:65:    Thread.sleep(5);
jaeger-core/src/test/java/com/uber/jaeger/reporters/protocols/TestTServer.java:78:      Thread.sleep(1);
jaeger-core/src/test/java/com/uber/jaeger/samplers/RemoteControlledSamplerTest.java:119:    Thread.sleep(20);
jaeger-core/src/test/java/com/uber/jaeger/senders/UdpSenderTest.java:67:    Thread.sleep(5);
```

jaeger-core/src/test/java/com/uber/jaeger/reporters/RemoteReporterTest.java
jkandasa: @pavolloffay Thank you for the review. I will update

vprithvi: Another option is to use JUnit 5's [assertTimeout](https://howtoprogram.xyz/2016/12/01/junit-5-timeout-test/). However, this depends on moving the project to Java 8, and updating the JUnit dependency.

yurishkuro: in Go we just write a loop that checks & breaks or sleeps. Then after the loop we check again via assert.

jkandasa:
@pavolloffay I do not know, what to check on these files with `await`. Could you please guide me?
```
jaeger-core/src/test/java/com/uber/jaeger/reporters/protocols/TestTServer.java:78:      Thread.sleep(1);
jaeger-core/src/test/java/com/uber/jaeger/samplers/RemoteControlledSamplerTest.java:119:    Thread.sleep(20);
```

jaeger-core/src/test/java/com/uber/jaeger/senders/UdpSenderTest.java
```diff
@@ -64,7 +68,9 @@ private TestTServer startServer() throws Exception {
     Thread t = new Thread(server);
     t.start();
 
-    Thread.sleep(5);
+    // wait up to 5000ms to get this thread started
+    await().with().pollDelay(5, TimeUnit.MILLISECONDS).pollInterval(5, TimeUnit.MILLISECONDS).atMost(
+            5000, TimeUnit.MILLISECONDS).until(() -> t.isAlive());
```
yurishkuro: does it matter if we use `pollDelay`? I'd prefer to keep it as simple as possible. Also in Go code we typically poll every 1ms, for up to 1sec (here you have 5sec, which is fine too).

nit: can you use `atMost(5, TimeUnit.SECOND)` instead of ms?

jaeger-core/src/test/java/com/uber/jaeger/reporters/RemoteReporterTest.java
```diff
@@ -68,10 +72,13 @@ public void setUp() throws Exception {
   public void testRemoteReporterReport() throws Exception {
     Span span = (Span) tracer.buildSpan("raza").start();
     reporter.report(span);
-    Thread.sleep(50);
+    // do sleep until automatic flush happens on 'reporter'
+    // added 20ms on top of 'flushInterval' to avoid corner cases
+    await().with().pollInterval(50, TimeUnit.MILLISECONDS).atMost(
+            flushInterval + 20, TimeUnit.MILLISECONDS).until(() -> sender.getReceived().size() > 0);
```
yurishkuro: the flush interval at L53 is 1000ms, which is too long for the tests. But here the previous test was only sleeping for 50ms, so it's not clear to me how it was working to begin with.

Per comment below, the poll interval can be 1ms. Basically, the goal is to speed up the test as much as possible, while keeping it reliable.

jaeger-core/src/test/java/com/uber/jaeger/senders/UdpSenderTest.java
jkandasa: @yurishkuro Sure, I will update

jaeger-core/src/test/java/com/uber/jaeger/reporters/RemoteReporterTest.java
jkandasa: @yurishkuro as `flushInterval`is 1000ms. flush might happen on any ms in the window of 1000ms, for corner case I have added another 20ms. Here I can reduce `pollInterval`.

jaeger-core/src/test/java/com/uber/jaeger/reporters/RemoteReporterTest.java
```diff
@@ -74,8 +74,10 @@ public void testRemoteReporterReport() throws Exception {
     reporter.report(span);
     // do sleep until automatic flush happens on 'reporter'
     // added 20ms on top of 'flushInterval' to avoid corner cases
-    await().with().pollInterval(50, TimeUnit.MILLISECONDS).atMost(
-            flushInterval + 20, TimeUnit.MILLISECONDS).until(() -> sender.getReceived().size() > 0);
+    await()
+        .with()
+        .pollInterval(10, TimeUnit.MILLISECONDS)
```
jkandasa: @yurishkuro There is no specific reason for 10ms. moved to 1ms

jaeger-core/src/test/java/com/uber/jaeger/reporters/RemoteReporterTest.java
```diff
@@ -74,8 +74,10 @@ public void testRemoteReporterReport() throws Exception {
...
+        .atMost(flushInterval + 20, TimeUnit.MILLISECONDS).until(() -> sender.getReceived().size() > 0);
```
jkandasa: @yurishkuro Thank you updated

jaeger-core/src/test/java/com/uber/jaeger/senders/UdpSenderTest.java
```diff
@@ -68,9 +68,11 @@ private TestTServer startServer() throws Exception {
     Thread t = new Thread(server);
     t.start();
 
-    // wait up to 5000ms to get this thread started
-    await().with().pollDelay(5, TimeUnit.MILLISECONDS).pollInterval(5, TimeUnit.MILLISECONDS).atMost(
-            5000, TimeUnit.MILLISECONDS).until(() -> t.isAlive());
+    // wait up to 5 seconds to get this thread started
+    await()
+        .with()
+        .pollInterval(5, TimeUnit.MILLISECONDS)
```
jkandasa: @yurishkuro There is no specific reason for 5ms. moved to 1ms

jaeger-core/src/test/java/com/uber/jaeger/senders/UdpSenderTest.java
```diff
@@ -68,9 +68,11 @@ private TestTServer startServer() throws Exception {
...
+        .atMost(5, TimeUnit.SECONDS).until(() -> t.isAlive());
```
jkandasa: @yurishkuro Thank you! Updated

yurishkuro:
Fixes #184

yurishkuro:
Actually, does it fix 184 or something else?

jkandasa:
@yurishkuro this PR fixes portion of https://github.com/uber/jaeger-client-java/pull/180 and doesn't fix https://github.com/uber/jaeger-client-java/issues/184

PR #180, opened 2017-05-13
Handle full queue in RemoteReporter

### LINKED ISSUE #116: Error handing missing for RemoteReporter TimerTask
In the following code for TimerTask, ArrayBlockingQueue could throw exceptions such as IllegalStateException and so on but the code is not catching and logging these errors properly. Code should catch and log proper exceptions via slf4j logger and may be capture this error in some metrics.

https://github.com/uber/jaeger-client-java/blob/ebda740068a0568650f321f56db6c2b658cb0e99/jaeger-core/src/main/java/com/uber/jaeger/reporters/RemoteReporter.java#L72

Exception in thread "Timer-3" java.lang.IllegalStateException: Queue full
at java.util.AbstractQueue.add(AbstractQueue.java:98)
at java.util.concurrent.ArrayBlockingQueue.add(ArrayBlockingQueue.java:312)
at com.uber.jaeger.reporters.RemoteReporter$1.run(RemoteReporter.java:73)
at java.util.TimerThread.mainLoop(Timer.java:555)
at java.util.TimerThread.run(Timer.java:505)

### LINKED ISSUE #178: RemoteReporter throws exception when commandQueue is full
RemoteReporter uses ArrayBlockingQueue to store commands, but adding new commands is done with `ArrayBlockingQueue.add(x)` method which:

>Inserts the specified element into this queue if it is possible to do so immediately without violating capacity restrictions, returning true upon success and throwing an IllegalStateException if no space is currently available. When using a capacity-restricted queue, it is generally preferable to use offer.

Right now `RemoteReporter.report` checks if queue is full but it is not thread-safe.

It would be better to avoid `IllegalStateException` by using `RemoteReporter.offer(x)`, and if it fails (returns false) increment `reporterDropped` metric.

New version:
```
  public void report(Span span) {
    // Its better to drop spans, than to block here
    boolean added = commandQueue.offer(new AppendCommand(span));
    if(!added) {
        metrics.reporterDropped.inc(1);
    }
  }
```
The other `adds` should also `offer` instead.

mabn:
This PR:
- Fixes #178 (RemoteReporter throws exception when commandQueue is full)
- Fixes #116 (Error handing missing for RemoteReporter TimerTask)

Currently when RemoteReporter's queue is full all 3 operations (append, close, flush) throw the exception:
```
java.lang.IllegalStateException: Queue full
at java.util.AbstractQueue.add(AbstractQueue.java:98)
at java.util.concurrent.ArrayBlockingQueue.add(ArrayBlockingQueue.java:312)
```

I've changed all 3 to use `offer` instead which does not block/throw. It's a clearly good choice for `append`and `flush`, but I'm not sure if in case of `close` this is the best approach. I made `close` wait for a short time (100ms) and then just ignore the `QueueProcessor` thread, but maybe it should throw to indicate that it failed? Or block indefinitely?

There's one more issue I just spotted - updating "reporterQueueLength" metric happens inside `FlushCommand` - but when the queue is full this command is almost never queued. E.g. when it has compete for a flow with 99 `AppendCommand`s then it will run only once in 100 attempts.
This would explain the weird plot I have when max size was set to 100
Alt text

To fix that I moved `reporterQueueLength` update to the `flush()` method triggered by the `TimerTask`


jaeger-core/src/test/java/com/uber/jaeger/reporters/RemoteReporterTest.java
```diff
@@ -62,7 +68,7 @@ public void setUp() throws Exception {
   public void testRemoteReporterReport() throws Exception {
     Span span = (Span) tracer.buildSpan("raza").start();
     reporter.report(span);
-    Thread.sleep(5);
+    Thread.sleep(50);
```
mabn: I increased it because it was failing every time on my laptop. There's already a PR to make it right: https://github.com/uber/jaeger-client-java/pull/177

jaeger-core/src/test/java/com/uber/jaeger/reporters/InMemorySender.java
```diff
@@ -28,11 +28,18 @@
 import com.uber.jaeger.thriftjava.Span;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.Semaphore;
 
+/**
+ * InMemorySender "appends" span only earlier "append" was permitted. Appending is a blocking
```
yurishkuro: did you mean "only if earlier"?

jaeger-core/src/main/java/com/uber/jaeger/reporters/RemoteReporter.java
```diff
@@ -78,20 +80,31 @@ public void run() {
...
   }
 
   @Override
   public void close() {
-    commandQueue.add(new CloseCommand());
+    boolean closeCommandAdded = false;
+    try {
+      // best-effort: if we can't add CloseCommand in this time then it probably will never happen
+      closeCommandAdded = commandQueue
+          .offer(new CloseCommand(), CLOSE_ENQUEUE_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS);
```
yurishkuro: makes sense. Only I would increase the timeout to `1sec`.

jaeger-core/src/test/java/com/uber/jaeger/reporters/InMemorySender.java
```diff
@@ -73,4 +85,18 @@ public int flush() throws SenderException {
   public int close() throws SenderException {
     return flush();
   }
+
+  /**
+   * InMemorySender "appends" span only if the "append" was permitted. Otherwise append blocks
+   * until it it permitted.
```
yurishkuro: this comment not needed here

jaeger-core/src/test/java/com/uber/jaeger/reporters/RemoteReporterTest.java
```diff
@@ -103,4 +109,97 @@ public void testRemoteReporterFlushTimerThread() throws Exception {
...
+  public void testAppendWhenQueueFull() {
+    // change sender to blocking mode
+    sender.permitAppend(0);
+
+    for (int i = 0; i < maxQueueSize; i++) {
+      reporter.report(newSpan());
+    }
+
+    // When: at this point the queue is full or there is one slot empty (if the worker thread has
+    // already picked up some command). We add two spans to make sure that we overfill the queue
```
yurishkuro: if the workers are still reading from the queue, then it looks like you have a race condition and no guarantee that the next two commands would actually cause a queue overflow

jaeger-core/src/test/java/com/uber/jaeger/reporters/InMemorySender.java
mabn: Actually I'd rather change it to:
> Sender which stores spans in memory. Appending a new span is a blocking operation unless "permitted". By default Integer.MAX_VALUE "appends" are permitted.

jaeger-core/src/test/java/com/uber/jaeger/reporters/RemoteReporterTest.java
mabn: That's why I changed InMemorySender to become blocking:
```
// change sender to blocking mode
sender.permitAppend(0);
```
Thanks to this nothing is processing the queue - QueueProcessor picks only 1 AppendCommand and blocks on the append call.
So this test:
1. adds maxQueueSize spans to the queue
2. QueueProcessor takes 1 commands and blocks
3. adds additional span to make the queue full
4. adds one more span and it is dropped without throwing exception

Point 2 sometimes might not happen (due to scheduling) - then both span from points 3) and 4) are dropped.

I tried to come with some simple way of testing various cases without possibility of race conditions - adding that semaphore/permits to InMemorySender was the simplest solution I could find.

jaeger-core/src/main/java/com/uber/jaeger/reporters/RemoteReporter.java
mabn: done

jaeger-core/src/main/java/com/uber/jaeger/reporters/RemoteReporter.java
```diff
@@ -78,20 +80,31 @@ public void run() {
...
     }
-
-    commandQueue.add(new AppendCommand(span));
   }
 
   @Override
   public void close() {
-    commandQueue.add(new CloseCommand());
+    boolean closeCommandAdded = false;
+    try {
```
yurishkuro: was just having a second look - why do we need a separate try/catch? Seems like you can just call `offer` in L102

mabn: You're right - removed

yurishkuro:
Thanks, @mabn

PR #185, opened 2017-05-17
Fix Flaky RemoteReporter test

### LINKED ISSUE #184: Flaky RemoteReporter test
https://travis-ci.org/uber/jaeger-client-java/jobs/232521027#L863

```
com.uber.jaeger.reporters.RemoteReporterTest > testFlushUpdatesQueueLength FAILED
    java.lang.AssertionError: expected:<0> but was:<1>
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.failNotEquals(Assert.java:834)
        at org.junit.Assert.assertEquals(Assert.java:645)
        at org.junit.Assert.assertEquals(Assert.java:631)
        at com.uber.jaeger.reporters.RemoteReporterTest.testFlushUpdatesQueueLength(RemoteReporterTest.java:213)
```

mabn:
Explicitly waiting for TimerTask to call flush() - it never failed for me, but I think that this is the reason why it failed on CI

Fixes #184

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/reporters/RemoteReporter.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/reporters/RemoteReporter.java
@@ -73,7 +73,7 @@ public class RemoteReporter implements Reporter {
             flush();
           }
         },
-        0,
+        flushInterval,
         flushInterval);
   }
 
--- a/jaeger-core/src/test/java/com/uber/jaeger/reporters/RemoteReporterTest.java
+++ b/jaeger-core/src/test/java/com/uber/jaeger/reporters/RemoteReporterTest.java
@@ -214,7 +214,10 @@ public class RemoteReporterTest {
     // change sender to blocking mode
     sender.permitAppend(0);
     RemoteReporter remoteReporter = (RemoteReporter) reporter;
-    remoteReporter.flush();
+
+    await("flush() execution via TimerTask").atMost(5, TimeUnit.SECONDS).until(
+        () -> metricsReporter.gauges.get("jaeger.reporter-queue") != null
+    );
 
     for (int i = 0; i < 10; i++) {
       reporter.report(newSpan());
--- a/jaeger-core/src/test/java/com/uber/jaeger/reporters/RemoteReporterTest.java
+++ b/jaeger-core/src/test/java/com/uber/jaeger/reporters/RemoteReporterTest.java
@@ -213,7 +213,6 @@ public class RemoteReporterTest {
   public void testFlushUpdatesQueueLength() throws Exception {
     // change sender to blocking mode
     sender.permitAppend(0);
-    RemoteReporter remoteReporter = (RemoteReporter) reporter;
 
     await("flush() execution via TimerTask").atMost(5, TimeUnit.SECONDS).until(
         () -> metricsReporter.gauges.get("jaeger.reporter-queue") != null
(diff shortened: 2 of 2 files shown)


jaeger-core/src/test/java/com/uber/jaeger/reporters/RemoteReporterTest.java
```diff
@@ -213,15 +213,18 @@ public void testFlushWhenQueueFull() {
   public void testFlushUpdatesQueueLength() throws Exception {
     // change sender to blocking mode
     sender.permitAppend(0);
-    RemoteReporter remoteReporter = (RemoteReporter) reporter;
-    remoteReporter.flush();
+
+    await("flush() execution via TimerTask").atMost(5, TimeUnit.SECONDS).until(
+        () -> metricsReporter.gauges.get("jaeger.reporter-queue") != null
+    );
```
yurishkuro: I have a couple issues with this. First, because the flush interval is 1s, this test can block for up to a second. It's better to create a new instance of the remote reporter just for this test with a shorter interval.

Second, why wait for timer flush in the first place? That doesn't seem to be the intention of this test. I could see the test could be simpler:
  * use reporter with large flush interval to ensure timer flush won't happen in the test
  * block the sender
  * retrieve the current value of the queue length metric (null == 0)
  * append N spans (maybe N=3)
  * call `flush()` explicitly
  * verify that the gauge has a value of N or N-1

Is there a race condition in such algorithm?

mabn: TimerTask fires with 0 delay and `await()` waits just for that (until metric != null). It waits ~100ms because that's the default poll interval, that could be set to 5ms.

I wait for TimerTask to avoid having it call flush at some random time. Ideally I'd want to write just these:
1. add 10 spans
2. check that reporter-queue is still 0
3. flush()
4. check that it was updated and is no longer 0

But TimerTask first run (after ~0 delay) might happen between 1. and 2. and check in 2. would fail - I think that's why it is flaky right now (I'm not 100% sure, it doesn't fail for me). The same can happen in the scenario you described.
`awaiting` the first TimerTask run solves this.

I'll change that poll interval to 5ms and add only 3 spans. Let me know if I should rework it more.

Now, there is another test `testCloseWhenQueueFull` which takes 1 second - because it tests the case where sender blocks when reporter is closing and there's that 1sec `CLOSE_ENQUEUE_TIMEOUT_MILLIS`. I don't see an easy way to make that test fast other than by making this timeout configurable. Which might be a good idea...

yurishkuro: TBH, I don't know why we're executing flush immediately
```java
    flushTimer.schedule(
        new TimerTask() {
          @Override
          public void run() {
            flush();
          }
        },
        0,
        flushInterval);
  }
```

It's that `0` argument - I think we should remove it.

mabn: Makes sense - I did just that. I've also opened #187 to deal with slow `testCloseWhenQueueFull`

jaeger-core/src/main/java/com/uber/jaeger/reporters/RemoteReporter.java
```diff
@@ -73,7 +73,7 @@ public void run() {
             flush();
           }
         },
-        0,
+        flushInterval,
```
yurishkuro: I think we can simply drop this argh, isn't it the same?

mabn: The version of `schedule()` with just `delay` argument runs just once (after specified delay) - both are needed here.

yurishkuro:
Thanks!

PR #189, opened 2017-05-24
Save baggage in span

black-adder:
Fixes https://github.com/uber/jaeger/issues/179

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/Span.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Span.java
@@ -117,6 +117,7 @@ public class Span implements io.opentracing.Span {
   @Override
   public Span setBaggageItem(String key, String value) {
     synchronized (this) {
+      // TODO emit a metric whenever baggage is updated
       String prevItem = this.getBaggageItem(key);
       this.context = this.context.withBaggageItem(key, value);
       if (context.isSampled()) {
--- a/jaeger-core/src/main/java/com/uber/jaeger/Span.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Span.java
@@ -117,12 +117,16 @@ public class Span implements io.opentracing.Span {
   @Override
   public Span setBaggageItem(String key, String value) {
     synchronized (this) {
+      String prevItem = this.getBaggageItem(key);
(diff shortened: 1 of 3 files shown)


jaeger-core/src/test/java/com/uber/jaeger/SpanTest.java
```diff
@@ -307,15 +313,18 @@ public void testBaggageOneReference() {
   public void testBaggageMultipleReferences() {
     io.opentracing.Span parent1 = tracer.buildSpan("foo").start();
     parent1.setBaggageItem("foo", "bar");
+    this.assertBaggageLogs(parent1, "foo", "bar");
     io.opentracing.Span parent2 = tracer.buildSpan("foo").start();
     parent2.setBaggageItem("foo2", "bar");
+    this.assertBaggageLogs(parent2, "foo2", "bar");
```
yurishkuro: That makes me think, should we also be recording new vs override of the baggage? It cannot new reliably inferred ink the full trace because the parent might set the baggage after the child has been started.

black-adder:
Started a github project for it: https://github.com/uber/jaeger/projects/1

jaeger-core/src/test/java/com/uber/jaeger/SpanTest.java
black-adder: I don't follow

yurishkuro: iphone's autocorrect. I edited the question. When both parent and child spans set the same baggage key, we don't record if the child overrides the parent or not.

black-adder: I'll add fields.put("overwrite", True) to signify if we're doing a parent override.

jaeger-core/src/main/java/com/uber/jaeger/Span.java
```diff
@@ -117,7 +117,18 @@ public String getOperationName() {
   @Override
   public Span setBaggageItem(String key, String value) {
     synchronized (this) {
+      String prevItem = this.getBaggageItem(key);
       this.context = this.context.withBaggageItem(key, value);
```
yurishkuro: add TODO to emit a metric

PR #194, opened 2017-06-12
HTTP sender for jaeger.thrift

pavolloffay:
HTTP sender for jaeger.thrift.

resolves: https://github.com/uber/jaeger-client-java/issues/186

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
@@ -36,11 +36,10 @@ import com.uber.jaeger.samplers.ProbabilisticSampler;
 import com.uber.jaeger.samplers.RateLimitingSampler;
 import com.uber.jaeger.samplers.RemoteControlledSampler;
 import com.uber.jaeger.samplers.Sampler;
+import com.uber.jaeger.senders.Sender;
 import com.uber.jaeger.senders.UdpSender;
-
 import java.text.NumberFormat;
 import java.text.ParseException;
-
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
@@ -124,8 +123,8 @@ public class Configuration {
       String serviceName,
       SamplerConfiguration samplerConfig,
       ReporterConfiguration reporterConfig) {
-    if (serviceName == null || serviceName.length() == 0) {
-      throw new RuntimeException("Must provide a service name for Jaeger Configuration");
+    if (serviceName == null || serviceName.isEmpty()) {
+      throw new IllegalArgumentException("Must provide a service name for Jaeger Configuration");
     }
 
     this.serviceName = serviceName;
@@ -273,24 +272,23 @@ public class Configuration {
   }
 
   public static class ReporterConfiguration {
+    private static final int DEFAULT_FLUSH_INTERVAL_MS = 1000;
+    private static final int DEFAULT_MAX_QUEUE_SIZE = 100;
 
-    private static final String defaultAgentHost = "localhost";
-    private static final int defaultAgentPort = 5775;
-    private static final int defaultFlushIntervalMs = 1000;
-    private static final int defaultMaxQueueSize = 100;
+    private Boolean logSpans;
+    private String agentHost;
+    private Integer agentPort;
+    private Integer flushIntervalMs;
+    private Integer maxQueueSize;
 
-    private final Boolean logSpans;
-
-    private final String agentHost;
-
-    private final Integer agentPort;
-
-    private final Integer flushIntervalMs;
-
-    private final Integer maxQueueSize;
+    private Sender sender;
 
     public ReporterConfiguration() {
       this(null, null, null, null, null);
+    }
+
+    public ReporterConfiguration(Sender sender) {
+      this.sender = sender;
     }
 
     public ReporterConfiguration(
@@ -315,18 +313,20 @@ public class Configuration {
           getPropertyAsInt(JAEGER_REPORTER_MAX_QUEUE_SIZE));
     }
 
-    private Reporter getReporter(Metrics metrics) {
-      UdpSender sender =
+    private Sender getSender() {
+      return this.sender != null ? this.sender :
           new UdpSender(
-              stringOrDefault(this.agentHost, defaultAgentHost),
-              numberOrDefault(this.agentPort, defaultAgentPort).intValue(),
+              stringOrDefault(this.agentHost, UdpSender.DEFAULT_AGENT_UDP_HOST),
+              numberOrDefault(this.agentPort, UdpSender.DEFAULT_AGENT_UDP_COMPACT_PORT).intValue(),
               0 /* max packet size */);
+    }
 
+    private Reporter getReporter(Metrics metrics) {
       Reporter reporter =
           new RemoteReporter(
-              sender,
-              numberOrDefault(this.flushIntervalMs, defaultFlushIntervalMs).intValue(),
-              numberOrDefault(this.maxQueueSize, defaultMaxQueueSize).intValue(),
+              getSender(),
+              numberOrDefault(this.flushIntervalMs, DEFAULT_FLUSH_INTERVAL_MS).intValue(),
+              numberOrDefault(this.maxQueueSize, DEFAULT_MAX_QUEUE_SIZE).intValue(),
               metrics);
 
       if (Boolean.TRUE.equals(this.logSpans)) {
--- a/jaeger-core/src/main/java/com/uber/jaeger/senders/HttpSender.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/senders/HttpSender.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright (c) 2016, Uber Technologies, Inc
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.uber.jaeger.senders;
+
+import com.uber.jaeger.thriftjava.Batch;
+import com.uber.jaeger.thriftjava.Process;
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.List;
+import org.apache.http.HttpResponse;
+import org.apache.http.HttpStatus;
+import org.apache.http.client.HttpClient;
+import org.apache.http.client.methods.HttpPost;
+import org.apache.http.entity.ByteArrayEntity;
+import org.apache.http.impl.client.DefaultHttpClient;
+import org.apache.thrift.TException;
+import org.apache.thrift.TSerializer;
+import org.apache.thrift.protocol.TBinaryProtocol.Factory;
+import org.apache.thrift.protocol.TProtocolFactory;
+
+public class HttpSender extends ThriftSender {
+  private static final String HTTP_COLLECTOR_JAEGER_THRIFT_FORMAT_PARAM = "format=jaeger.thrift";
+  private static final TProtocolFactory PROTOCOL_FACTORY = new Factory();
+  private static final TSerializer SERIALIZER = new TSerializer(PROTOCOL_FACTORY);
+
+  private final HttpClient httpClient = new DefaultHttpClient();
+  private final URI collectorUri;
+
+  /**
+   * @param endpoint Jaeger REST endpoint consuming jaeger.thrift, e.g http://localhost:14268/api/traces
+   * @param maxPacketSize max packet size
+   */
+  public HttpSender(String endpoint, int maxPacketSize) {
+    super(PROTOCOL_FACTORY, maxPacketSize);
+    this.collectorUri = constructCollectorUri(endpoint);
+  }
+
+  @Override
+  public void send(Process process, List<com.uber.jaeger.thriftjava.Span> spans) throws TException {
+    Batch batch = new Batch(process, spans);
+    byte[] bytes = SERIALIZER.serialize(batch);
+
+    try {
+      HttpPost httpPost = new HttpPost(this.collectorUri);
+      httpPost.setEntity(new ByteArrayEntity(bytes));
+      HttpResponse response = httpClient.execute(httpPost);
+      if (response.getStatusLine().getStatusCode() != HttpStatus.SC_OK) {
+        throw new TException("Could not send " + spans.size() + " spans, response "
+            + response.getStatusLine().getStatusCode());
+      }
+    } catch (IOException e) {
+      throw new TException("Could not send " + spans.size() + ", spans", e);
+    }
+  }
+
+  private URI constructCollectorUri(String endpoint) {
+    try {
+      return new URI(String.format("%s?%s", endpoint, HTTP_COLLECTOR_JAEGER_THRIFT_FORMAT_PARAM));
+    } catch (URISyntaxException e) {
+      throw new IllegalArgumentException("Wrong collector host", e);
+    }
+  }
+}
--- a/jaeger-core/src/main/java/com/uber/jaeger/senders/ThriftSender.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/senders/ThriftSender.java
@@ -1,0 +1,135 @@
+/*
+ * Copyright (c) 2016, Uber Technologies, Inc
+ *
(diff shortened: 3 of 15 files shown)


pavolloffay:
I get 200 from HttpClient but traces are not stored :/

yurishkuro:
There is a crossdock module that is used for end-to-end testing of spans collection. I would recommend adding an env var to that module that would control whether the client should use a UDP sender or HTTP sender. Then we can have both modes tested via end to end test.

pavolloffay:
+1 I haven't worked on tests yet. 

When I'm using `THttpClient(url, new DefaultHttpClient())` I cannnot see any traces being stored, however server returns 200. When I create a direct http request with thrift converted to `byte[]` it's stored fine.

yurishkuro:
> I cannnot see any traces being stored, however server returns 200.

@vprithvi do we have logging if we cannot parse the HTTP request?

pavolloffay:
hmm when I print `fmt.Println(req.GetBatches())` after `tdes.Read(&req, bodyBytes);` it shows `[]`

vprithvi:
>@vprithvi do we have logging if we cannot parse the HTTP request?

We do, but it might also mean that the thrift deserializer is failing silently.

pavolloffay:
Now the `HttpSender` is directly using http client. I think the problem is that `Collector.Client#send_submitBatches()` calls `TServiceClient#sendBase()` which adds `TMessage` to the body and on the server side it fails to deserialize.

jaeger-core/src/main/java/com/uber/jaeger/senders/HttpSender.java
```diff
@@ -0,0 +1,85 @@
...
+
+  public HttpSender(String host, int maxPacketSize) {
+    super(PROTOCOL_FACTORY, maxPacketSize);
+    this.collectorUri = constructCollectorUri(host);
+    this.httpClient = new DefaultHttpClient();
+  }
+
+  @Override
+  public void send(Process process, List<com.uber.jaeger.thriftjava.Span> spans) throws TException {
+    submitBatches_args submitBatches =
```
pavolloffay: It would be cleaner if the server accepted `Batch` and not `submitBatches_args` which is basically a list of batches.

pavolloffay:
@yurishkuro I have a question about crossdock.. Do you mean add another axis (AXIS_SENDER=udp,tcp) ? It also requires to add it to the `jaeger-client-go` and other services in order to propagate that value. It also means that tracer will be changed for the the lifetime of crossdock test which I guess can cause problems.

jaeger-core/src/main/java/com/uber/jaeger/senders/HttpSender.java
yurishkuro: @vprithvi ?  Why do we need to accept _list of batches_?

yurishkuro:
> Do you mean add another axis (AXIS_SENDER=udp,tcp) ?

Not quite. We define the test with `BEHAVIOR_ENDTOEND=testdriver,services`, which means the "services" axis must be enumerated. That axis is currently defined as `AXIS_SERVICES=java`, where `java` refers to Java server using the UDP sender.

```
    java:
        build: .
        ports:
            - "8080-8082"
```

We can add another type of server

```
    java_http:
        build: .
        ports:
            - "8080-8082"
        environment:
            - SENDER=http
```

Then we change the crossdock java server to look at that env var when instantiating the tracer with different sender. Then if we change the axis to `AXIS_SERVICES=java, java_http` the end-to-end test will be run twice, against both servers.

We can also rename `java -> java_udp` and give it `SENDER=udp` to be explicit and not have any default for sender type when creating the tracer.

Side note: "services" is not a good name for the axis, `e2eserver` / `AXIS_E2ESERVER` would be better (`e2e_server` even better, but not sure if crossdock will recognize `AXIS_E2E_SERVER`).

jaeger-core/src/main/java/com/uber/jaeger/senders/HttpSender.java
pavolloffay: It refers to https://github.com/uber/jaeger/blob/master/cmd/collector/app/handler.go#L77 cc @badiib

yurishkuro: @pavolloffay we discussed internally, the potential use case for submitting a list of batches is from the agent (we're using tchannel today, but would like to move away from it). The suggestion is to declare a thrift type `BatchList` and send that. wdyt?

pavolloffay:
I have made some progress. Currently this is failing:
```
[endtoend] test_driver→ (services=java-http testdriver=test_driver) ⇒ Fail: could not retrieve traces from query service
✗ [endtoend] test_driver→ (services=java-udp testdriver=test_driver) ⇒ Fail: could not retrieve traces from query service
```
I'm not sure what should I change to make it work. Probably tracer's service name? When I change it here https://github.com/pavolloffay/jaeger-client-java/blob/dd026c7a12ca48dca5f2e0983a1b315a2ecdda7c/jaeger-crossdock/src/main/java/com/uber/jaeger/crossdock/resources/behavior/EndToEndBehavior.java#L46-L46 it fails.

yurishkuro:
Yeah, need to be careful with service names. I opened https://github.com/uber/jaeger/issues/205 to clarify this.

jaeger-core/src/main/java/com/uber/jaeger/senders/HttpSender.java
yurishkuro: @pavolloffay decided to go with your suggestion, just taking a single Batch. The PR https://github.com/uber/jaeger/pull/209 should merge tomorrow.

pavolloffay:
> Side note: "services" is not a good name for the axis, e2eserver / AXIS_E2ESERVER would be better (e2e_server even better, but not sure if crossdock will recognize AXIS_E2E_SERVER).

This requires changes in `test-driver`: 
https://github.com/pavolloffay/jaeger/blob/bb790e790a3850c66c7f8a4bf049379bce8c320f/crossdock/services/tracehandler.go#L41-L41

jaeger-core/src/main/java/com/uber/jaeger/senders/HttpSender.java
pavolloffay: If it is expected that http collector will receive data only from one tracer per request than single `Batch` makes this API better.

pavolloffay:
it's ready for review :octocat: . 

https://github.com/uber/jaeger/issues/205 should go first, then I will adjust this.

pavolloffay:
I little bit refactored xdock to make it more consistent.

pavolloffay:
PR updated and HTTP sender sends a single `Batch`. 

Could this make it to `0.20.0` release?

jaeger-core/src/main/java/com/uber/jaeger/senders/HttpSender.java
```diff
@@ -0,0 +1,87 @@
...
+import org.apache.http.client.methods.HttpPost;
+import org.apache.http.entity.ByteArrayEntity;
+import org.apache.http.impl.client.DefaultHttpClient;
+import org.apache.thrift.TException;
+import org.apache.thrift.TSerializer;
+import org.apache.thrift.protocol.TBinaryProtocol.Factory;
+import org.apache.thrift.protocol.TProtocolFactory;
+
+/**
+ * @author Pavol Loffay
```
black-adder: I prefer we don't do this

jaeger-core/src/main/java/com/uber/jaeger/senders/ThriftSender.java
```diff
@@ -0,0 +1,137 @@
...
+import org.apache.thrift.TBase;
+import org.apache.thrift.TException;
+import org.apache.thrift.protocol.TProtocolFactory;
+import org.apache.thrift.transport.AutoExpandingBufferWriteTransport;
+
+/**
+ * @author Pavol Loffay
+ */
+@ToString(exclude = {"spanBuffer", "memoryTransport"})
+public abstract class ThriftSender implements Sender {
```
black-adder: +1 good refactor

jaeger-crossdock/src/main/java/com/uber/jaeger/crossdock/JerseyServer.java
```diff
@@ -121,9 +126,33 @@ public Tracer getTracer() {
...
+
+    String serviceName = serviceNameFromEnv();
+
+    JerseyServer server = new JerseyServer("0.0.0.0:8081", serviceName,
+        Arrays.asList(new TraceBehaviorResource(),
+            new EndToEndBehaviorResource(new EndToEndBehavior("test_driver",
+                "crossdock-" + serviceName, senderFromEnv("test_driver"))),
+            new HealthResource()));
+
+    Builder tchannelBuilder = new Builder(serviceNameFromEnv());
```
black-adder: use serviceName

jaeger-crossdock/src/main/java/com/uber/jaeger/crossdock/JerseyServer.java
```diff
@@ -121,9 +126,33 @@ public Tracer getTracer() {
...
+    tchannelBuilder.setServerPort(8082);
+    new TChannelServer(tchannelBuilder, new TraceBehavior(), server.getTracer()).start();
+  }
+
+  private static Sender senderFromEnv(String jaegerHost) {
+    String senderEnvVar = System.getenv(Constants.ENV_PROP_SENDER_TYPE);
+    System.out.println(Constants.ENV_PROP_SENDER_TYPE + "=" + senderEnvVar);
```
black-adder: is this for debugging purposes?

jaeger-crossdock/build.gradle
```diff
@@ -34,3 +33,7 @@ jar {
         attributes 'Main-Class': 'com.uber.jaeger.crossdock.JerseyServer'
     }
 }
+
+animalsniffer {
```
black-adder: not sure what this is, is it required for this PR?

pavolloffay: xdock is using java8. But animalsniffer fails when you try to use lambda.

jaeger-crossdock/src/main/java/com/uber/jaeger/crossdock/JerseyServer.java
pavolloffay: Yes, I have removed it

pavolloffay:
PR updated.

> do we need to make any configuration changes so that users can configure the client to use the HttpSender?

Configuration by default uses UDP sender, however RemoteReporter accepts Sender interface so users can use HTTP impl.

> Can you cover the code that you wrote?

I have added only HTTP sender, which is tested in xdock. Do we want to add other integration test for it?

black-adder:
> Configuration by default uses UDP sender, however RemoteReporter accepts Sender interface so users can use HTTP impl.

Right, but won't it make sense to add `String senderType` to ReporterConfiguration so that the configuration can take care of initializing the correct Sender rather than have the user manually create the reporter?

> I have added only HTTP sender, which is tested in xdock. Do we want to add other integration test for it?

Looking at codecov, it looks like code coverage dropped for this PR by 1%. Whatever new code was added should be covered.

jaeger-core/src/main/java/com/uber/jaeger/senders/HttpSender.java
```diff
@@ -0,0 +1,84 @@
...
+import org.apache.http.entity.ByteArrayEntity;
+import org.apache.http.impl.client.DefaultHttpClient;
+import org.apache.thrift.TException;
+import org.apache.thrift.TSerializer;
+import org.apache.thrift.protocol.TBinaryProtocol.Factory;
+import org.apache.thrift.protocol.TProtocolFactory;
+
+public class HttpSender extends ThriftSender {
+  public static final int HTTP_COLLECTOR_PORT = 14268;
+  public static final String HTTP_COLLECTOR_URL_FORMAT = "http://%s:%d/api/traces?format=jaeger.thrift";
```
vprithvi: I suggest making the path/port configurable. Some collector deployments might be behind a proxy, and might not have an trivial way of controlling the path/port they are launched on.

jaeger-core/src/main/java/com/uber/jaeger/senders/HttpSender.java
```diff
@@ -0,0 +1,84 @@
...
+  public void send(Process process, List<com.uber.jaeger.thriftjava.Span> spans) throws TException {
+    Batch batch = new Batch(process, spans);
+    byte[] bytes = SERIALIZER.serialize(batch);
+
+    try {
+      HttpPost httpPost = new HttpPost(this.collectorUri);
+      httpPost.setEntity(new ByteArrayEntity(bytes));
+      HttpResponse response = httpClient.execute(httpPost);
+      if (response.getStatusLine().getStatusCode() != HttpStatus.SC_OK) {
+        throw new TException("Could not send " + spans.size() + ", spans");
```
vprithvi: I recommend adding the response status code and body to the exception to aid in debugging.

jaeger-core/src/main/java/com/uber/jaeger/senders/HttpSender.java
```diff
@@ -0,0 +1,84 @@
...
+    } catch (IOException e) {
+      throw new TException("Could not send " + spans.size() + ", spans", e);
+    }
+  }
+
+  private URI constructCollectorUri(String host) {
+    try {
+      return new URI(String.format(HTTP_COLLECTOR_URL_FORMAT, host, HTTP_COLLECTOR_PORT));
+    } catch (URISyntaxException e) {
+      throw new IllegalArgumentException("Wrong collector host");
```
vprithvi: nit: suggest wrapping the `URISyntaxException` to aid in troubleshooting when the host is incorrect.

jaeger-core/src/main/java/com/uber/jaeger/senders/ThriftSender.java
```diff
@@ -0,0 +1,134 @@
...
+      process = new Process(span.getTracer().getServiceName());
+      process.setTags(JaegerThriftSpanConverter.buildTags(span.getTracer().tags()));
+      processBytesSize = getSizeOfSerializedThrift(process);
+      byteBufferSize += processBytesSize;
+    }
+
+    com.uber.jaeger.thriftjava.Span thriftSpan = JaegerThriftSpanConverter.convertSpan(span);
+    int spanSize = getSizeOfSerializedThrift(thriftSpan);
+    if (spanSize > maxSpanBytes) {
+      throw new SenderException("ThriftSender received a span that was too large", null, 1);
```
vprithvi: nit: recommend adding the spanSize and maxSpanBytes to exception

jaeger-core/src/main/java/com/uber/jaeger/senders/HttpSender.java
pavolloffay: I will make sender to accept endpoint param e.g. `http://host:port/api/traces` and the sender will just add query param

pavolloffay:
> Right, but won't it make sense to add String senderType to ReporterConfiguration so that the configuration can take care of initializing the correct Sender rather than have the user manually create the reporter?

Why string senderType? Actually is there any requirement to have string sampler type? Configuration what we have right now is not the nicest thing, there are two ugly things: constructors like Foo(0,0,0,0) for default values and these string types things (currently only for sampler) .

> Looking at codecov, it looks like code coverage dropped for this PR by 1%. Whatever new code was added should be covered.

From the report it's not clear to me what should be covered. As I said this adds only HttpSender for which integration test has been added in crossdock. If you have something specific in mind let me know please.

jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
```diff
@@ -315,18 +313,20 @@ public static ReporterConfiguration fromEnv() {
...
 
-    private Reporter getReporter(Metrics metrics) {
-      UdpSender sender =
+    private Sender getSender() {
+      return this.sender != null ? this.sender :
           new UdpSender(
-              stringOrDefault(this.agentHost, defaultAgentHost),
-              numberOrDefault(this.agentPort, defaultAgentPort).intValue(),
+              stringOrDefault(this.agentHost, UdpSender.DEFAULT_AGENT_UDP_HOST),
+              numberOrDefault(this.agentPort, UdpSender.DEFAULT_AGENT_UDP_COMPACT_PORT).intValue(),
```
pavolloffay: This was using a wrong port before

jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
```diff
@@ -273,26 +272,25 @@ public String getManagerHostPort() {
...
+    private Integer maxQueueSize;
 
-    private final Integer maxQueueSize;
+    private Sender sender;
 
     public ReporterConfiguration() {
       this(null, null, null, null, null);
     }
 
+    public ReporterConfiguration(Sender sender) {
```
pavolloffay: @black-adder here is the constructor, if there is no need to create reporter with http sender from env props it does not make sense to add string type. We should clearly document Jaeger API, so users know what parameters to use.

black-adder: +1 I'm fine with this

pavolloffay:
PR rebased and squashed. Could somebody review? 

@black-adder about the tests for `HttpSender` do you want to create integration test which would start http server and check if it works, btw the same is doing xdock test proposed by @yurishkuro.

black-adder:
Integration tests would be nice but looking at the coverage, it looks like we don't have any unit tests for httpsender and udpsender after the refactor.

pavolloffay:
@black-adder I don't know where are you looking at coverage but there are unit tests for udpSender.

black-adder:
https://codecov.io/gh/uber/jaeger-client-java/compare/1d9715841e4a7235a1dd0be727bb360da5e23286...d8f122efddf4614976c06ab1fef57e2040c8f526/diff

pavolloffay:
It seems that jacoco does not work properly with inheritance..?

jaeger-core/src/main/java/com/uber/jaeger/senders/HttpSender.java
```diff
@@ -0,0 +1,84 @@
...
+public class HttpSender extends ThriftSender {
+  private static final String HTTP_COLLECTOR_JAEGER_THRIFT_FORMAT_PARAM = "format=jaeger.thrift";
+  private static final TProtocolFactory PROTOCOL_FACTORY = new Factory();
+  private static final TSerializer SERIALIZER = new TSerializer(PROTOCOL_FACTORY);
+
+  private final HttpClient httpClient = new DefaultHttpClient();
+  private final URI collectorUri;
+
+  /**
+   * @param endpoint Jaeger REST endpoint consuming jaeger.thrift, e.g http://localhost:14268/api/traces
```
pilhuhn: I think it would have been good not to require the local part of the endpoint

PR #197, opened 2017-06-19
Add Process IP to jaeger.thrift

black-adder:


jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -93,21 +93,19 @@ private Tracer(
...
     this.version = loadVersion();
 
     tags.put("jaeger.version", this.version);
     String hostname = getHostName();
     if (hostname != null) {
       tags.put("jaeger.hostname", hostname);
     }
+    int ipv4 = 0;
+    try {
+      tags.put("ip", InetAddress.getLocalHost().getHostAddress());
```
vprithvi: Could you use a standard tag like `PEER_HOST_IPV4`>

vprithvi: Also, I'd recommend putting `InetAddress.getLocalHost().getHostAddress()` into a variable because the host address might change due to DHCP.

black-adder: it's not a standard opentracing tag, I can add it to constants I guess with jaeger.version and jaeger.hostname

black-adder: so keep it as a variable and only commit to the ip only when we report? We report once a second, is it really that bad if for 1 second we report the wrong ip?

yurishkuro: We define tracer tags at initialization and never update them. If IP address changes, there's nothing we can do without refactoring. I don't think it's a big concerns.

jaeger-core/src/main/java/com/uber/jaeger/Constants.java
```diff
@@ -42,4 +42,19 @@
...
+
+  /**
+   * The name of the tag used to report client version.
+   */
+  public static final String JAEGER_CLIENT_VERSION_TAG_KEY = "jaeger.version";
+
+  /**
+   * The name used to report host name of the process.
+   */
+  public static final String TRACER_HOSTNAME_TAG_KEY = "jaeger.hostname";
```
yurishkuro: I think this should be just `hostname`. It needed to be `jaeger.xxx` when we were sending Zipkin format since it did not have the distinction between span and process tags.

black-adder: Checked the backend zipkin to jaeger converter to make sure all is kosher, looks good. Making the change.

yurishkuro: what we may want to do is when the tags are added to Zipkin spans, add a `tracer.` prefix to them

jaeger-core/src/test/java/com/uber/jaeger/TracerTagsTest.java
```diff
@@ -61,39 +61,25 @@ public TracerTagsTest(SpanType spanType, Map<String, Object> expectedTags) {
...
 
     Map<String, Object> rootTags = new HashMap<>();
-    rootTags.put("jaeger.version", tracer.getVersion());
-    rootTags.put("jaeger.hostname", hostname);
-    rootTags.put("tracer.tag.str", "y");
-    rootTags.put("tracer.tag.bool", true);
-    rootTags.put("tracer.tag.num", 1);
+    rootTags.put("jaeger.version", SENTINEL);
+    rootTags.put("jaeger.hostname", SENTINEL);
+    rootTags.put("tracer.tag.str", SENTINEL);
```
yurishkuro: Confused by this change. The SENTINEL comment says "sentinel value is used to mark tags that should *not* be present". So if you're setting `tracer.tag.str`, why would it not be present?

jaeger-core/src/main/java/com/uber/jaeger/Constants.java
black-adder: maybe a later PR? need to coordinate with a backend change

yurishkuro: sure. Although I don't know why you need a backend change. The idea here is that when people use Jaeger backend, they should be using jaeger.thrift reporters (udp or http). Only when they use Zipkin backend they would use zipkin.thrift reporter, where we convert tracer tags into first-in-process span tags.

jaeger-core/src/test/java/com/uber/jaeger/TracerTagsTest.java
black-adder: this is for jaeger.thrift, these tags will only be present in process.tags or else we'll be double reporting the same tags in both process.tags and tags.

jaeger-core/src/main/java/com/uber/jaeger/Constants.java
black-adder: nvm, I thought we were considering when the user uses zipkin reporter and the jaeger backend and we need to convert from zipkin to jaeger. I'll make the change.

jaeger-core/src/test/java/com/uber/jaeger/TracerTagsTest.java
yurishkuro: Well, I think this whole test was meant for `zipkin.thrift`, because it had the logic of only reporting the process tags in the first-in-process span. You're essentially changing everything to SENTINEL, meaning these tags are not going to be set on the span.

black-adder: I already have similar tests for this in zipkin.thrift. I'd argue this test still provides some value to ensure tags aren't double reported in jaeger.thrift; but I'll update this test to provide some better value (ie check the process.tags actually exist)

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -397,11 +398,6 @@ private String debugId() {
         }
       }
 
-      // TODO move this to jaeger-zipkin, this adds tracer tags to zipkin first span in a process
```
pavolloffay: Does this mean that tracer tags won't be added to Zipkin spans(root in a process)?

black-adder: It's still added, but I've moved that logic into ZipkinSpanConverter as per the TODO

jaeger-zipkin/src/main/java/com/uber/jaeger/senders/zipkin/ThriftSpanConverter.java
```diff
@@ -88,6 +89,19 @@
     Map<String, Object> tags = span.getTags();
     boolean isRpc = isRpc(span);
     boolean isClient = isRpcClient(span);
+    boolean firstSpanInProcess = span.getReferences().isEmpty() || isRpcServer(span);
+
+    if (firstSpanInProcess && span.getTracer().tags() != null) {
+      // add tracer tags to first zipkin span in a process but remove "ip" tag as it is
+      // taken care of separately.
+      for (String tagKey : span.getTracer().tags().keySet()) {
+        if (!tagKey.equals("ip")) {
```
pavolloffay: nit `TRACER_IP_TAG_KEY`?

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -93,21 +94,21 @@ private Tracer(
...
+    tags.put(Constants.JAEGER_CLIENT_VERSION_TAG_KEY, this.version);
     String hostname = getHostName();
     if (hostname != null) {
-      tags.put("jaeger.hostname", hostname);
+      tags.put(Constants.TRACER_HOSTNAME_TAG_KEY, hostname);
     }
+    int ipv4 = 0;
+    try {
+      tags.put(Constants.TRACER_IP_TAG_KEY, InetAddress.getLocalHost().getHostAddress());
+      ipv4 = Utils.ipToInt(Inet4Address.getLocalHost().getHostAddress());
```
yurishkuro: DRY: define local var for InetAddress.getLocalHost().getHostAddress()

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -93,21 +94,21 @@ private Tracer(
...
     this.version = loadVersion();
 
-    tags.put("jaeger.version", this.version);
+    tags.put(Constants.JAEGER_CLIENT_VERSION_TAG_KEY, this.version);
     String hostname = getHostName();
     if (hostname != null) {
-      tags.put("jaeger.hostname", hostname);
+      tags.put(Constants.TRACER_HOSTNAME_TAG_KEY, hostname);
     }
+    int ipv4 = 0;
```
yurishkuro: remove default value since you're setting to 0 in `catch`

jaeger-core/src/test/java/com/uber/jaeger/TracerTagsTest.java
```diff
@@ -133,5 +117,9 @@ public void testTracerTagsZipkin() throws Exception {
         assertEquals("Expecting " + key + " for " + spanType, expectedValue, actualValue);
       }
     }
+    Map<String, ?> processTags = tracer.tags();
+    assertEquals(true, processTags.containsKey("jaeger.version"));
+    assertEquals(true, processTags.containsKey("hostname"));
+    assertEquals(true, processTags.containsKey("tracer.tag.str"));
```
yurishkuro: This test still doesn't make sense to me. The reason it was parameterized was to ensure that tracer tags are converted into span tags when the span is first-in-process (2 out of 3 parameterized test cases). Since that logic has moved to Zipkin converter, this test should be moved there and operate only on the converted Zipkin span. The test that exists there is not as comprehensive as this one.

We may add a separate test here that tracer-level tags do not make it to Span tags, as well as that `withTag` method of the TracerBuilder works (which probably exists already elsewhere).

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
black-adder: one is InetAddress and the other is Inet4Address

jaeger-zipkin/src/main/java/com/uber/jaeger/senders/zipkin/ThriftSpanConverter.java
```diff
@@ -88,6 +90,20 @@
     Map<String, Object> tags = span.getTags();
     boolean isRpc = isRpc(span);
     boolean isClient = isRpcClient(span);
+    boolean firstSpanInProcess = span.getReferences().isEmpty() || isRpcServer(span);
+
+    Map<String, ?> processTags = span.getTracer().tags();
+    if (firstSpanInProcess && processTags != null) {
```
yurishkuro: tracer tags are never null: `this.tags = Collections.unmodifiableMap(tags);

jaeger-zipkin/src/test/java/com/uber/jaeger/senders/zipkin/ThriftSpanConverterTest.java
```diff
@@ -56,6 +64,105 @@ public void setUp() {
             .build();
   }
 
+  // sentinel value is used to mark tags that should *not* be present
+  private static final String SENTINEL = "sentinel";
+
+  // NA value is used to mark tags that should be present but we don't care what it's value is
+  private static final String NA = "na";
```
yurishkuro: s/NA/ANY

jaeger-zipkin/src/test/java/com/uber/jaeger/senders/zipkin/ThriftSpanConverterTest.java
```diff
@@ -56,6 +64,105 @@ public void setUp() {
             .build();
   }
 
+  // sentinel value is used to mark tags that should *not* be present
+  private static final String SENTINEL = "sentinel";
```
yurishkuro: maybe rename this to UNDEF, easier to understand/read

jaeger-zipkin/src/test/java/com/uber/jaeger/senders/zipkin/ThriftSpanConverterTest.java
```diff
@@ -56,6 +64,105 @@ public void setUp() {
...
+                  .withTag(Tags.SPAN_KIND.getKey(), Tags.SPAN_KIND_SERVER)
+                  .startManual();
+    }
+    com.twitter.zipkin.thriftjava.Span zipkinSpan = ThriftSpanConverter.convertSpan(span);
+
+    List<BinaryAnnotation> annotations = zipkinSpan.getBinary_annotations();;
+    for (String key : expectedTags.keySet()) {
+      Object expectedValue = expectedTags.get(key);
+      BinaryAnnotation anno = findBinaryAnnotation(annotations, key);
+      if (expectedValue == SENTINEL) {
```
yurishkuro: compare strings with equal()

jaeger-zipkin/src/test/java/com/uber/jaeger/senders/zipkin/ThriftSpanConverterTest.java
```diff
@@ -70,19 +177,18 @@ public void testSpanKindServerCreatesAnnotations() {
...
     }
-
     assertTrue(serverReceiveFound);
     assertTrue(serverSendFound);
   }
 
   @Test
   public void testSpanKindClientCreatesAnnotations() {
-    Span span = (com.uber.jaeger.Span) tracer.buildSpan("operation-name").startManual();
+    Span parent = (com.uber.jaeger.Span) tracer.buildSpan("operation-name").startManual();
```
yurishkuro: why does this test require a parent?

black-adder: it doesn't remnant of previous change, removing

jaeger-zipkin/src/main/java/com/uber/jaeger/senders/zipkin/ThriftSpanConverter.java
```diff
@@ -88,6 +90,20 @@
     Map<String, Object> tags = span.getTags();
     boolean isRpc = isRpc(span);
     boolean isClient = isRpcClient(span);
+    boolean firstSpanInProcess = span.getReferences().isEmpty() || isRpcServer(span);
+
+    Map<String, ?> processTags = span.getTracer().tags();
```
yurishkuro: scope this inside the if{}

PR #199, opened 2017-06-19
Fix conversion of ip address to int

mabn:
Properly convert bytes to unsigned integers

Fixes https://github.com/uber/jaeger-client-java/issues/198

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/utils/Utils.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/utils/Utils.java
@@ -50,7 +50,7 @@ public class Utils {
 
     int intIp = 0;
     for (byte octet : octets.getAddress()) {
-      intIp = (intIp << 8) | (octet);
+      intIp = (intIp << 8) | (octet & 0xFF);
     }
     return intIp;
   }
--- a/jaeger-core/src/test/java/com/uber/jaeger/UtilsTest.java
+++ b/jaeger-core/src/test/java/com/uber/jaeger/UtilsTest.java
@@ -49,9 +49,30 @@ public class UtilsTest {
   }
 
   @Test
-  public void testIpToInt32() {
+  public void testIpToInt32_localhost() {
     int expectedIp = (127 << 24) | 1;
     int actualIp = Utils.ipToInt("127.0.0.1");
+    assertEquals(expectedIp, actualIp);
+  }
+
+  @Test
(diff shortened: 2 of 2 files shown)


jaeger-core/src/main/java/com/uber/jaeger/utils/Utils.java
```diff
@@ -50,7 +50,7 @@ public static int ipToInt(String ip) throws EmptyIpException, NotFourOctetsExcep
 
     int intIp = 0;
     for (byte octet : octets.getAddress()) {
-      intIp = (intIp << 8) | (octet);
+      intIp = (intIp << 8) | (octet & 0xFF);
```
yurishkuro: for my education, since `octet` is already of type `byte`, why does `& 0xFF` make a difference?

mabn: `octet` is promoted to integer via binary numeric promotion (https://docs.oracle.com/javase/specs/jls/se7/html/jls-5.html#jls-5.6.2), but it's a signed integer so the mask is filled with 1s on the left for negative numbers. `& 0xFF` converts it to unsigned (It's a byte so no need to care about int overflow. If it mattered then it should be 0xFFL). At least that's how I understand that.

In Java 8 that would instead be: `Byte.toUnsignedInt(octet)`

jaeger-core/src/test/java/com/uber/jaeger/UtilsTest.java
```diff
@@ -49,9 +49,30 @@ public void testIpToInt32EmptyIpException() {
...
-  public void testIpToInt32() {
+  public void testIpToInt32_localhost() {
     int expectedIp = (127 << 24) | 1;
     int actualIp = Utils.ipToInt("127.0.0.1");
     assertEquals(expectedIp, actualIp);
   }
+
+  @Test
+  public void testIpToInt32_above127() {
+    int expectedIp = 176750850;
```
yurishkuro: the magic number is hard to validate, how about a sanity check
```java
assertEquals(expectedIp, (10 << 24) | (137 << 16) | (1 << 8) | 2); // sanity check
```

mabn: Makes sense. I've also changed assertions to `assertThat(x, equals(y))`, I think that they are more readable

yurishkuro:
Thanks!

PR #200, opened 2017-06-22
Bump gradle to 4.0

pavolloffay:
Reason: build in https://github.com/uber/jaeger-client-java/pull/194 failed a couple of times. Locally it was passing fine, I find out that `--parallel` was causing the hang. Upgrade gradle to 4.0 helped too.


pavolloffay:
It still sometimes hangs on:
```
> :jaeger-core:test > 80 tests completed
```

parallel tests takes ~40s
normal: ~1min 5s

pavolloffay:
I would rather disable `--parallel` in travis to avoid annoying restarts.

jaeger-crossdock/src/test/java/com/uber/jaeger/crossdock/resources/behavior/http/TraceBehaviorResourceTest.java
```diff
@@ -82,12 +82,10 @@ public static void setUpLogger() {
 
   @Before
   public void setUp() throws Exception {
-    port = System.getenv("TRACE_TEST_BEHAVIOR_PORT");
```
yurishkuro: what was setting this env variable before & why do we not need it?  cc @black-adder

pavolloffay: I did not find any usage and this test is not bundled in jar so it's not used in the container.

black-adder: looks like this variable is only used for testing but even then it's never set anywhere. Looks ok for removal. The crossdock tests are still passing.

PR #201, opened 2017-06-23
Automated the release process

jpkrohling:
This PR automates the release process, by setting the version to -SNAPSHOT
when on a regular build, but to a proper fixed version when building for a tag.

This effectively means that releasing is just a matter of tagging the code at
a given point.

It's still needed to manually bump the version after a tag.

At this stage, closing/releasing of the Nexus repository is still manual and
will be automated if the main release procedure proves stable enough.


pavolloffay:
Could you please describe the process in `RELEASE.md`?

jpkrohling:
`RELEASE.md` updated

build.gradle
```diff
@@ -31,7 +31,7 @@ allprojects {
     apply plugin: 'jacoco'
 
     group = 'com.uber.jaeger'
-    version = '0.20.0-SNAPSHOT'
+    version = System.env.TRAVIS_TAG ? System.env.TRAVIS_TAG : '0.20.0-SNAPSHOT'
```
pavolloffay: We could also check the tag on regex.

jpkrohling: Why?

pavolloffay: somebody could accidentally push `i-had-a-hard-time-yesterday-1.0` :/

jpkrohling: At the current state, this is not a problem. Either this will fail during the upload phase, or we could just not close/release the staging repo at Nexus.

.travis.yml
```diff
@@ -1,53 +1,54 @@
...
+  - "$HOME/.gradle/wrapper/"
 before_install:
+- if [ "$TRAVIS_SECURE_ENV_VARS" == true ]; then openssl aes-256-cbc -K $encrypted_677f232983c0_key -iv $encrypted_677f232983c0_iv -in travis/signing-key.asc.enc -out travis/signing-key.asc -d ; fi
+- if [ "$TRAVIS_SECURE_ENV_VARS" == true ]; then gpg --no-tty --batch --allow-secret-key-import --import travis/signing-key.asc ; fi
+- rm -rf "$HOME/.gradle/gradle.properties"
+- echo signing.keyId="${SIGNING_KEY_ID}" > "$HOME/.gradle/gradle.properties"
+- echo signing.password="${SIGNING_KEY_PASSPHRASE}" >> "$HOME/.gradle/gradle.properties"
+- echo ossrhUsername="${SONATYPE_USERNAME}" >> "$HOME/.gradle/gradle.properties"
+- echo ossrhPassword="${SONATYPE_PASSWORD}" >> "$HOME/.gradle/gradle.properties"
+- echo signing.secretKeyRingFile="${HOME}/.gnupg/secring.gpg" >> "$HOME/.gradle/gradle.properties"
```
yurishkuro: I think these all should be skipped when not `"$TRAVIS_SECURE_ENV_VARS`

RELEASE.md
```diff
@@ -1,45 +1,16 @@
...
 The release process consists of these steps:
-  1. Edit build.gradle and remove the `-SNAPSHOT` suffix from the `version` parameter
-  1. Add an entry to CHANGELOG.pm with description of changes in the new release
-  1. Commit as "Preparing release"
-  1. Run `make release`
-     * Make sure you have the right version 0.9.2 of Thrift installed: `thrift -version`
-  1. Once the artifacts are uploaded to Sonatype staging, release them to Maven Central
+  1. Add an entry to [`CHANGELOG.md`](CHANGELOG.rst) with description of changes in the new release
+  1. Commit your `CHANGELOG` changes
+  1. Tag as a version (`git tag 0.20.0`). Use the one from [`build.gradle`](build.gradle) as reference.
```
yurishkuro: in [zipkin release scripts](https://github.com/openzipkin/zipkin/blob/master/RELEASE.md) the manual tag has a format `release-M.N.P`, and the script creates another tag (I think). If here the manual tag is the only one, it should have the `v` prefix, a common convention for semver.

And as @pavolloffay said, I think the build should match on the vM.N.P pattern when publishing a release, instead of any tag.

jpkrohling:
I implemented two changes based on the comments:

1. The logic to determine the version is now on a `def` of its own, matching the regex `[v]?((\d+)\.(\d+)\.(\d+))`. Note that the very first part makes the `v` optional. Should it be required? When the value for the tag is not a match for this regex, the -SNAPSHOT version is used.
1. Instead of having several `before_install` commands, I created a new script (`prepare-signing.sh`) that would make it easier to maintain and debug.

I need the following from @yurishkuro :

- [ ] a real `signing-key.asc.enc`, with the encrypted private key that will be used to sign the artifacts. Ideally, this would be a "project" key, based on an email that is clear that comes from this project. Something like, `build@jaegertracing.io`.
- [ ] encrypted values for SIGNING_KEY_ID and SIGNING_KEY_PASSPHRASE. In fact, the KEY_ID does *not* need to be secure, but I'll need to know what it is, to add to the Travis configuration file

```
travis encrypt-file travis/signing-key.asc -r uber/jaeger-client-java
travis encrypt SIGNING_KEY_ID=5557EADD
travis encrypt SIGNING_KEY_PASSPHRASE=...
```

yurishkuro:
added in https://github.com/uber/jaeger-client-java/pull/205

RELEASE.md
```diff
@@ -1,45 +1,16 @@
...
+  soon reach Maven Central.
      * https://oss.sonatype.org/, log in, go to Staging Repositories
      * In the top-right corner search box type `uber`, and find your uploaded repository
      * Hit Close button. Monitor status on Activity tab at the bottom, hitting Refresh.
      * Once Close is successful, the Release button will become available, so hit it
      * Keep hitting Refresh while sync to Maven is in progress. Once it's complete, the repository will disappear.
-  1. Bump the version in `build.gradle` to the next release and add `-SNAPSHOT suffix
-  1. Commit as "Back to development"
-
-The `uploadArchives` step performs signing of the artifacts. For that to work the script
```
yurishkuro: I would prefer to keep this section, in case we need to re-generate the key.

jpkrohling:
Once this is OK, I'll squash the commits and update the PR. I'll also try to release `v0.20.1` (and possibly other `v0.20.x`) to test the procedure.

PR #203, opened 2017-06-26
Fix crossdock

black-adder:
If the environment variable is empty, default to UDP. This is breaking the crossdock tests for all jaeger repos.


travis/build-crossdock.sh
```diff
@@ -10,7 +10,7 @@ export TAG=`if [ "$BRANCH" == "master" ]; then echo "latest"; else echo "${BRANC
 echo "TRAVIS_BRANCH=$TRAVIS_BRANCH, REPO=$REPO, PR=$PR, BRANCH=$BRANCH, TAG=$TAG"
 
 # Only push the docker container to Docker Hub for master branch
-if [ "$BRANCH" == "master" ]; then echo 'upload to Docker Hub'; else echo 'skip docker upload for PR'; exit 0; fi
+#if [ "$BRANCH" == "master" ]; then echo 'upload to Docker Hub'; else echo 'skip docker upload for PR'; exit 0; fi
```
black-adder: will revert back after testing this works with other repos

PR #206, opened 2017-06-29
Fixed signing keys

jpkrohling:


jpkrohling:
I changed the build to use this key:
https://pgp.mit.edu/pks/lookup?op=get&search=0x8ECC15DC9A2E1C5E

Not sure if there's any other way of testing it than to merge and trigger a new release :/

PR #210, opened 2017-07-07
Add baggage restriction manager

black-adder:


jaeger-core/src/main/java/com/uber/jaeger/baggage/RemoteBaggageRestrictionManager.java
```diff
@@ -0,0 +1,136 @@
...
+      BaggageRestrictionProxy proxy,
+      Metrics metrics,
+      boolean denyBaggageOnInitializationFailure,
+      int refreshIntervalMs,
+      int refreshDelayMs
+  ) {
+    this.serviceName = serviceName;
+    this.proxy = proxy;
+    this.metrics = metrics;
+    this.denyBaggageOnInitializationFailure = denyBaggageOnInitializationFailure;
```
vprithvi: This name is misleading because of the initialization race condition.

Specifically, there is a period of time after the creation of the object but before the first successful response from jaeger agent where the failure behavior applies. 

Perhaps `denyBaggageDuringInitializationOrFailure` is more descriptive. 

Also, I don't really like how we handle initialization. I'd like for object initialization not be blocking (as it is), but also have a means of signalling that the object is ready for accepting requests.  A lightweight interface with `startUp` and `shutdown` methods might make sense.

jaeger-core/src/main/java/com/uber/jaeger/baggage/RemoteBaggageRestrictionManager.java
```diff
@@ -0,0 +1,136 @@
...
+    metrics.baggageRestrictionsUpdateSuccess.inc(1);
+  }
+
+  private void updateBaggageRestrictions(List<com.uber.jaeger.baggage.http.BaggageRestriction> restrictions) {
+    Map<String, Integer> baggageRestrictions = new HashMap<String, Integer>();
+    for (com.uber.jaeger.baggage.http.BaggageRestriction restriction : restrictions) {
+      baggageRestrictions.put(restriction.getBaggageKey(), restriction.getMaxValueLength());
+    }
+    ReentrantReadWriteLock.WriteLock writeLock = this.lock.writeLock();
+    writeLock.lock();
```
vprithvi: I'm not convinced that we need synchronization because we are using `baggageRestrictions` like an Immutable Object. 
Simply declaring `this.restrictions` and `this.initialized` to be volatile should be sufficient. 
See https://docs.oracle.com/javase/tutorial/essential/concurrency/atomic.html

jaeger-core/src/main/java/com/uber/jaeger/baggage/RemoteBaggageRestrictionManager.java
```diff
@@ -0,0 +1,136 @@
...
+    this.restrictions = baggageRestrictions;
+    this.initialized = true;
+    writeLock.unlock();
+  }
+
+  public void close() {
+    synchronized (this) {
```
vprithvi: Don't need this. 
`pollTimer` handles concurrent calls fine on its own.

jaeger-core/src/main/java/com/uber/jaeger/baggage/RemoteBaggageRestrictionManager.java
```diff
@@ -0,0 +1,136 @@
...
+
+  @Override
+  public BaggageValidity isValidBaggageKey(String key) {
+    BaggageValidity restriction;
+
+    if (!this.initialized) {
+      if (this.denyBaggageOnInitializationFailure) {
+        restriction = BaggageValidity.of(false, 0);
+      } else {
+        restriction = BaggageValidity.of(true, DEFAULT_MAX_VALUE_LENGTH);
```
vprithvi: Perhaps `BaggageValidity.of(true, DEFAULT_MAX_VALUE_LENGTH)` can go into `BaggageRestrictionManager`

jaeger-core/src/main/java/com/uber/jaeger/baggage/RemoteBaggageRestrictionManager.java
```diff
@@ -0,0 +1,136 @@
...
+      }
+    } else {
+      ReentrantReadWriteLock.ReadLock readLock = this.lock.readLock();
+      readLock.lock();
```
vprithvi: Don't need this.

jaeger-core/src/main/java/com/uber/jaeger/baggage/RemoteBaggageRestrictionManager.java
```diff
@@ -0,0 +1,136 @@
...
+      if (this.restrictions.containsKey(key)) {
```
vprithvi: It's more efficient to simply do `restrictions.get(key)` and check for `null`.

jaeger-core/src/test/java/com/uber/jaeger/samplers/HttpSamplingManagerTest.java
```diff
@@ -27,6 +27,7 @@
 import static org.junit.Assert.assertNull;
 
 import com.uber.jaeger.exceptions.SamplingStrategyErrorException;
+import com.uber.jaeger.mocks.MockAgentResource;
```
vprithvi: Unrelated/unnecessary change.

jaeger-core/src/main/java/com/uber/jaeger/Span.java
```diff
@@ -117,8 +118,18 @@ public String getOperationName() {
   @Override
   public Span setBaggageItem(String key, String value) {
```
vprithvi: `setBaggageItem` should check that `key` is not null or empty and `value` isn't null.

jaeger-core/src/main/java/com/uber/jaeger/Span.java
```diff
@@ -117,8 +118,18 @@ public String getOperationName() {
...
     synchronized (this) {
-      // TODO emit a metric whenever baggage is updated
+      BaggageValidity validity = this.tracer.getBaggageRestrictionManager().isValidBaggageKey(key);
+      if (!validity.isValid()) {
+        this.tracer.getMetrics().baggageUpdateFailure.inc(1);
+        return this;
+      }
       String prevItem = this.getBaggageItem(key);
+      boolean truncated = false;
+      if (value.length() > validity.getMaxValueLength()) {
```
vprithvi: This could result in a NPE.

jaeger-core/src/main/java/com/uber/jaeger/Span.java
```diff
@@ -117,8 +118,18 @@ public String getOperationName() {
   @Override
   public Span setBaggageItem(String key, String value) {
     synchronized (this) {
-      // TODO emit a metric whenever baggage is updated
+      BaggageValidity validity = this.tracer.getBaggageRestrictionManager().isValidBaggageKey(key);
```
vprithvi: This could result in a NPE.

jaeger-core/src/main/java/com/uber/jaeger/baggage/BaggageValidity.java
```diff
@@ -0,0 +1,31 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.uber.jaeger.baggage;
+
+import lombok.Value;
+
+@Value(staticConstructor = "of")
+public class BaggageValidity {
```
vprithvi: Perhaps `BaggageValidityResponse` is a better name?

jaeger-core/src/main/java/com/uber/jaeger/baggage/BaggageRestrictionManager.java
```diff
@@ -0,0 +1,27 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.uber.jaeger.baggage;
+
+public interface BaggageRestrictionManager {
+  BaggageValidity isValidBaggageKey(String key);
```
vprithvi: I feel that this interface could be better by encapsulating the action to be taken when restrictions are in place. 
More concretely, I suggest replacing `isValidBaggageKey(key)` with `sanitizeBaggage(key, value)`
This abstracts an implementation detail (length of baggage) leaking out to `setBaggage`. Further, it allows for future customizability. 
This method could simply return a `Map<String,String>` and the sanitized key/value. The returned map could contain any logs that are to be added. 
On success, the map would be empty. When truncating a value, it could contain `truncated:true`, and if in the future, we decide to add other restrictions, it could easily accommodate those.

jaeger-core/src/main/java/com/uber/jaeger/utils/Utils.java
```diff
@@ -66,4 +73,27 @@ public static long uniqueId() {
   public static boolean equals(Object a, Object b) {
     return (a == b) || (a != null && a.equals(b));
   }
+
+  public static String makeGetRequest(String urlToRead) throws IOException {
+    URL url = new URL(urlToRead);
+    HttpURLConnection conn = (HttpURLConnection) url.openConnection();
```
mabn: I'd consider setting timeouts to some reasonable values.

jaeger-core/src/test/java/com/uber/jaeger/samplers/HttpSamplingManagerTest.java
black-adder: I moved the mock file to a more convenient location so both this and the httpbaggageproxy can both use it.

jaeger-core/src/main/java/com/uber/jaeger/baggage/BaggageRestrictionManager.java
```diff
@@ -0,0 +1,70 @@
...
+import lombok.AccessLevel;
+import lombok.Getter;
+
+public abstract class BaggageRestrictionManager {
+  static final int DEFAULT_MAX_VALUE_LENGTH = 2048;
+  static final SanitizedBaggage INVALID_BAGGAGE = SanitizedBaggage.of(false, null, null, null, false);
+
+  public abstract SanitizedBaggage sanitizeBaggage(String key, String value, String prevValue);
+
+  SanitizedBaggage sanitizeBaggage(String key, String value, String prevValue, int maxValueLength) {
```
yurishkuro: What is the semantic difference between truncated and sanitized? What other modification could be done aside from truncating?

jaeger-core/src/main/java/com/uber/jaeger/baggage/BaggageRestrictionManager.java
```diff
@@ -0,0 +1,70 @@
...
+    Map<String, String> fields = new HashMap<String, String>();
+    fields.put("event", "baggage");
+    fields.put("key", key);
+    fields.put("value", baggage.getTruncatedValue());
+    if (baggage.isTruncated()) {
+      fields.put("truncated", "true");
+    }
+    if (prevValue != null) {
+      fields.put("override", "true");
+    }
```
yurishkuro: this is an abstraction leak. Why should baggage manager care about log fields? Also, because there is no span context here, we're creating the map every time even though the span may be unsampled and logs will be no-op.

jaeger-core/src/main/java/com/uber/jaeger/baggage/BaggageRestrictionProxy.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package com.uber.jaeger.baggage;
+
+import com.uber.jaeger.baggage.http.BaggageRestriction;
+import com.uber.jaeger.exceptions.BaggageRestrictionException;
+
+import java.util.List;
+
+public interface BaggageRestrictionProxy {
```
yurishkuro: baggage restriction _manager_ proxy

jaeger-core/src/main/java/com/uber/jaeger/baggage/HttpBaggageRestrictionProxy.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+  }
+
+  @Override
+  public List<BaggageRestriction> getBaggageRestrictions(String serviceName)
+      throws BaggageRestrictionException {
+    String jsonString;
+    try {
+      jsonString =
+          makeGetRequest(builder.setParameter("service", serviceName).build());
```
yurishkuro: is the url builder really reusable like that? Especially from multiple threads?

jaeger-core/src/main/java/com/uber/jaeger/baggage/BaggageRestrictionManager.java
```diff
@@ -0,0 +1,70 @@
...
+
+  TruncatedBaggage truncateBaggage(String value, int maxValueLength) {
+    if (value.length() > maxValueLength) {
+      return new TruncatedBaggage(value.substring(0, maxValueLength), true);
+    }
+    return new TruncatedBaggage(value, false);
+  }
+
+  @Getter(AccessLevel.PACKAGE)
+  class TruncatedBaggage {
```
yurishkuro: we should really consider the performance impact of all these helper classes. Conceptually we're doing this:

```
func setBaggageItem(key, value) {
    if !mgr.allows(key) return
    value2 = mgr.truncate(value)
    span.baggage.put(key, value2)
    if span.sampled {
       logFields = ...
       if value.length() != value2.length() {
         logFields['truncated'] = true
       }
       span.log(logFields)
    }
}
```

Aside from logFields there are no memory allocations here.

jaeger-core/src/main/java/com/uber/jaeger/baggage/BaggageRestrictionManager.java
black-adder: we could be normalizing the key ala: https://github.com/uber/jaeger-client-go/blob/master/span.go#L259 or we might have to obfuscate the value, etc.

yurishkuro: Doesn't make sense to normalize the key since the user will be expecting to get back the same key that they put in. We may encode the key when serialised, but not here.

oibe:
@black-adder Is this an out of date diff? If so close it.

PR #217, opened 2017-07-24
Add BaggageRestrictionManager

black-adder:
Subset of this: https://github.com/uber/jaeger-client-java/pull/210/files. Only adding the BaggageRestrictionManager abstract class and the DefaultBaggageRestrictionManager.

- [x] Tests


jaeger-core/src/main/java/com/uber/jaeger/Span.java
```diff
@@ -117,21 +118,35 @@ public String getOperationName() {
   @Override
   public Span setBaggageItem(String key, String value) {
     synchronized (this) {
-      // TODO emit a metric whenever baggage is updated
+      if (key == null || value == null) {
+        return this;
+      }
+      SanitizedBaggage baggage = this.tracer.getBaggageRestrictionManager().sanitizeBaggage(key, value);
+      if (!baggage.isValid()) {
+        this.tracer.getMetrics().baggageUpdateFailure.inc(1);
```
yurishkuro: might be useful to log this to span as well

black-adder:
I attempted the strategy pattern approach and it's a hacky mess. If you're concerned about minimizing memory allocations, maybe I can go back to my first ever approach where I only return (isValid, maxValueLength) which are created once (a minute) and cached.

jaeger-core/src/main/java/com/uber/jaeger/baggage/BaggageValidity.java
```diff
@@ -0,0 +1,77 @@
...
+    boolean truncated = false;
+    if (value.length() > maxValueLength) {
+      truncated = true;
+      value = value.substring(0, maxValueLength);
+      metrics.baggageTruncate.inc(1);
+    }
+
+    String prevItem = span.getBaggageItem(key);
+    this.logFields(span, key, value, prevItem, truncated, false);
+    SpanContext context = span.context().withBaggageItem(key, value);
```
black-adder: I have to create the new context and pass it back given the setter isn't public. Also the BaggageValidity class reeks of feature envy. I could move this file down to the same namespace as Span and make the context setter protected or I could remove the functionality in this class completely and inline it in Span.setBaggageItem()

jaeger-core/src/main/java/com/uber/jaeger/Span.java
```diff
@@ -117,21 +118,13 @@ public String getOperationName() {
...
-        fields.put("key", key);
-        fields.put("value", value);
-        if (prevItem != null) {
-          fields.put("override", "true");
-        }
-        return this.log(fields);
+      if (key == null || value == null) {
+        return this;
       }
+      BaggageValidity baggageValidity = this.getTracer().getBaggageRestrictionManager().isBaggageValid(key, value);
```
yurishkuro: how about 
  * s/BaggageValidity/BaggageRestriction/
  * s/isBaggageValue(key, value)/getBaggageRestriction(key)/
  * s/sanitizeBaggage/setBaggage/
  * move the call to getBaggageRestriction outside of `synchronized` since it's not dependent on the span

yurishkuro: or `s/BaggageValidity/BaggageSetter/` (back to "strategy" pattern)

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -450,13 +450,14 @@ public ActiveSpan startActive() {
...
-    private Metrics metrics;
+    private Metrics metrics = new Metrics(new StatsFactoryImpl(new NullStatsReporter()));
     private String serviceName;
     private Clock clock = new SystemClock();
     private Map<String, Object> tags = new HashMap<String, Object>();
     private boolean zipkinSharedRpcSpan;
     private ActiveSpanSource activeSpanSource = new ThreadLocalActiveSpanSource();
-    private BaggageRestrictionManager baggageRestrictionManager = new DefaultBaggageRestrictionManager();
+    private BaggageRestrictionManager baggageRestrictionManager =
+        new DefaultBaggageRestrictionManager(this.metrics);
```
vprithvi: This introduces a bug; the instance of metrics used by `DefaultBaggageRestrictionManager` would always be the one set in L453.

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -499,6 +508,9 @@ public Builder withZipkinSharedRpcSpan() {
 
     public Builder withMetrics(Metrics metrics) {
       this.metrics = metrics;
+      if (this.baggageRestrictionManager instanceof DefaultBaggageRestrictionManager) {
+        this.baggageRestrictionManager = new DefaultBaggageRestrictionManager(this.metrics);
+      }
```
vprithvi: This is bad because `withMetrics` now has multiple responsibilities. 
A better way to handle this is to leave `baggageRestrictionManager` uninitialized, and initialize it in `build` with the right metrics instance.

black-adder: dope

jaeger-core/src/main/java/com/uber/jaeger/baggage/BaggageSetter.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    String prevItem = span.getBaggageItem(key);
+    this.logFields(span, key, value, prevItem, truncated, false);
+    SpanContext context = span.context().withBaggageItem(key, value);
+    metrics.baggageUpdateSuccess.inc(1);
+    return context;
+  }
+
+  void logFields(Span span, String key, String value, String prevItem, boolean truncated, boolean invalid) {
```
vprithvi: This should be `private`.

jaeger-core/src/main/java/com/uber/jaeger/baggage/BaggageSetter.java
```diff
@@ -0,0 +1,78 @@
...
+import lombok.Value;
+
+@Value(staticConstructor = "of")
+public class BaggageSetter {
+  final boolean valid;
+  final int maxValueLength;
+  final Metrics metrics;
+
+  public SpanContext setBaggage(Span span, String key, String value) {
+    if (!this.isValid()) {
```
vprithvi: It's simpler to check the boolean directly instead of using the getter

jaeger-core/src/main/java/com/uber/jaeger/baggage/BaggageSetter.java
```diff
@@ -0,0 +1,78 @@
...
+      metrics.baggageUpdateFailure.inc(1);
+      this.logFields(span, key, value, null, false, true);
```
vprithvi: There is no need to use `this` here

jaeger-core/src/main/java/com/uber/jaeger/baggage/BaggageRestrictionManager.java
```diff
@@ -0,0 +1,29 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.uber.jaeger.baggage;
+
+public abstract class BaggageRestrictionManager {
```
vprithvi: Could you add `javadocs` on what this is?

jaeger-core/src/main/java/com/uber/jaeger/baggage/BaggageSetter.java
```diff
@@ -0,0 +1,78 @@
...
+
+import lombok.Value;
+
+@Value(staticConstructor = "of")
+public class BaggageSetter {
+  final boolean valid;
+  final int maxValueLength;
+  final Metrics metrics;
+
+  public SpanContext setBaggage(Span span, String key, String value) {
```
vprithvi: Add javadoc, simply based on the method name, I'd conclude that it simply sets baggage.

jaeger-core/src/main/java/com/uber/jaeger/baggage/BaggageRestrictionManager.java
```diff
@@ -0,0 +1,29 @@
...
+  static final int DEFAULT_MAX_VALUE_LENGTH = 2048;
```
vprithvi: This is only used in the `DefaultBaggageRestrictionManager`. Do you see this being made use of elsewhere? Perhaps it can be moved to `DefaultBaggageRestrictionManager`

jaeger-core/src/main/java/com/uber/jaeger/baggage/DefaultBaggageRestrictionManager.java
```diff
@@ -0,0 +1,42 @@
...
+  public DefaultBaggageRestrictionManager(Metrics metrics) {
+    this(metrics, DEFAULT_MAX_VALUE_LENGTH);
+  }
+
+  public DefaultBaggageRestrictionManager(Metrics metrics, int maxValueLength) {
+    baggageSetter = BaggageSetter.of(true, maxValueLength, metrics);
+  }
+
+  @Override
+  public BaggageSetter getBaggageSetter(String key) {
```
vprithvi: Do we need this? 
Can't we use lombok's `@Getter` annotation?

jaeger-core/src/test/java/com/uber/jaeger/baggage/BaggageSetterTest.java
```diff
@@ -0,0 +1,116 @@
...
+    assertBaggageLogs(span, KEY, expected, true, false);
+    assertEquals(expected, ctx.getBaggageItem(KEY));
+
+    assertEquals(
+        1L, metricsReporter.counters.get("jaeger.baggage-truncate").longValue());
+    assertEquals(
+        1L, metricsReporter.counters.get("jaeger.baggage-update.result=ok").longValue());
+  }
+
+  private void assertBaggageLogs(
```
vprithvi: It looks like testing concerns aren't separated between SpanTest and BaggageSetter.

Ideally, we would only do the following in SpanTest
Assert that `span.setBaggageItem` invokes `BaggageSetter` with the right arguments. 
Assert that `span.getBaggeItem` can retrieve baggage that has been set

All of the detailed test cases that only involve setting baggage should be under the purview of `BaggageSetterTest`. This results in a clear separation of concerns and code deduplication.

jaeger-core/src/main/java/com/uber/jaeger/baggage/BaggageRestrictionManager.java
black-adder: it'll be used in another manager in the follow up PR

jaeger-core/src/main/java/com/uber/jaeger/baggage/DefaultBaggageRestrictionManager.java
black-adder: getBaggageSetter is an interface function

vprithvi: discussed this offline: my suggestion doesn't work because the getter requires a `key`

jaeger-core/src/main/java/com/uber/jaeger/baggage/BaggageSetter.java
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * BaggageSetter is a class that sets baggage and the logs associated
+ * with the baggage on a given {@link Span}.
+ */
+@Value(staticConstructor = "of")
+public class BaggageSetter {
+  final boolean valid;
+  final int maxValueLength;
+  final Metrics metrics;
```
yurishkuro: * to make fields visible to subclasses, they should be `protected`. Did you also need package visibility?
 * given that these fields are final, there should be a constructor setting them. If they are only set by subclasses, then the constructor can be protected (but then this class should be abstract).
 * the concept of `valid` is not clear, please document

jaeger-core/src/main/java/com/uber/jaeger/baggage/BaggageSetter.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    String prevItem = span.getBaggageItem(key);
+    logFields(span, key, value, prevItem, truncated, false);
+    SpanContext context = span.context().withBaggageItem(key, value);
+    metrics.baggageUpdateSuccess.inc(1);
+    return context;
+  }
+
+  private void logFields(Span span, String key, String value, String prevItem, boolean truncated, boolean invalid) {
```
yurishkuro: avoid negatives - s/invalid/valid/

jaeger-core/src/main/java/com/uber/jaeger/baggage/BaggageSetter.java
```diff
@@ -0,0 +1,96 @@
...
+   *
+   * @param  span  the span to set the baggage on
+   * @param  key   the baggage key to set
+   * @param  value the baggage value to set
+   * @return       the SpanContext with the baggage set
+   */
+  public SpanContext setBaggage(Span span, String key, String value) {
+    if (!valid) {
+      metrics.baggageUpdateFailure.inc(1);
+      logFields(span, key, value, null, false, true);
```
yurishkuro: if you move prevItem and truncated declarations to the top, this could be made a lot more readable: `logFields(span, key, value, prevValue, truncated, valid);`

jaeger-core/src/main/java/com/uber/jaeger/baggage/BaggageSetter.java
```diff
@@ -0,0 +1,96 @@
...
+      fields.put("key", key);
+      fields.put("value", value);
+      if (prevItem != null) {
+        fields.put("override", "true");
+      }
+      if (truncated) {
+        fields.put("truncated", "true");
+      }
+      if (invalid) {
+        fields.put("invalid", "true");
```
yurishkuro: why are you logging these as strings instead of booleans?

yurishkuro: ah, nvm

jaeger-core/src/main/java/com/uber/jaeger/baggage/BaggageSetter.java
black-adder: @Value(staticConstructor = "of") generates a private constructor and a public factory for this class.

jaeger-core/src/main/java/com/uber/jaeger/baggage/BaggageSetter.java
```diff
@@ -0,0 +1,101 @@
...
+ */
+@Value(staticConstructor = "of")
+public class BaggageSetter {
+
+  private final String key;
+  /**
+   * This flag represents whether the key is a valid baggage key. If valid
+   * the baggage key:value will be written to the span.
+   */
+  private final boolean valid;
```
oibe: If something is not a valid baggage key then why create a BaggageSetter for it?

yurishkuro: I assume it would create a Noop setter. This way the validity does not need to be exposed outside of baggage manager into the Span

jaeger-core/src/main/java/com/uber/jaeger/baggage/DefaultBaggageRestrictionManager.java
```diff
@@ -0,0 +1,54 @@
...
+
+  public DefaultBaggageRestrictionManager(Metrics metrics, int maxValueLength) {
+    this.maxValueLength = maxValueLength;
+    this.metrics = metrics;
+    this.baggageSetters = new HashMap<String, BaggageSetter>();
+  }
+
+  @Override
+  public BaggageSetter getBaggageSetter(String key) {
+    BaggageSetter setter = baggageSetters.get(key);
```
yurishkuro: unsynchronized map

black-adder:
Rewrote to fit into new API.

jaeger-core/src/main/java/com/uber/jaeger/baggage/BaggageSetter.java
```diff
@@ -0,0 +1,99 @@
...
+
+  public BaggageSetter(BaggageRestrictionManager restrictionManager, Metrics metrics) {
+    this.restrictionManager = restrictionManager;
+    this.metrics = metrics;
+  }
+
+  /**
+   * Sets the baggage key:value on the {@link Span} and the corresponding
+   * logs. Whether the baggage is set on the span depends on if the key
+   * is valid.
```
yurishkuro: s/is valid/is allowed to be set by this service/

jaeger-core/src/main/java/com/uber/jaeger/Span.java
```diff
@@ -116,22 +117,13 @@ public String getOperationName() {
...
-        Map<String, String> fields = new HashMap<String, String>();
-        fields.put("event", "baggage");
-        fields.put("key", key);
-        fields.put("value", value);
-        if (prevItem != null) {
-          fields.put("override", "true");
-        }
-        return this.log(fields);
-      }
+      this.context = getTracer().setBaggage(this, key, value);
```
yurishkuro: why getTracer instead of just `tracer`?

jaeger-core/src/main/java/com/uber/jaeger/baggage/BaggageRestrictionManager.java
```diff
@@ -0,0 +1,35 @@
...
+ * THE SOFTWARE.
+ */
+
+package com.uber.jaeger.baggage;
+
+/**
+ * BaggageRestrictionManager is an interface for a class that manages baggage
+ * restrictions for baggage keys. The manager will return a {@link Restriction}
+ * for a specific baggage key which will determine whether the baggage key is
+ * allowed and any other applicable restrictions on the baggage value.
```
yurishkuro: s/is allowed/is allowed for the current service/

jaeger-core/src/main/java/com/uber/jaeger/baggage/BaggageRestrictionManager.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+public interface BaggageRestrictionManager {
+  int DEFAULT_MAX_VALUE_LENGTH = 2048;
+
+  Restriction getRestriction(String key);
```
yurishkuro: shouldn't we pass the service name?

black-adder:
I'm not sure how the whole client will support the per service use case. Maybe we should have a plan for that before making the baggagerestrictionmanager battle ready?

yurishkuro:
the service name will be associated with the span, not the tracer. But I see how in this version you filter on the service name at the time of querying the backend, so that's ok for now. But it will be an API change in the future, so it might make sense to just pass the service name from the tracer to baggage restriction manager, even if it ignores it for now.

black-adder:
> just pass the service name from the tracer to baggage restriction manager (BRM)

I'm not sure if that's future proof. My assumption is that in this dystopian future with no rules, people will bring up a tracer without a service name which means we can't initialize the BRM with a service. This means that we have to change `getRestriction(key)` to `getRestriction(service, key)` and then have BRM asynchronously fill its cache with restrictions for the service.

1. Looks like we'd need a batch endpoint to agent to retrieve restrictions for multiple services
2. First call to getRestriction will always fail (unless we have a failClosed flag)

PR #218, opened 2017-07-24
Use Pooling for HttpSender

### LINKED ISSUE #216: httpSender throws IllegalStateException due to the usage of defaultHttpClientsd
```127.0.0.1 - - [21/Jul/2017:16:22:37 +0000] "GET /single-span HTTP/1.1" 200 72 "-" "curl/7.51.0" 159
Exception in thread "jaeger.RemoteReporter-QueueProcessor" java.lang.IllegalStateException: Invalid use of BasicClientConnManager: connection still allocated.
Make sure to release the connection before allocating another one.
	at org.apache.http.impl.conn.BasicClientConnectionManager.getConnection(BasicClientConnectionManager.java:162)
	at org.apache.http.impl.conn.BasicClientConnectionManager$1.getConnection(BasicClientConnectionManager.java:139)
	at org.apache.http.impl.client.DefaultRequestDirector.execute(DefaultRequestDirector.java:456)
	at org.apache.http.impl.client.AbstractHttpClient.execute(AbstractHttpClient.java:906)
	at org.apache.http.impl.client.AbstractHttpClient.execute(AbstractHttpClient.java:805)
	at org.apache.http.impl.client.AbstractHttpClient.execute(AbstractHttpClient.java:784)
	at com.uber.jaeger.senders.HttpSender.send(HttpSender.java:67)
	at com.uber.jaeger.senders.ThriftSender.flush(ThriftSender.java:111)
	at com.uber.jaeger.reporters.RemoteReporter$FlushCommand.execute(RemoteReporter.java:164)
	at com.uber.jaeger.reporters.RemoteReporter$QueueProcessor.run(RemoteReporter.java:183)
	at java.lang.Thread.run(Thread.java:745)
```

What happened was that, the first span is sent to remote collector through the httpSender. 
The rest are not. I guess  the root cause is due to the use of defaultHttpClient because the basicClientConnectionManager is a simple connection manager that maintains only one connection at a time. Even though it is thread safe, it ought to be used by one execution thread only. 
Once we meet the illegalStateException, the rest spans cannot be sent to the remote collector through httpSender. 

Here is what i did to pass in the httpsender
```
            final Configuration.ReporterConfiguration reportConf =
                    createJaegerReportConfiguration(conf);
            final Configuration.SamplerConfiguration samplerConf =
                    new Configuration.SamplerConfiguration("const", 1);
            final B3TextMapCodec b3Codec = new B3TextMapCodec();
            GlobalTracer.register(
                    new com.uber.jaeger.Configuration(
                            componentName,
                            samplerConf,
                            reportConf).getTracerBuilder().
                            registerExtractor(Format.Builtin.HTTP_HEADERS, b3Codec).
                            registerInjector(Format.Builtin.HTTP_HEADERS, b3Codec).build());
```
could anyone help look into this please? thanks a lot

vprithvi:
- Fixes #216, we previously used `BasicClientConnManager`, which only uses a single connection.

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/senders/HttpSender.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/senders/HttpSender.java
@@ -24,6 +24,7 @@ package com.uber.jaeger.senders;
 
 import com.uber.jaeger.thriftjava.Batch;
 import com.uber.jaeger.thriftjava.Process;
+import com.uber.jaeger.thriftjava.Span;
 import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
@@ -34,6 +35,7 @@ import org.apache.http.client.HttpClient;
 import org.apache.http.client.methods.HttpPost;
 import org.apache.http.entity.ByteArrayEntity;
 import org.apache.http.impl.client.DefaultHttpClient;
+import org.apache.http.impl.conn.PoolingClientConnectionManager;
 import org.apache.thrift.TException;
 import org.apache.thrift.TSerializer;
 import org.apache.thrift.protocol.TBinaryProtocol.Factory;
@@ -44,7 +46,7 @@ public class HttpSender extends ThriftSender {
   private static final TProtocolFactory PROTOCOL_FACTORY = new Factory();
   private static final TSerializer SERIALIZER = new TSerializer(PROTOCOL_FACTORY);
 
-  private final HttpClient httpClient = new DefaultHttpClient();
+  private final HttpClient httpClient = new DefaultHttpClient(new PoolingClientConnectionManager());
   private final URI collectorUri;
 
   /**
@@ -57,7 +59,7 @@ public class HttpSender extends ThriftSender {
   }
 
   @Override
-  public void send(Process process, List<com.uber.jaeger.thriftjava.Span> spans) throws TException {
+  public void send(Process process, List<Span> spans) throws TException {
     Batch batch = new Batch(process, spans);
     byte[] bytes = SERIALIZER.serialize(batch);
 


jaeger-core/src/main/java/com/uber/jaeger/senders/HttpSender.java
```diff
@@ -44,7 +46,7 @@
   private static final TProtocolFactory PROTOCOL_FACTORY = new Factory();
   private static final TSerializer SERIALIZER = new TSerializer(PROTOCOL_FACTORY);
 
-  private final HttpClient httpClient = new DefaultHttpClient();
+  private final HttpClient httpClient = new DefaultHttpClient(new PoolingClientConnectionManager());
```
black-adder: According to https://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/impl/conn/PoolingClientConnectionManager.html it says it's deprecated and we should use PoolingHttpClientConnectionManager. But I don't know enough whether it's needed or not.

vprithvi: That is correct, but we are using a older version of apache httpclient where `PoolingClientConnectionManager` isn't deprecated. 
Seeing that nobody is using this, I don't see any harm in moving to the latest httpclient...

vprithvi: Our dependency on `httpclient` seems to be a mess. We are depending transitively on the `httpclient` bought in by `libthrift`, which is bad.

The right way™ to fix this would be to depend on `httpclient` directly. Because `httpclient` is very popular, we should shade `httpclient` and depend on that instead, so that our clients don't run into dependency issues. 

I'll tackle this tomorrow. @yurishkuro any objections?

yurishkuro: Maybe use okhttp instead? I think brave is using it and it's smaller than apache.

adriancole:
Not just smaller.. less likely to conflict with the application classpath!

jaeger-core/src/main/java/com/uber/jaeger/senders/HttpSender.java
pavolloffay: Ideally, it would be better to give users a choice. We could split it into a separate module.

> I think brave is using it and it's smaller than apache.

https://github.com/openzipkin/zipkin-reporter-java , there is okhttp, urlconnection...

vprithvi: @yurishkuro okhttp looks good, and a smaller size helps especially if we need to shade it.

@pavolloffay I think so as well, but I think moving to separate modules can wait until we start supporting transports other than http/udp

yurishkuro:
this is superseded by #223, @vprithvi can we close it?

Issue #15, opened 2016-07-25
Add gradle task to initialize  idl submodule


nklmish:
When I execute:

``` bash
./gradlew clean test
```

I receive

``` java
Jar was not signed since this was a snapshot release.
Jar was not signed since this was a snapshot release.
Jar was not signed since this was a snapshot release.
:jaeger-context:clean
:jaeger-core:clean
:jaeger-jaxrs2:clean UP-TO-DATE
:jaeger-context:compileJava
warning: [options] bootstrap class path not set in conjunction with -source 1.7
1 warning
:jaeger-context:licenseFormatMain UP-TO-DATE
:jaeger-context:licenseFormatTest UP-TO-DATE
:jaeger-context:licenseFormat UP-TO-DATE
:jaeger-context:processResources UP-TO-DATE
:jaeger-context:classes
:jaeger-context:compileTestJava UP-TO-DATE
:jaeger-context:processTestResources UP-TO-DATE
:jaeger-context:testClasses UP-TO-DATE
:jaeger-context:test UP-TO-DATE
:jaeger-core:compileThrift UP-TO-DATE
:jaeger-core:compileJava
warning: [options] bootstrap class path not set in conjunction with -source 1.7
/Users/nklmish/workspace/jaeger-client-java/jaeger-core/src/main/java/com/uber/jaeger/Span.java:29: error: package com.twitter.zipkin.thriftjava does not exist
import com.twitter.zipkin.thriftjava.Endpoint;
                                    ^
/Users/nklmish/workspace/jaeger-client-java/jaeger-core/src/main/java/com/uber/jaeger/Span.java:39: error: cannot find symbol
    private Endpoint peer;
            ^
  symbol:   class Endpoint
  location: class Span
/Users/nklmish/workspace/jaeger-client-java/jaeger-core/src/main/java/com/uber/jaeger/Span.java:65: error: cannot find symbol
    public Endpoint getPeer() {
           ^
  symbol:   class Endpoint
  location: class Span
/Users/nklmish/workspace/jaeger-client-java/jaeger-core/src/main/java/com/uber/jaeger/reporters/protocols/ThriftSpanConverter.java:24: error: package com.twitter.zipkin.thriftjava does not exist
import com.twitter.zipkin.thriftjava.Annotation;
                                    ^
[...]

PR #220, opened 2017-07-25
submodulesUpdate - gradle task to initialize git submodules (#15)

alshopov:
Add and document submodulesUpdate

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ This allows using Jaeger UI to find the trace by this tag.
 
 ## Developing
 
- 1. `git submodule init update`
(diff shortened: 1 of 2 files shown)


alshopov:
Decided to just create a small exec task as @yurishkuro mentioned in task rather than integrate and depend on https://github.com/ajoberstar/gradle-git

yurishkuro:
I think #219 is better, less magic.

alshopov:
In such a case close https://github.com/uber/jaeger-client-java/issues/15 perhaps?

PR #221, opened 2017-07-25
Refactor inner class to outer

alshopov:


alshopov:
Since this class does not use anything from outer class it may be static -
i.e. without reference to the outer class. When compiled it will be smaller
but the main reason is that during runtime optimizations can kick in faster.

2017-07-25 17:33 GMT+02:00 Yuri Shkuro <notifications@github.com>:

jaeger-context/src/main/java/com/uber/jaeger/context/ThreadLocalTraceContext.java
```diff
@@ -60,3 +54,11 @@ public Span getCurrentSpan() throws EmptyStackException {
     return threadLocal.get().peek();
   }
 }
+
+class StackSpan extends ThreadLocal<Stack<Span>> {
```
yurishkuro: s/StackSpan/SpanStackThreadLocal/ for better readability.

jaeger-context/src/main/java/com/uber/jaeger/context/ThreadLocalTraceContext.java
```diff
@@ -60,3 +54,11 @@ public Span getCurrentSpan() throws EmptyStackException {
...
+
+    @Override
+    public Stack<Span> initialValue() {
+        return new Stack<>();
```
yurishkuro: empty `<>` is breaking the build

vprithvi:
>Since this class does not use anything from outer class it may be static -
i.e. without reference to the outer class. 

I agree. I think it would be best to make the ThreadLocal class `static` to enforce a per-thread behavior instead of the per-thread per-instance behavior that we currently have. 

I don't see the advantage of making this an outer class and expanding the visibility of it, especially because nothing else needs to instantiate it directly. 

>When compiled it will be smaller but the main reason is that during runtime optimizations can kick in faster.

I'm not familiar with the literature behind this. Could you provide some references about these optimizations, or microbenchmarks showing improvement?
Which versions of JRE are these optimizations available on?

jaeger-context/src/main/java/com/uber/jaeger/context/ThreadLocalTraceContext.java
alshopov: Will do

jaeger-context/src/main/java/com/uber/jaeger/context/ThreadLocalTraceContext.java
alshopov: To be honest I only compiled this against jdk 1.8. Will try again with jdk 1.7

alshopov:
> I don't see the advantage of making this an outer class

Will refactor

> provide some references
# Size
```
$ javac -version
javac 1.7.0_80
```
```
$ for i in *java; do echo '--------'; cat $i; done
--------
// A.java - two separate classes
public class A{}
class A_in{}
--------
// B - inner static class
public class B{
    static class B_in{}
}
--------
// C - inner class
public class C{
    class C_in{}
}
```
```
$ javac -g:none *java
$ ls -l *class | cut -d ' ' -f 8,12
116 A.class
119 A_in.class
166 B$B_in.class
166 B.class
213 C$C_in.class
166 C.class
```
# Optimization - I am taking my words back actually in this case
```
javap -p -s -constants *class
public class A {
  public A();
    Signature: ()V
}
class A_in {
  A_in();
    Signature: ()V
}
class B$B_in {
  B$B_in();
    Signature: ()V
}
public class B {
  public B();
    Signature: ()V
}
class C$C_in {
  final C this$0;
    Signature: LC;
  C$C_in(C);
    Signature: (LC;)V
}
public class C {
  public C();
    Signature: ()V
}
```
Inner classes have references to outer but in this case outer is not used. Just checked the Hotspot source code - I thought there could be bias against inlining inner classes but this is not so: http://hg.openjdk.java.net/jdk8u/jdk8u/hotspot/file/tip/src/share/vm/opto/bytecodeInfo.cpp#l106

If the method in the inner class used things from the outer - things would have been somewhat different:

```
$ cat D.java
public class D {
    private String s = "String";
    Runnable r = new Runnable() {
        @Override
	public void run(){
	    System.out.print(D.this.s);
	}
    };
}
```
```
javac -g:none D.java
javap -p  D*class
```
```
class D$1 implements java.lang.Runnable {
  final D this$0;
  D$1(D);
  public void run();
}
public class D {
  private java.lang.String s;
  java.lang.Runnable r;
  public D();
  static java.lang.String access$000(D);
}
```
Note that in this case there is actually a call  to the synthetic `access$000` inside `run`:
```
$ javap -c D\$1.class
class D$1 implements java.lang.Runnable {
  final D this$0;

  D$1(D);
    Code:
       0: aload_0
       1: aload_1
       2: putfield      #1                  // Field this$0:LD;
       5: aload_0
       6: invokespecial #2                  // Method java/lang/Object."<init>":()V
       9: return

  public void run();
    Code:
       0: getstatic     #3                  // Field java/lang/System.out:Ljava/io/PrintStream;
       3: aload_0
       4: getfield      #1                  // Field this$0:LD;
       7: invokestatic  #4                  // Method D.access$000:(LD;)Ljava/lang/String;
      10: invokevirtual #5                  // Method java/io/PrintStream.print:(Ljava/lang/String;)V
      13: return
}
```
In that case inlining would have been harder (however in current situation it is not so)

vprithvi:
Thanks for the detailed explanation! 
I think simply adding `static` on L30 is a good solution!

jaeger-context/src/main/java/com/uber/jaeger/context/ThreadLocalTraceContext.java
```diff
@@ -59,4 +53,12 @@ public Span pop() throws EmptyStackException {
   public Span getCurrentSpan() throws EmptyStackException {
     return threadLocal.get().peek();
   }
+  
+  static class SpanStackThreadLocal extends ThreadLocal<Stack<Span>> {
```
yurishkuro: private static...

alshopov: Done

jaeger-context/src/main/java/com/uber/jaeger/context/ThreadLocalTraceContext.java
```diff
@@ -59,4 +53,12 @@ public Span pop() throws EmptyStackException {
   public Span getCurrentSpan() throws EmptyStackException {
     return threadLocal.get().peek();
   }
+  
+  private static class SpanStackThreadLocal extends ThreadLocal<Stack<Span>> {
```
vprithvi: I apologize for the confusion, but I was recommending that we maintain the old code and simply add `static` to the nested class. 
I agree that the class names are longer when we do that, but I don't see that being an issue. I don't see the advantage of expanding the visibility scope of this class.

PR #226, opened 2017-08-02
Support span log fields in zipkin sender

pvlugter:
We're using the Jaeger client to support reporting to both Jaeger and Zipkin. The Zipkin sender currently fails on spans with logs using fields, when the Annotation value (from the LogData message) is null. Recent versions of the Jaeger client also log structured data when baggage is attached, so the Zipkin sender fails on spans with baggage too.

Changes here check whether there is a message or fields, and create a single string message from the structured data fields. An alternative would be to just have the null check and only add annotations for simple message logs.

### DIFF
--- a/jaeger-zipkin/src/main/java/com/uber/jaeger/senders/zipkin/ThriftSpanConverter.java
+++ b/jaeger-zipkin/src/main/java/com/uber/jaeger/senders/zipkin/ThriftSpanConverter.java
@@ -22,6 +22,7 @@
 
 package com.uber.jaeger.senders.zipkin;
 
+import com.google.gson.Gson;
 import com.twitter.zipkin.thriftjava.Annotation;
 import com.twitter.zipkin.thriftjava.AnnotationType;
 import com.twitter.zipkin.thriftjava.BinaryAnnotation;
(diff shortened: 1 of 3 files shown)


jaeger-zipkin/src/main/java/com/uber/jaeger/senders/zipkin/ThriftSpanConverter.java
```diff
@@ -78,13 +78,39 @@
...
+        } else if (logFields != null) {
+          annotations.add(new Annotation(logData.getTime(), logFieldsAsMessage(logFields)));
+        }
       }
     }
 
     return annotations;
   }
 
+  private static String logFieldsAsMessage(Map<String, ?> logFields) {
```
yurishkuro: In other languages we used JSON encoding for structured data when converting to Zipkin. The Zipkin UI understands those and displays then better too. The openracing library for Zipkin (at least in Go) also does JSON encoding of KV logs.

pvlugter: Even better. Will switch to a JSON encoding for this. Looks like Gson is on the classpath already.

yurishkuro:
PS. @pvlugter https://github.com/uber/jaeger/issues/252 and https://github.com/uber/jaeger/issues/207

vprithvi:
lgtm

yurishkuro:
@pvlugter thanks for the fix!

PR #229, opened 2017-08-15
Add RemoteBaggageRestrictionManager

black-adder:


jaeger-core/src/main/java/com/uber/jaeger/baggage/HttpBaggageRestrictionManagerProxy.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+  }
+
+  @Override
+  public List<BaggageRestrictionResponse> getBaggageRestrictions(String serviceName)
+      throws BaggageRestrictionManagerException {
+    String jsonString;
+    try {
+      jsonString =
+          makeGetRequest("http://" + hostPort + "/baggageRestrictions?service=" + URLEncoder.encode(serviceName, "UTF-8"));
```
black-adder: I'm not using URIbuilder here since it doesn't seem to be thread safe. (Albeit the usage pattern for this class is once a minute so we shouldn't be getting into situations where thread safety is required)

jaeger-core/src/main/java/com/uber/jaeger/baggage/RemoteBaggageRestrictionManager.java
```diff
@@ -0,0 +1,141 @@
...
+    this(serviceName, proxy, metrics, denyBaggageOnInitializationFailure, refreshIntervalMs, DEFAULT_INITIAL_DELAY_MS);
+  }
+
+  RemoteBaggageRestrictionManager(
+      String serviceName,
+      BaggageRestrictionManagerProxy proxy,
+      Metrics metrics,
+      boolean denyBaggageOnInitializationFailure,
+      int refreshIntervalMs,
+      int initialDelayMs
```
black-adder: initialDelayMs is only exposed for testing purposes so I can determine when the first call to remote agent is made. Under normal operations, this class will start up and asynchronously fetch restrictions. If the user wants to know if restrictions are ready, they can check via isReady()

yurishkuro: this makes more sense as code  comment

jaeger-core/src/main/java/com/uber/jaeger/samplers/HttpSamplingManager.java
```diff
@@ -22,52 +22,27 @@
 
 package com.uber.jaeger.samplers;
 
+import static com.uber.jaeger.utils.Utils.makeGetRequest;
```
yurishkuro: seems `com.uber.jaeger.utils.HTTP` would've been a better class name

jaeger-core/src/main/java/com/uber/jaeger/samplers/HttpSamplingManager.java
```diff
@@ -22,52 +22,27 @@
...
-        rd.close();
-      }
-    } finally {
-      conn.disconnect();
-    }
-    return result.toString();
+    this.hostPort = hostPort != null ? hostPort : DEFAULT_HOST_PORT;
   }
 
   SamplingStrategyResponse parseJson(String json) {
```
yurishkuro: why not move this to util class as well, e.g. 
```
class com.uber.jaeger.utils.HTTP

public <T> T getJSON(String url, Class<T> cls) throws ...;
```

PR #230, opened 2017-08-23
Fix metric name for reporterDropped counter.

isaachier:
Based on the go client (https://github.com/uber/jaeger-client-go/blob/master/metrics.go#L63), the metric name is incorrect.


jpkrohling:
I'm closing this in favor of #335. If you think this should not be the case, please reopen and leave a comment.

PR #233, opened 2017-09-15
Senders fix, do not use static thrift factory

pavolloffay:
When I was using multiple tracers in one process `HttpSender` was sending corrupted data.

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/senders/HttpSender.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/senders/HttpSender.java
@@ -36,15 +36,13 @@ import okhttp3.Response;
 import org.apache.thrift.TException;
 import org.apache.thrift.TSerializer;
 import org.apache.thrift.protocol.TBinaryProtocol;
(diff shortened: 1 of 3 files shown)


pavolloffay:
We would have to emulate deserialization on the server size. 

Could we get it in and if necessary write test in separate PR? It's blocking my work `spark-depenencies` PR https://github.com/pavolloffay/spark-dependencies/pull/1.

pavolloffay:
Once this is merged. Could you please do micro release? (I don't have rights and @jpkrohling  will be here on Wednesday)

vprithvi:
Yeah, we can write a test as a separate PR. I'll merge and release

yurishkuro:
@pavolloffay are you saying the factory is not thread safe?

pavolloffay:
I haven't digged very deep. I will write a test later and look more closely.

vprithvi:
The serializer isn't threadsafe, but looking at https://github.com/apache/thrift/blob/master/lib/java/src/org/apache/thrift/protocol/TBinaryProtocol.java#L85 it seems that we should be able to continue having the factory as static.

PR #236, opened 2017-09-19
Series of updates

alshopov:
I have divided the updates to the many components in different patches for ease of reviewing.
One of the principles was nonruntime dependencies to runtime one
Another was the probability of affecting other projects. 

Just mentioning that both styles of dependencies (abbreviated and full_ is currently used and I have not changed this in this series.

### DIFF
--- a/jaeger-b3/src/main/java/com/uber/jaeger/propagation/b3/B3TextMapCodec.java
+++ b/jaeger-b3/src/main/java/com/uber/jaeger/propagation/b3/B3TextMapCodec.java
@@ -68,7 +68,8 @@ public final class B3TextMapCodec implements Injector<TextMap>, Extractor<TextMa
     byte flags = 0;
     for (Map.Entry<String, String> entry : carrier) {
       if (entry.getKey().equalsIgnoreCase(SAMPLED_NAME)) {
-        if (entry.getValue().equals("1") || entry.getValue().toLowerCase().equals("true")) {
+        String value = entry.getValue();
+        if ("1".equals(value) || "true".equalsIgnoreCase(value)) {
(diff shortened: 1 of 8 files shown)


build.gradle
```diff
@@ -7,7 +7,9 @@ plugins {
     id 'com.github.sherter.google-java-format' version '0.3.2'
     id "com.github.johnrengelman.shadow" version "2.0.1"
     id "net.ltgt.errorprone" version "0.0.10"
-    id 'ru.vyarus.animalsniffer' version '1.3.0'
+    id 'ru.vyarus.animalsniffer' version '1.4.2'
+    // './gradlew dependencyUpdates' will give you the updates of dependencies
```
vprithvi: `./gradlew tasks` should list this documentation, I don't see the utility of this comment

build.gradle
```diff
@@ -154,7 +155,8 @@ task printVersion {
 configure(subprojects.findAll {it.name != 'jaeger-thrift'}) {
    apply plugin: 'net.ltgt.errorprone'
    dependencies {
-     errorprone 'com.google.errorprone:error_prone_core:2.0.15'
+     // stay on 2.0.5, 2.1.1 throws on analysing
+     errorprone 'com.google.errorprone:error_prone_core:2.0.5'
```
vprithvi: We shouldn't be downgrading `error_prone` because we might be losing some of the more recent checks.

build.gradle
alshopov: will remove

build.gradle
alshopov: I agree, .15 -> .5 was a typo

PR #237, opened 2017-09-20
Series of small patches with small improvemts grouped by type

alshopov:


jaeger-context/src/main/java/com/uber/jaeger/context/ThreadLocalTraceContext.java
```diff
@@ -27,7 +27,7 @@
 import java.util.Stack;
 
 public class ThreadLocalTraceContext implements TraceContext {
-  private final ThreadLocal<Stack<Span>> threadLocal = new SpanStackThreadLocal();
+  private static final ThreadLocal<Stack<Span>> threadLocal = new SpanStackThreadLocal();
```
yurishkuro: this is a change in semantics - there's no requirement that all instances of this class share the same thread-local

jaeger-core/src/main/java/com/uber/jaeger/propagation/TextMapCodec.java
```diff
@@ -75,7 +76,7 @@ public SpanContext extract(TextMap carrier) {
     String debugId = null;
     for (Map.Entry<String, String> entry : carrier) {
       // TODO there should be no lower-case here
-      String key = entry.getKey().toLowerCase();
+      String key = entry.getKey().toLowerCase(Locale.ROOT);
```
yurishkuro: -1 : I don't know that root even means here, but the no-arg toLowerCase () method clearly says "using the default locale"

jaeger-core/src/main/java/com/uber/jaeger/utils/Utils.java
```diff
@@ -27,10 +27,11 @@
 
 import java.net.InetAddress;
 import java.net.UnknownHostException;
+import java.util.Locale;
 
 public class Utils {
   public static String normalizeBaggageKey(String key) {
-    return key.replaceAll("_", "-").toLowerCase();
+    return key.replaceAll("_", "-").toLowerCase(Locale.ROOT);
```
yurishkuro: same

jaeger-zipkin/src/main/java/com/uber/jaeger/senders/zipkin/ThriftSpanEncoder.java
```diff
@@ -58,7 +58,7 @@ public Encoding encoding() {
     }
   }
 
-  final ThreadLocal<ReusableTBinaryProtocol> protocol =
+  final static ThreadLocal<ReusableTBinaryProtocol> protocol =
```
yurishkuro: again, not necessary, and potentially not safe since different instances will be reusing the same protocol

jaeger-core/src/main/java/com/uber/jaeger/propagation/TextMapCodec.java
alshopov: Can you read the whole javadoc for this method and not only the title sentence?
https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#toLowerCase--
It gives example of what is wrong and how one should use toLowerCase without surprises.

yurishkuro: fair enough, +1

jaeger-context/src/main/java/com/uber/jaeger/context/ThreadLocalTraceContext.java
alshopov: You are absolutely right, I am removing this. Basically you bind thread locals per instance.

yurishkuro:
Thanks!

PR #242, opened 2017-09-25
Tracer implements Closeable

pavolloffay:


jpkrohling:
Shouldn't be hard to write a test for this :) Other than that, LGTM

jpkrohling:
I would expect the test to be like the `testClose`, but with a `try-with-resources` approach. But ensuring it implements the proper interface + keeping the `testClose` would be enough.

pavolloffay:
Can somebody merge? (I don't have write access)

jpkrohling:
Looks like I can't:

image

yurishkuro:
any idea what's up with coverage? why does it show a drop in a file that should not be affected?

PR #247, opened 2017-10-03
Use grizzly tests for jaxrs2 instrumentation

### LINKED ISSUE #204: Make the span active after extraction within the jersey filter
Hello,

Maybe I misunderstand the way the span needs to be used but, in jaeger-jaxrs2, when using the com.uber.jaeger.filters.jaxrs2.ServerFilter, the span created after extraction of the request headers is currently not made active within the tracer instance.

```
SpanContext spanContext = tracer.extract(Format.Builtin.HTTP_HEADERS, carrier);
if (spanContext != null) {
    builder = builder.asChildOf(spanContext);
}
Span serverSpan = builder.startManual();

// the following line is missing
tracer.makeActive(serverSpan);

traceContext.push(serverSpan);
```

Therefor, when building a new span within the application, a new span is created instead of a child of the one extracted from the request.

```
try(ActiveSpan activeSpan = tracer.buildSpan("jumpcargo-my-friend").startActive()) {
}
```

Version of opentracing: 0.30.0
Version of jaeger-client-java: 0.20.0

vprithvi:
- Add grizzly tests for jaxrs2 instrumentation

### DIFF
--- a/jaeger-jaxrs2/build.gradle
+++ b/jaeger-jaxrs2/build.gradle
@@ -8,6 +8,7 @@ dependencies {
     testCompile project(path: ':jaeger-core', configuration: 'tests')
     testCompile group: 'org.glassfish.jersey.containers', name: 'jersey-container-grizzly2-http', version: jerseyVersion
     testCompile group: 'org.glassfish.jersey.containers', name: 'jersey-container-servlet', version: jerseyVersion
+    testCompile group: 'org.glassfish.jersey.test-framework.providers', name: 'jersey-test-framework-provider-grizzly2', version: jerseyVersion
     testCompile group: 'org.glassfish.jersey.core', name: 'jersey-client', version: jerseyVersion
     testCompile group: 'org.glassfish.jersey.media', name: 'jersey-media-json-jackson', version: jerseyVersion
     testCompile group: 'org.glassfish.jersey.media', name: 'jersey-media-moxy', version: jerseyVersion
--- a/jaeger-jaxrs2/src/test/java/com/uber/jaeger/filters/jaxrs2/ServerFilterTest.java
+++ b/jaeger-jaxrs2/src/test/java/com/uber/jaeger/filters/jaxrs2/ServerFilterTest.java
@@ -14,90 +14,148 @@
 
 package com.uber.jaeger.filters.jaxrs2;
 
-import static org.junit.Assert.assertEquals;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
-
 import com.uber.jaeger.Constants;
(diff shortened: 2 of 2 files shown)


jaeger-jaxrs2/src/main/java/com/uber/jaeger/filters/jaxrs2/ServerFilter.java
```diff
@@ -85,6 +85,8 @@ public void filter(ContainerRequestContext containerRequestContext) throws IOExc
       }
       Span serverSpan = builder.startManual();
 
+      tracer.makeActive(serverSpan);
+
       traceContext.push(serverSpan);
```
yurishkuro: I don't think using both the traceContext and active span is a good idea, they can easily diverge. I thought we would reimplement the trace context via the active span API.

vprithvi: Oh, I misunderstood. I had made an assumption that the ActiveSpanSource used Jaeger's ThreadLocal storage implementation, but I see that this is not the case. 

I feel that having Jaeger's `ThreadLocalTraceContext` implement `io.opentracing.ActiveSpanSource` might work better for interoperability with code using older instrumentation. I think that we should switch to using `io.opentracing.util.ThreadLocalActiveSpanSource` when we update instrumentation. 

Is this what you were proposing?

yurishkuro: I was thinking `ThreadLocalTraceContext` would delegate to `ActiveSpanSource`, but implementing it might be even better, although I would still prefer that implementation delegates to the existing ref-counting OT implementation of the `ActiveSpanSource`. The main trick is that currently `ThreadLocalTraceContext` is independent of the tracer, we may need to make a breaking API change to force them to be bound together.

vprithvi: Sounds reasonable, I'll attempt to reuse what we can, and put my work up in a separate commit. (Will update this once that is merged)

I haven't thought through the API changes yet, but I'm hoping that it ends up being non breaking for the users who only access the `ThreadLocalTraceContext` statically through `TracingUtils`.

yurishkuro: > who only access the ThreadLocalTraceContext statically through TracingUtils.

+1

pavolloffay:
jax-rs instrumentation has been moved to legacy-client-java repo.

Issue #240, opened 2017-09-21
[Apache] Split out span creation from baggage injection for interceptors


dray92:
I think we need to split out the logic in the Request Interceptor. [`Span creation`](https://github.com/uber/jaeger-client-java/blob/master/jaeger-apachehttpclient/src/main/java/com/uber/jaeger/httpclient/TracingRequestInterceptor.java#L69) and [`injection`](https://github.com/uber/jaeger-client-java/blob/master/jaeger-apachehttpclient/src/main/java/com/uber/jaeger/httpclient/TracingRequestInterceptor.java#L81) might be a little **too** coupled. 

Suggested solution: 
 - Have one interceptor tasked with starting span -- will be added to HttpProcessor chain using [`addInterceptorFirst(HttpRequestInterceptor ..)`](https://hc.apache.org/httpcomponents-asyncclient-dev/httpasyncclient/apidocs/org/apache/http/impl/nio/client/HttpAsyncClientBuilder.html#addInterceptorFirst(org.apache.http.HttpRequestInterceptor)).
Say, `JaegerSpanCreationRequestInterceptor`
 - Have second interceptor that is tasked with span baggage injection -- will be added to the **end** of the HttpProcessor chain using [`addInterceptorLast(HttpRequestInterceptor ..)`](https://hc.apache.org/httpcomponents-asyncclient-dev/httpasyncclient/apidocs/org/apache/http/impl/nio/client/HttpAsyncClientBuilder.html#addInterceptorLast(org.apache.http.HttpRequestInterceptor)).
Say, `JaegerSpanInjectionRequestInterceptor`

This can be default behavior for something like [`TracingInterceptors.java`](https://github.com/uber/jaeger-client-java/blob/master/jaeger-apachehttpclient/src/main/java/com/uber/jaeger/httpclient/TracingInterceptors.java) - but we have to find a way to enable some interceptors to come in between the span creation and injection stages.

Cleaner way below:

As a consumer, I can do something like 
```
HttpProcessor inHttpProc = new ImmutableHttpProcessor(
    new HttpRequestInterceptor[] {
        new JaegerSpanCreationRequestInterceptor(tracer),
        new PreprocessingRequestInterceptor(),
        new AuthRequestInterceptor(),
        new PostprocessingRequestInterceptor(),
        new JaegerSpanInjectionRequestInterceptor(tracer)
    }
);
```

Then, client initialization can be something like,
```
[...]

PR #248, opened 2017-10-03
Request for comments: Breaking apart functionality of TracingRequestInterceptor

dray92:
For #240

### DIFF
--- a/jaeger-apachehttpclient/src/main/java/com/uber/jaeger/httpclient/SpanCreationRequestInterceptor.java
+++ b/jaeger-apachehttpclient/src/main/java/com/uber/jaeger/httpclient/SpanCreationRequestInterceptor.java
@@ -1,0 +1,114 @@
+/*
+ * Copyright (c) 2017, Uber Technologies, Inc
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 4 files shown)


jaeger-apachehttpclient/src/main/java/com/uber/jaeger/httpclient/TracingInterceptors.java
```diff
@@ -32,13 +32,15 @@
 public class TracingInterceptors {
   public static HttpAsyncClientBuilder addTo(HttpAsyncClientBuilder clientBuilder, Tracer tracer) {
     return clientBuilder
```
yurishkuro: shouldn't `TracingRequestInterceptor` be also changed in this PR to invoke the two new interceptors? And this class won't need to change.

dray92: So `TracingRequestInterceptor` is the one that was introducing the coupling. I chose to leave it as is, since there might be existing Jaeger users that are directly using that to instrument their clients. Given that you bring up this point, a change can be made to [`TracingRequestInterceptor.process(..)`](https://github.com/jaegertracing/jaeger-client-java/blob/master/jaeger-apachehttpclient/src/main/java/com/uber/jaeger/httpclient/TracingRequestInterceptor.java#L58), to have instances of `SpanCreationRequestInterceptor` and `SpanInjectionRequestInterceptor`, and then invoking the respective `process` methods serially. This would look something like:

```
public class TracingRequestInterceptor implements HttpRequestInterceptor {
  private final Tracer tracer;
  private final SpanCreationRequestInterceptor spanCreationInterceptor;
  private final SpanInjectionRequestInterceptor spanInjectionInterceptor;

  public TracingRequestInterceptor(Tracer tracer) {
    this.tracer = tracer;
    this.spanCreationInterceptor = new SpanCreationRequestInterceptor(tracer);
    this.spanInjectionInterceptor = new SpanInjectionRequestInterceptor(tracer);
  }

  @Override
  public void process(HttpRequest httpRequest, HttpContext httpContext)
      throws HttpException, IOException {
    this.spanCreationInterceptor.process(httpRequest, httpContext);
    this.spanInjectionInterceptor.process(httpRequest, httpContext);
  }
  ..........
}
``` 


Since `TracingInterceptors` is the guy responsible for giving the "complete package", I made the change in for this to use the new interceptors, and thus, proving complete backwards compatibility.

yurishkuro: > This would look something like:

Yes, that's what I think we should have. Currently you copy & pasted the same code, instead of refactoring to have a single instance of the actual code.

debosmit:
At time of last conversation, @vprithvi deferred to @yurishkuro and @black-adder, since he does not have enough context. FWIW, this is doing to apache, exactly what #273 did to jax-rs2.

PR #255, opened 2017-10-06
Use io.opentracing.GlobalTracer instead of singleton in Configuration

vprithvi:
- Register the tracer created by com.uber.jaeger.Configuration#getTracer
  with opentracing's GlobalTracer mechanism so that opentracing instrumentations
  may find it.

- Disable the lazily loaded singleton field by default.

This is a prerequisite for #257

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
@@ -188,42 +188,21 @@ public class Configuration {
   }
 
   public synchronized io.opentracing.Tracer getTracer() {
-    if (disableGlobalTracer) {
-      return getLocalTracer();
-    }
-
-    return getGlobalTracer();
-  }
-
-  private io.opentracing.Tracer getGlobalTracer() {
-    if (GlobalTracer.isRegistered()) {
-      return GlobalTracer.get();
-    }
-
-    Tracer tracer = getTracerBuilder().build();
-    GlobalTracer.register(tracer);
-    log.info("Initialized and registered global tracer={}", tracer);
-    return tracer;
-  }
-
-  @Deprecated
-  private io.opentracing.Tracer getLocalTracer() {
     if (tracer != null) {
       return tracer;
     }
 
     tracer = getTracerBuilder().build();
     log.info("Initialized tracer={}", tracer);
+
+    if (!(disableGlobalTracer || GlobalTracer.isRegistered())) {
+      GlobalTracer.register(tracer);
+    }
+
     return tracer;
   }
 
   public synchronized void closeTracer() {
-    if (GlobalTracer.isRegistered()) {
-      Tracer tracer = (Tracer) GlobalTracer.get();
-      tracer.close();
-      return;
-    }
-
     if (tracer != null) {
       tracer.close();
     }
--- a/jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
@@ -38,6 +38,7 @@ import com.uber.jaeger.samplers.RemoteControlledSampler;
 import com.uber.jaeger.samplers.Sampler;
 import com.uber.jaeger.senders.Sender;
 import com.uber.jaeger.senders.UdpSender;
+import io.opentracing.util.GlobalTracer;
 import java.text.NumberFormat;
 import java.text.ParseException;
 import java.util.HashMap;
@@ -105,6 +106,11 @@ public class Configuration {
   public static final String JAEGER_TAGS = JAEGER_PREFIX + "TAGS";
 
   /**
+   * Disables registration with {@link GlobalTracer}.
+   */
+  public static final String JAEGER_DISABLE_GLOBAL_TRACER = JAEGER_PREFIX + "DISABLE_GLOBAL_TRACER";
+
+  /**
    * The serviceName that the tracer will use
    */
   private final String serviceName;
@@ -119,7 +125,13 @@ public class Configuration {
   private StatsFactory statsFactory;
 
   /**
-   * lazy singleton Tracer initialized in getTracer() method
+   * Use {@link GlobalTracer} to store the tracer instead of
+   * storing an instance of the tracer in {@link #tracer}.
+   */
+  private final boolean disableGlobalTracer;
+
+  /**
+   * lazy singleton Tracer initialized in getTracer() method.
    */
   private Tracer tracer;
 
@@ -131,6 +143,14 @@ public class Configuration {
       String serviceName,
       SamplerConfiguration samplerConfig,
       ReporterConfiguration reporterConfig) {
+    this(serviceName, samplerConfig, reporterConfig, true);
+  }
+
+  public Configuration(
+      String serviceName,
+      SamplerConfiguration samplerConfig,
+      ReporterConfiguration reporterConfig,
+      boolean disableGlobalTracer) {
     if (serviceName == null || serviceName.isEmpty()) {
       throw new IllegalArgumentException("Must provide a service name for Jaeger Configuration");
     }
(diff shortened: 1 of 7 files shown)


jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
```diff
@@ -165,6 +188,26 @@ public static Configuration fromEnv() {
...
+  private io.opentracing.Tracer getGlobalTracer() {
+    if (GlobalTracer.isRegistered()) {
+      return GlobalTracer.get();
+    }
+
+    Tracer tracer = getTracerBuilder().build();
+    GlobalTracer.register(tracer);
+    log.info("Initialized and registered global tracer={}", tracer);
+    return tracer;
+  }
```
yurishkuro: this seems unnecessary complicated - why not keep the single getTracer method and in the lazy init path (otherwise unchanged) just add `if (!disableGlobalTracer) { GlobalTracer.register(tracer);}`? Everything else keeps working off the local `tracer` var.

jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
```diff
@@ -131,6 +143,14 @@ public Configuration(
       String serviceName,
       SamplerConfiguration samplerConfig,
       ReporterConfiguration reporterConfig) {
+    this(serviceName, samplerConfig, reporterConfig, true);
```
yurishkuro: didn't we want to default this to `disableGlobalTracer = false`?

also, let's stop adding constructors and add a builder, which we should've done in the first place. The new constructor below should be private.

jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
```diff
@@ -148,13 +168,16 @@ public Configuration(
...
+    this.disableGlobalTracer = disableGlobalTracer;
   }
 
   public static Configuration fromEnv() {
     return new Configuration(
         getProperty(JAEGER_SERVICE_NAME),
         SamplerConfiguration.fromEnv(),
-        ReporterConfiguration.fromEnv());
+        ReporterConfiguration.fromEnv(),
+        Boolean.valueOf(getProperty(JAEGER_DISABLE_GLOBAL_TRACER)));
```
yurishkuro: use getPropertyAsBoolean

yurishkuro:
please link to the relevant issue (iirc this is a pre-requisite)

jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
```diff
@@ -119,7 +125,13 @@
   private StatsFactory statsFactory;
 
   /**
-   * lazy singleton Tracer initialized in getTracer() method
+   * Use {@link GlobalTracer} to store the tracer instead of
+   * storing an instance of the tracer in {@link #tracer}.
+   */
+  private final boolean disableGlobalTracer;
```
mabn: This comment should be negated: "Do not use.."

jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
vprithvi: `getPropertyAsBoolean` returns null when the property isn't set, which isn't what we want here.

jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
vprithvi: While I agree, I think it should be part of a separate commit

jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
vprithvi: I wrote it this way because I wanted to remove tracer from jaeger config in a future release, and instead depend on opentracing's GlobalTracer for management of the singleton. 

Do you think we need to continue support holding on to a tracer in the config?
 
The only case where I see this being useful is for advanced users who need multiple tracers in the same application. I expect that such users to use `getTracerBuilder` and not maintain multiple copies of JaegerConfiguration.

yurishkuro: if or when we decide on that separation (I am not convinced of its value), we can do that, but right now I don't see a value in this big change that can be instead achieved with just 3 lines.

yurishkuro:
> While I agree, I think it should be part of a separate commit

if you want to do a separate pr, it's fine, but right now you're adding a new public constructor that we won't be able to hide later for backwards compatibility.

jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
vprithvi: ok

jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
```diff
@@ -402,11 +430,7 @@ private static Number getPropertyAsNum(String name) {
   }
 
   private static Boolean getPropertyAsBoolean(String name) {
-    String value = getProperty(name);
-    if (value != null) {
-      return Boolean.valueOf(value);
-    }
-    return null;
+    return Boolean.valueOf(getProperty(name));
```
yurishkuro: You're changing the semantics of the method, previously it could return null, now it performs defaulting of null string to `false`. It still works given how it's used in this class, but by accident. I would rename it to `getPropertyAsBool` and return `bool`, and add a comment about defaulting.

jaeger-core/src/test/java/com/uber/jaeger/ConfigurationTest.java
```diff
@@ -54,14 +57,36 @@ public void clearProperties() {
     System.clearProperty(Configuration.JAEGER_SAMPLER_MANAGER_HOST_PORT);
     System.clearProperty(Configuration.JAEGER_SERVICE_NAME);
     System.clearProperty(Configuration.JAEGER_TAGS);
+    System.clearProperty(Configuration.JAEGER_DISABLE_GLOBAL_TRACER);
 
     System.clearProperty(TEST_PROPERTY);
+
+    // Hack to reset opentracing's global tracer
```
yurishkuro: I wouldn't call it a hack

jaeger-core/src/test/java/com/uber/jaeger/ConfigurationTest.java
```diff
@@ -54,14 +57,36 @@ public void clearProperties() {
...
+    System.setProperty(Configuration.JAEGER_SERVICE_NAME, "Test");
+    System.setProperty(Configuration.JAEGER_DISABLE_GLOBAL_TRACER, "true");
+    assertNotNull(Configuration.fromEnv().getTracer());
+    assertFalse(GlobalTracer.isRegistered());
+  }
+
+  @Test
+  public void testDefaultGlobalTracer() {
+    Configuration config = new Configuration("Test");
+    config.getTracer();
```
yurishkuro: maybe assertNotNull? Otherwise it looks weird.

jaeger-dropwizard/src/test/java/com/uber/jaeger/dropwizard/JerseyServerFilterTest.java
```diff
@@ -48,21 +49,12 @@
...
     reporter = new InMemoryReporter();
-    com.uber.jaeger.Tracer tracer = (com.uber.jaeger.Tracer) config.getTracer();
-    try {
-      Field reporter = com.uber.jaeger.Tracer.class.getDeclaredField("reporter");
-      reporter.setAccessible(true);
-      reporter.set(tracer, this.reporter);
-    } catch (NoSuchFieldException | IllegalAccessException e) {
-      throw new RuntimeException(e);
-    }
+    Tracer tracer = new Tracer.Builder("world service", reporter, new ConstSampler(true)).build();
```
yurishkuro: nice! don't know what ^^ was about.

jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
```diff
@@ -171,6 +194,11 @@ public static Configuration fromEnv() {
 
     tracer = getTracerBuilder().build();
     log.info("Initialized tracer={}", tracer);
+
+    if (!(disableGlobalTracer || GlobalTracer.isRegistered())) {
```
pavolloffay: The same tracer instance can be registered multiple times.

I am wondering whether it would be better to remove `GlobalTracer.isRegistered()` check and propagate potential issues with multiple registered traces at process start time.

jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
vprithvi: True, I didn't see any cases where it was necessary to differentiate null from false. I'll update as you recommended.

jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
vprithvi: Oh, you mean that if somebody accidentally calls `fromEnv()` many times, it'll lead to drift between the `GlobalTracer` and the local tracer.
 
I agree, I think it is better to error out. I'll update the code.

PR #260, opened 2017-10-07
This closes #256 by better handling baggage null values

### LINKED ISSUE #256: Null baggage item causes NPE
Adding a baggage item with a null value throws an exception during serialization:

Caused by: java.lang.NullPointerException
	at java.net.URLEncoder.encode(URLEncoder.java:204)
	at com.uber.jaeger.propagation.TextMapCodec.encodedValue(TextMapCodec.java:124)
	at com.uber.jaeger.propagation.TextMapCodec.inject(TextMapCodec.java:67)
	at com.uber.jaeger.propagation.TextMapCodec.inject(TextMapCodec.java:34)
	at com.uber.jaeger.Tracer.inject(Tracer.java:159)
	at io.opentracing.util.GlobalTracer.inject(GlobalTracer.java:118)
	at io.opentracing.contrib.jaxrs2.client.ClientTracingFilter.filter(ClientTracingFilter.java:76)
	at org.jboss.resteasy.client.jaxrs.internal.ClientInvocation.invoke(ClientInvocation.java:413)

Ideally, the baggage items would appear for the recipient with a null value, but FWIW, at least the RESTEasy implementation appears to drop any request headers with a null value so this may be difficult and/or inconsistent. Ignoring a null value may not be ideal, but the current behavior can be very problematic.

bmozaffa:
This closes #256 by ignoring attempts to set a new baggage value of null, but also removing existing baggage when their value has been updated as null

Signed-off-by: Babak Mozaffari <bmozaffa@redhat.com>

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/Span.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Span.java
@@ -122,7 +122,8 @@ public class Span implements io.opentracing.Span {
 
   @Override
   public Span setBaggageItem(String key, String value) {
-    if (key == null || value == null) {
+    if (key == null || (value == null && context.getBaggageItem(key) == null)) {
+      //Ignore attempts to add new baggage items with null values, they're not accessible anyway
       return this;
     }
     synchronized (this) {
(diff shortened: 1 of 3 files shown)


yurishkuro:
@bmozaffa would you like to rebase from master and resubmit?

PR #261, opened 2017-10-07
Make gradle use travis/docker-thrift/thrift script

yurishkuro:
Right now the thrift plugin uses the default location of thrift compiler, which forces us to override the compiler in Travis, and locally you just hope that you have the correct version (0.9.2) installed.

This change tells the plugin to use the wrapper script we have in `travis/docker-thrift/thrift` that uses a thrift compiler from a Docker image.

It means in order to build the project locally one must have Docker installed.

### DIFF
--- a/jaeger-thrift/build.gradle
+++ b/jaeger-thrift/build.gradle
@@ -18,6 +18,7 @@ dependencies {
  * `brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/56d8c1eba1e5ac30290dd0c486f4bba37f821e42/Formula/thrift.rb`
  */
 compileThrift {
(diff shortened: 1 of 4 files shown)


jaeger-thrift/build.gradle
```diff
@@ -18,6 +18,7 @@ dependencies {
  * `brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/56d8c1eba1e5ac30290dd0c486f4bba37f821e42/Formula/thrift.rb`
```
vprithvi: Is this comment still accurate?

yurishkuro: no, it's not

PR #266, opened 2017-10-09
Add a ActiveSpanSource backed TraceContext

vprithvi:
- use `ActiveSpanSource` for propagation instead of `ThreadLocalTraceContext`

Signed-off-by: Prithvi Raj <p.r@uber.com>


jaeger-context/src/main/java/com/uber/jaeger/context/ActiveSpanSourceTraceContext.java
```diff
@@ -0,0 +1,91 @@
...
+
+  @Override
+  public boolean isEmpty() {
+    return tracer.activeSpan() == null;
+  }
+
+  private Span getSpan(ActiveSpan activeSpan) {
+    Span wrappedSpan;
+    try {
+      wrappedSpan = (Span) ActiveSpanSourceTraceContext.wrappedSpan.get(activeSpan);
```
yurishkuro: unnecessary qualifier `ActiveSpanSourceTraceContext.`

jaeger-context/src/main/java/com/uber/jaeger/context/ActiveSpanSourceTraceContext.java
```diff
@@ -0,0 +1,91 @@
...
+ * This is a {@link TraceContext} that relies on the {@link ActiveSpanSource} registered with {@link
+ * GlobalTracer}.
+ */
+public class ActiveSpanSourceTraceContext implements TraceContext {
+
+  private final Tracer tracer;
+  /**
+   * This is a hack to retrieve the span wrapped by the {@link ThreadLocalActiveSpan} implementation
+   * to shoehorn into the {@link TraceContext} implementation. This is being done so that
+   * instrumentation relying on {@link Tracer} has a is consistent with instrumentation using
```
yurishkuro: typo: "has a is"

jaeger-context/src/main/java/com/uber/jaeger/context/ActiveSpanSourceTraceContext.java
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * This is a {@link TraceContext} that relies on the {@link ActiveSpanSource} registered with {@link
+ * GlobalTracer}.
+ */
+public class ActiveSpanSourceTraceContext implements TraceContext {
+
+  private final Tracer tracer;
+  /**
+   * This is a hack to retrieve the span wrapped by the {@link ThreadLocalActiveSpan} implementation
```
yurishkuro: let's mention that this is a temporary hack until 0.31

jaeger-context/src/main/java/com/uber/jaeger/context/ActiveSpanSourceTraceContext.java
```diff
@@ -0,0 +1,91 @@
...
+import io.opentracing.util.GlobalTracer;
+import io.opentracing.util.ThreadLocalActiveSpan;
+
+import java.lang.reflect.Field;
+
+/**
+ * This is a {@link TraceContext} that relies on the {@link ActiveSpanSource} registered with {@link
+ * GlobalTracer}.
+ */
+public class ActiveSpanSourceTraceContext implements TraceContext {
```
yurishkuro: Are you sure we need a new class instead of changing ThreadLocalTraceContext?  If someone is already using ThreadLocalTraceContext directly, then it won't work correctly any longer, so I think it warrants a breaking API change (namely having an argument to the constructor).

jaeger-context/src/main/java/com/uber/jaeger/context/ActiveSpanSourceTraceContext.java
```diff
@@ -0,0 +1,91 @@
...
+  static {
+    try {
+      wrappedSpan = ThreadLocalActiveSpan.class.getDeclaredField("wrapped");
+      wrappedSpan.setAccessible(true);
+    } catch (NoSuchFieldException e) {
+      throw new RuntimeException("Unable to access ThreadLocalActiveSpan.wrapped reflectively.", e);
+    }
+  }
+
+  public ActiveSpanSourceTraceContext(Tracer tracer) {
```
yurishkuro: * I think you can use a more narrow dependency, ActiveSpanSource instead of Tracer.
  * instead of depending on ActiveSpanSource, it makes sense to depend directly on ThreadLocalActiveSpanSource

jaeger-context/src/main/java/com/uber/jaeger/context/TracingUtils.java
```diff
@@ -14,10 +14,19 @@
...
 
 public class TracingUtils {
-  private static final TraceContext traceContext = new ThreadLocalTraceContext();
+  private static final TraceContext traceContext;
+
+  static {
+    if (!GlobalTracer.isRegistered()) {
+      throw new IllegalStateException("Please register a io.opentracing.util.GlobalTracer");
+    }
+    traceContext = new ActiveSpanSourceTraceContext(GlobalTracer.get());
```
yurishkuro: I am concerned with the race conditions here, as you don't know whether TracingUtils class will be loaded before we have a chance to init global tracer. You may need to change this into a lazy dependency, i.e. instead of passing tracer (or active span source, per the other comment) to the TraceContext impl, you mass a _function that returns an active span source_, so that it's dereferenced as needed.

jaeger-jaxrs2/src/main/java/com/uber/jaeger/filters/jaxrs2/ServerFilter.java
```diff
@@ -94,10 +94,10 @@ public void filter(
         // hitting this case means previous filter was not called
         return;
       }
-      serverSpan = traceContext.pop();
+      serverSpan = traceContext.getCurrentSpan();
 
       Tags.HTTP_STATUS.set(serverSpan, containerResponseContext.getStatus());
-      serverSpan.finish();
+      traceContext.pop();
```
yurishkuro: please add comments explaining this order (i.e. that we cannot call `pop` as before since it will deactivate & finish the span).

yurishkuro: is this the only place in the whole project where we call `pop`?

jaeger-context/src/main/java/com/uber/jaeger/context/ActiveSpanSourceTraceContext.java
vprithvi: We could also simply remove the `ThreadLocalTraceContext` instead of marking it `@Deprecated`

jaeger-jaxrs2/src/main/java/com/uber/jaeger/filters/jaxrs2/ServerFilter.java
vprithvi: Will do. No, we call it in our wrapped Callable and Runnable

jaeger-context/src/main/java/com/uber/jaeger/context/TracingUtils.java
vprithvi: I thought this was was fine because `GlobalTracer.get()` returns an instance that is a new reference to itself. However, I agree that we cannot guarantee that TracingUtil is loaded after the tracer is initialized. I'll move this to a lazy init block.

jaeger-context/src/main/java/com/uber/jaeger/context/ActiveSpanSourceTraceContext.java
vprithvi: >instead of depending on ActiveSpanSource, it makes sense to depend directly on ThreadLocalActiveSpanSource

This is a bit involved because we need to use reflection to get the instance of the `ThreadLocalActiveSpanSource` from the tracer. I updated to use the `ActiveSpanSource` interface instead.

jaeger-context/src/main/java/com/uber/jaeger/context/ActiveSpanSourceTraceContext.java
yurishkuro: why create extra churn with classes/filenames/history when we can reuse the same class and re-implement it as needed?

jaeger-context/src/main/java/com/uber/jaeger/context/TracingUtils.java
```diff
@@ -14,18 +14,31 @@
...
     return traceContext;
   }
 
   public static ExecutorService tracedExecutor(ExecutorService wrappedExecutorService) {
+    initializeTraceContext();
     return new TracedExecutorService(wrappedExecutorService, traceContext);
   }
 
-  private TracingUtils(){}
+  private static synchronized void initializeTraceContext() {
```
yurishkuro: in my suggestion there was no additional `synchronized` on the hot path

jaeger-context/src/main/java/com/uber/jaeger/context/ActiveSpanSourceTraceContext.java
yurishkuro: you don't need reflection, just a cast: `(ThreadLocalActiveSpanSource) tracer.activeSpanSource()`

jaeger-context/src/main/java/com/uber/jaeger/context/ActiveSpanSourceTraceContext.java
vprithvi: Could you provide justification for sticking with the old name? 

The new name is more descriptive. 

Javac itself doesn't care about what this class used to be called. 

As far as git goes, it uses binary blobs to track file content. I expect it to be able to handle this operation efficiently. Have you seen examples where it hasn't? Should we be optimizing for git history size?

jaeger-context/src/main/java/com/uber/jaeger/context/TracingUtils.java
vprithvi: I apologize, I didn't understand your previous suggestion, could you elaborate?

As such, `getTraceContext` and `tracedExecutor` are only used when initializing filters and executors. I expect these methods to be called a handful of times during application startup.

jaeger-context/src/main/java/com/uber/jaeger/context/ActiveSpanSourceTraceContext.java
vprithvi: `tracer.activeSpanSource` isn't public. Further, it is defined in `com.uber.jaeger.Tracer` in `jaeger-core`, for which we'd have to add a dependency from `jaeger-context` instead of depending on `opentracing-api`.

jaeger-context/src/main/java/com/uber/jaeger/context/TracingUtils.java
yurishkuro: instead of `ActiveSpanSourceTraceContext(Tracer)` declare `ActiveSpanSourceTraceContext(TracerProvider tracerProvider)` and then in the implementation every time you need a tracer you retrieve it dynamically via`tracerProvider.get()`, which will delegate to `GlobalTracer.get()`. This way you leave a lot more opportunities to the application to init the global tracer before it stumbles on the need to push/pop spans into TraceContext

jaeger-context/src/main/java/com/uber/jaeger/context/TracingUtils.java
```diff
@@ -14,18 +14,28 @@
...
 
   public static ExecutorService tracedExecutor(ExecutorService wrappedExecutorService) {
+    assertGlobalTracerIsRegistered();
     return new TracedExecutorService(wrappedExecutorService, traceContext);
   }
 
-  private TracingUtils(){}
+  private static void assertGlobalTracerIsRegistered() {
+    if (!GlobalTracer.isRegistered()) {
+      throw new IllegalStateException("Please register a io.opentracing.util.GlobalTracer.");
```
vprithvi: @yurishkuro I felt it was simpler to use `GlobalTracer.get()` directly in `ActiveSpanSourceTraceContext`. It makes tests dirtier, but it's temporary code. 
What is your opinion on the `assertGlobalTracerIsRegistered` check?

yurishkuro: why not pass it the GlobalTracer.INSTANCE as we discussed? By using GlobalTracer directly you're introducing tight coupling in the new class, which didn't need to happen.

vprithvi: The type of `GlobalTracer.INSTANCE` is `GlobalTracer` and I didn't see that as being particularly advantageous because it still couples `ActiveSpanSourceTraceContext` to the implementation of `GlobalTracer`.

yurishkuro:
Instance implements there Tracer interface, which decouples the new span from GlobalTracer dependency

vprithvi:
>Instance implements there Tracer interface, which decouples the new span from GlobalTracer dependency
@yurishkuro  My understanding is that we were going to use the `GlobalTracer` instead of the `TracerProvider` that you had suggested earlier, and call a `getInstance` method so that the `INSTANCE` resolution happens as late as possible. The `Tracer` interface doesn't provide such a method.

vprithvi:
@yurishkuro Coveralls diffs against the latest successful build on `master`. It shows unrelated classes because the build on `master` failed.

yurishkuro:
master build is green now. I restarted this build.

PR #269, opened 2017-10-11
Use Mockito's timeout for test

### LINKED ISSUE #268: Fix flaky test
```
com.uber.jaeger.samplers.RemoteControlledSamplerTest > testUpdatePerOperationSamplerUpdatesExistingPerOperationSampler FAILED
    org.mockito.exceptions.verification.TooLittleActualInvocations: 
    perOperationSampler.update(
        Mock for OperationSamplingParameters, hashCode: 115078806
    );
    Wanted 2 times:
    -> at com.uber.jaeger.samplers.PerOperationSampler.update(PerOperationSampler.java:56)
    But was 1 time:
    -> at com.uber.jaeger.samplers.PerOperationSampler.update(PerOperationSampler.java:56)
        at com.uber.jaeger.samplers.PerOperationSampler.update(PerOperationSampler.java:56)
        at com.uber.jaeger.samplers.RemoteControlledSamplerTest.testUpdatePerOperationSamplerUpdatesExistingPerOperationSampler(RemoteControlledSamplerTest.java:113)
```

vprithvi:
- use Mockito's timeout instead of Thread.sleep
- Fixes #268

Signed-off-by: Prithvi Raj <p.r@uber.com>

### DIFF
--- a/jaeger-core/src/test/java/com/uber/jaeger/samplers/RemoteControlledSamplerTest.java
+++ b/jaeger-core/src/test/java/com/uber/jaeger/samplers/RemoteControlledSamplerTest.java
@@ -17,7 +17,7 @@ package com.uber.jaeger.samplers;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotEquals;
 import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.timeout;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
@@ -108,9 +108,8 @@ public class RemoteControlledSamplerTest {
     undertest = new RemoteControlledSampler(SERVICE_NAME, samplingManager, perOperationSampler, metrics);
 
     undertest.updateSampler();
-    Thread.sleep(20);
     //updateSampler is hit once automatically because of the pollTimer
-    verify(perOperationSampler, times(2)).update(parameters);
+    verify(perOperationSampler, timeout(2000).times(2)).update(parameters);
   }
 
   @Test


jaeger-core/src/test/java/com/uber/jaeger/samplers/RemoteControlledSamplerTest.java
```diff
@@ -108,9 +108,8 @@ public void testUpdatePerOperationSamplerUpdatesExistingPerOperationSampler() th
     undertest = new RemoteControlledSampler(SERVICE_NAME, samplingManager, perOperationSampler, metrics);
 
     undertest.updateSampler();
-    Thread.sleep(20);
     //updateSampler is hit once automatically because of the pollTimer
-    verify(perOperationSampler, times(2)).update(parameters);
+    verify(perOperationSampler, timeout(2000).times(2)).update(parameters);
```
black-adder: ELI5, how does this fix the flakiness?

vprithvi: I assumed that the `Thread.sleep` was being set to a value too low was the cause of the test failure and changed it to use mockito's timeout, which would poll until timeout to see if the expectation was met. 

That being said, I haven't yet verified that my assumption is correct, and simply wanted to trigger a build on travis

yurishkuro: if it works similar to https://github.com/awaitility/awaitility, then it's a sensible approach to deal with flaky tests - the alternative is to build complicated locking/semaphores, which is easy to get wrong. But given that we're simply protecting against delays in thread scheduling on VMs, and that the mean delay is negligible, it makes sense to use longer timeouts, like 10sec (under normal circumstances it will finish a lot sooner).

However, if the flakiness of the test is not due to scheduling but fundamentally due to a race condition in the test, then the timeout doesn't help.

jpkrohling:
Is this still relevant?

vprithvi:
I'm not really sure, I'll take a look into it in a bit

PR #270, opened 2017-10-12
Update errorprone and lombok and fix warnings and errors

vprithvi:
- Update errorprone to 2.1.1
    - Update lombok to 1.16.18
    - Disable the InstanceOfAndCastMatchWrongType and NestedInstanceOfConditions checks to ensure compatibility with lombok. See https://github.com/google/error-prone/issues/750

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@ plugins {
     id "com.github.hierynomus.license" version "0.14.0"
     id 'com.github.sherter.google-java-format' version '0.3.2'
     id "com.github.johnrengelman.shadow" version "2.0.1"
-    id "net.ltgt.errorprone" version "0.0.10"
+    id "net.ltgt.errorprone" version "0.0.12"
     id 'ru.vyarus.animalsniffer' version '1.3.0'
     id 'io.codearte.nexus-staging' version '0.11.0'
 }
@@ -167,6 +167,10 @@ configure(subprojects.findAll {it.name != 'jaeger-thrift'}) {
    dependencies {
      errorprone 'com.google.errorprone:error_prone_core:2.1.1'
    }
+    tasks.withType(JavaCompile) {
+        // These are disabled because of a bug with errorprone. See https://github.com/google/error-prone/issues/750
+        options.compilerArgs += [ '-Xep:NestedInstanceOfConditions:OFF',  '-Xep:InstanceOfAndCastMatchWrongType:OFF' ]
+    }
 }
 
 def getVersionForBuild() {
--- a/build.gradle
+++ b/build.gradle
@@ -126,7 +126,7 @@ subprojects {
     }
 
     dependencies {
-        compileOnly 'org.projectlombok:lombok:1.16.12'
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -165,8 +165,12 @@ task printVersion {
 configure(subprojects.findAll {it.name != 'jaeger-thrift'}) {
    apply plugin: 'net.ltgt.errorprone'
    dependencies {
-     errorprone 'com.google.errorprone:error_prone_core:2.0.15'
+     errorprone 'com.google.errorprone:error_prone_core:2.1.1'
    }
+    tasks.withType(JavaCompile) {
+        // These are disabled because of a bug with errorprone. See https://github.com/google/error-prone/issues/750
```
yurishkuro: not following how that issue is related to disabling checks

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -268,7 +268,7 @@ private SpanContext createNewContext(String debugId) {
 
       byte flags = 0;
       if (debugId != null) {
-        flags |= SpanContext.flagSampled | SpanContext.flagDebug;
+        flags = (byte) (flags | SpanContext.flagSampled | SpanContext.flagDebug);
```
yurishkuro: all operands are already of type `byte`, why do we need to re-cast?

build.gradle
vprithvi: Enabling those checks causes a runtime exception in the error prone compiler

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
vprithvi: Because of assignment conversion. See http://docs.oracle.com/javase/specs/jls/se7/html/jls-5.html#jls-5.2 and http://errorprone.info/bugpattern/NarrowingCompoundAssignment

The result of the bitwise OR is being promoted to a `int` which was previously implicitly cast to a `byte`.

build.gradle
yurishkuro: so using `@Data` is a convenience, but turning off InstanceOfAndCastMatchWrongType sounds like it could mask real errors.

vprithvi: `InstanceOfAndCastMatchWrongType` doesn't exist in the current version of errorprone that the project uses. 
Could you please justify your statement of `using @Data is a convenience`. Have there been no bugs made in manually writing out the convenience functions that `@Data/@Value` provides? If there have, is the cost of fixing the bugs introduced by incorrect hashcode/tostring/getters less than the potential time saved by having these new error checks enabled?
Has the time taken to write boilerplate code been factored into your argument?

yurishkuro: the methods that `@Data/@Value` generate can be just as easily created by the IDE.

vprithvi: I really feel that you are arguing with me for the sake for arguing. 
How would you maintain consistency between the IDE generated code?

Should bumping a couple of versions have so much friction?

PR #271, opened 2017-10-16
Add ability to supply custom span id

fzakaria:
This PR allows users of Jaeger to supply a custom span id.
In order to ease adoption for Jaeger it is beneficial to allow users to potentially supply their own id.

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
@@ -338,7 +338,7 @@ public class Tracer implements io.opentracing.Tracer, Closeable {
 
       return new SpanContext(
           preferredReference.getTraceId(),
-          Utils.uniqueId(),
+          this.id,
           preferredReference.getSpanId(),
           // should we do OR across passed references?
           preferredReference.getFlags(),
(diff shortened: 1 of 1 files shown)


vprithvi:
Could you elaborate on how this eases adoption?

fzakaria:
@vprithvi I work in a codebase that already has some notion of 'unique-id' generation and a non-distributed type of tracing.
It would make it easy to onboard to use the same id generation for jaeger as the current system.
(especially since OpenTracing doesn't expose span id)

yurishkuro:
Conceptually I am not against this patch, but would like to understand the implications and have the documented in the API. For example, how is sampling decision being made in this case?

Purely on the implementation side, I would like the interface to take two `long`s, for high and low 64bits of the ID, so that we don't need to change the API again in the future.

fzakaria:
@yurishkuro i like that that in the case I put forward, the `Sampler` is still in charge ? This merely sets the id to a predefined value.
(up to the user to provide whatever guarentee of randomness).

We could go more "Java EE" without it -- and make it an interface (or Supplier<Long>) but that seemed like overkill.

I'm open to what might need to be done.

yurishkuro:
> I work in a codebase that already has some notion of 'unique-id' generation and a non-distributed type of tracing.

Are you planning on converting the data from "non-distributed type of tracing" and storing it in Jaeger?  If not, what's the benefit of reusing the existing ID instead of just recording it as a tag?

PR #273, opened 2017-10-18
Splitting out the jaxrs2 filters

dray92:

### DIFF
--- a/jaeger-jaxrs2/src/main/java/com/uber/jaeger/filters/jaxrs2/ClientFilter.java
+++ b/jaeger-jaxrs2/src/main/java/com/uber/jaeger/filters/jaxrs2/ClientFilter.java
@@ -34,21 +34,21 @@ public class ClientFilter implements ClientRequestFilter, ClientResponseFilter {
   private final TraceContext traceContext;
 
   private final ClientSpanCreationFilter spanCreationFilter;
-  private final ClientSpanInjectionFilter spaninjectionFilter;
+  private final ClientSpanInjectionFilter spanInjectionFilter;
 
   public ClientFilter(Tracer tracer, TraceContext traceContext) {
     this.tracer = tracer;
     this.traceContext = traceContext;
 
     this.spanCreationFilter = new ClientSpanCreationFilter(tracer, traceContext);
-    this.spaninjectionFilter = new ClientSpanInjectionFilter(tracer, traceContext);
+    this.spanInjectionFilter = new ClientSpanInjectionFilter(tracer, traceContext);
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


jaeger-jaxrs2/src/main/java/com/uber/jaeger/filters/jaxrs2/ClientFilter.java
```diff
@@ -36,30 +33,22 @@
...
   public ClientFilter(Tracer tracer, TraceContext traceContext) {
     this.tracer = tracer;
     this.traceContext = traceContext;
+
+    this.spanCreationFilter = new ClientSpanCreationFilter(tracer, traceContext);
+    this.spaninjectionFilter = new ClientSpanInjectionFilter(tracer, traceContext);
   }
 
   @Override
   public void filter(ClientRequestContext clientRequestContext) throws IOException {
```
black-adder: Just so I follow, for your use case, you'll extend the ClientFilter and then override this filter function to add your baggage between the spanCreationFilter and spanInjectionFilter?

What do we do in the future when other people want to add extra functionality? Will they add their functionality into your filter?

yurishkuro: then someone can add a CompositeFilter

black-adder: right, so is that the plan?

yurishkuro: there's no plan - it's not a problem today, and it's solvable in the future with composite filter, so we don't need to over-engineer today for non-existing need

black-adder:
every commit has to be signed off, you can squash your commits under one with the sign off

jaeger-jaxrs2/src/main/java/com/uber/jaeger/filters/jaxrs2/ClientSpanCreationFilter.java
```diff
@@ -0,0 +1,57 @@
...
+import javax.ws.rs.RuntimeType;
+import javax.ws.rs.client.ClientRequestContext;
+import javax.ws.rs.client.ClientRequestFilter;
+import javax.ws.rs.ext.Provider;
+import lombok.extern.slf4j.Slf4j;
+
+@Provider
+@ConstrainedTo(RuntimeType.CLIENT)
+@Slf4j
+public class ClientSpanCreationFilter implements ClientRequestFilter {
```
willsalz: I don't think that we should be releasing OSS without proper Javadocs.

PR #286, opened 2017-11-09
Updates of runtime components

alshopov:
This is up for discussion, we could do it partially


yurishkuro:
what is the motivation for these updates?  Some I don't have problems with, like mockito or brave/zipkin, as they are implementation details. Others are a bit more problematic, e.g. tchannel upgrade in crossdock needs testing (if works, then ok), or Jersey upgrades that appears to be actually breaking (since you had to change code).

alshopov:
Bugfixes and performance in the underlying libs is the main reason. As jaeger is used throughout Uber's services depending on old version causes problems.
I have removed the jersey update.
The tchannel 0.7.7 is far better than 0.5 as many memory leaks have been eliminated both in it and underlying netty.
What testing do you expect for crossdock? instantpay uses latest tchannel-java and xtchannel-java with much less problems than previous versions.

yurishkuro:
> As jaeger is used throughout Uber's services depending on old version causes problems.

that's the point I don't understand - unless there are API differences, the versions that Jaeger lib is using should be non-binding to the actual applications; if they want to use newer versions the build system should resolve accordingly.

alshopov:
Moved to #290

alshopov:
I get your point as a maintainer - you would rather ensure long term stability and minimize changes.
However you have to weigh this against other goals.
I do point releases updates because they usually contain bugfixes, especially security ones. I've had instances when due to an outdated library we have relied on a bug as a feature. These things get ironed out with the updates.
Larger than point releases updates are to minimize the usage of old, deprecated APIs. Unless you do these regularly, they keep piling up and it becomes increasingly more difficult to finally fix them.
While what you say is true - a project may declare updated dependencies, it also may not and that also causes maintenance burden. Additionally it becomes harder to convince a project to move to a new version if the library it depends on has not been verified to work with that version.
But yeah - these decisions rest finally on the maintainers.
I may reduce the number of updates or you can deny the pull request.

yurishkuro:
Given the current list of upgrades I don't have objections to them. But it looks like they are already introducing incompatibility as the build is broken with NoSuchMethod

alshopov:
@yurishkuro - is the patch fine now?

yurishkuro:
looks like we have flaky test coverage, [this](https://codecov.io/gh/jaegertracing/jaeger-client-java/compare/916de0c6193cc194a9a311928e72bbd489417c5b...d1b8e559ae84a4f8e1c50b2fb3018286f1d610f4/src/jaeger-core/src/main/java/com/uber/jaeger/reporters/protocols/ThriftUdpTransport.java) didn't execute
```
if (!this.isOpen()) {
 throw new TTransportException(TTransportException.NOT_OPEN);
```

PR #290, opened 2017-11-15
Iterate on Maps via Entry rather than key + retrieval of value

alshopov:
Signed-off-by: Alexander Shopov <ashopov@uber.com>


alshopov:
This is #286 but stemming outside master

yurishkuro:
thanks!

PR #291, opened 2017-11-15
Gonfiguration.getTracer returns the GlobalTracer if registered.

boiantz:
Signed-off-by: Boian Tzonev <boian@uber.com>

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
@@ -184,6 +184,10 @@ public class Configuration {
       return tracer;
     }
 
+    if (GlobalTracer.isRegistered()) {
+      return GlobalTracer.get();
+    }
+
     tracer = getTracerBuilder().build();
     log.info("Initialized tracer={}", tracer);
 
--- a/jaeger-core/src/test/java/com/uber/jaeger/ConfigurationTest.java
+++ b/jaeger-core/src/test/java/com/uber/jaeger/ConfigurationTest.java
@@ -66,13 +66,14 @@ public class ConfigurationTest {
     assertTrue(GlobalTracer.isRegistered());
   }
 
-  @Test (expected = IllegalStateException.class)
+  @Test
   public void testFromEnvWithoutDisabledTracer() {
     System.setProperty(Configuration.JAEGER_SERVICE_NAME, "Test");
     assertNotNull(Configuration.fromEnv().getTracer());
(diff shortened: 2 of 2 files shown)


boiantz:
There a situations when we don't have control when a tracer is intialized for the first time - for example thcannel's initialisation uses getTracer - if my application code registers the GlobalTracer before tchannel is initialised, tchannel initialisation will fail. In this case even protecting my code in a try catch clause will not help.

Currently the intent of this class is to return a 'singleton' instance of the tracer but that cannot be guaranteed since the Configuration class is not static itself. As such I think this change fulfils intent better.

Please share any recommendations how to solve this without this change if you have any.

boiantz:
I think I found the way to init tchannel with issues as well.

boiantz:
Actually I think this is still relevant. I think that the fact that this library is relying on a GlobalTracer is an internal detail that should not leak outside to library consumers.

vprithvi:
>I think that the fact that this library is relying on a GlobalTracer is an internal detail that should not leak outside to library consumers.

Registering with the GlobalTracer is optional, and may be disabled by setting the `DISABLE_DISABLE_GLOBAL_TRACER` env variable.  

We chose to registering with the global tracer as the default so that any other code relying on the `GlobalTracer` would work without additional configuration. Otherwise, using an unregistered `GlobalTracer` would simply default to using a `NoopTracer`, which is more difficult to detect and fix.  

Another reason why we did this was so that we can move to using an Opentracing context propagation mechanism instead of jaeger's own implementation.

yurishkuro:
@vprithvi did we document the "optional" part? 

> Another reason why we did this was so that we can move to using an Opentracing context propagation mechanism instead of jaeger's own implementation.

That was the primary reason. We previously provided a static class that allowed access to the current span through a custom thread-local mechanism. Since OT added support for active span, we switched to it from the custom thread-local, but to keep the old static class backwards compatible we needed access to the global tracer and its active span source (although I don't remember why we couldn't just pass our tracer to the static class instead - I think it could've been the circular module dependency issue).

boiantz:
> We chose to registering with the global tracer as the default so that any other code relying on the GlobalTracer would work without additional configuration. Otherwise, using an unregistered GlobalTracer would simply default to using a NoopTracer, which is more difficult to detect and fix.

Yes, I tried using DISABLE_GLOBAL_TRACER but there were some Utils that started failing at some places so I believe that it is not really a viable option. Either way I don't think there is any harm in the diff that I am proposing but as I said it is not a deal breaker anymore so I am closing this pull request.

Thanks!

PR #293, opened 2017-11-18
Add new module for JSON representation of Spans

fzakaria:
I am incorporating Jaeger and there is already a component which 
given a file will upload it to an ElasticSearch index.

I am now working on making available through the java-client the same
JSON respresentation that the collector serializes before sending
upstream to the datastore.

You could imagine also other very simple use cases where an AsyncReporter
could also send directly to the datastore.

I'm still thinking about how to make tests since the spanId is randomized.
Please feel free to close this PR if there is the belief there is no use for it -- I just thought I would propose it to see if anyone else finds it valuable.

Here is a sample of the JSON representation:

```json
{
  "duration" : 5,
  "tags" : [ {
    "key" : "peer.service",
    "value" : "service-y",
    "type" : "string"
  }, {
    "key" : "temperature",
    "value" : 72.5,
    "type" : "float64"
  }, {
    "key" : "error",
    "value" : true,
    "type" : "bool"
  }, {
    "key" : "peer.ipv4",
    "value" : 23456,
    "type" : "int64"
  } ],
  "traceID" : "1",
  "spanID" : "-1805624008102187457",
  "parentSpanID" : "2",
  "process" : {
    "serviceName" : "service-x",
    "tags" : [ {
      "key" : "hostname",
      "value" : "fmzakari-mac",
      "type" : "string"
    }, {
      "key" : "jaeger.version",
      "value" : "Java-0.22.0-RC1",
      "type" : "string"
    }, {
      "key" : "ip",
      "value" : "127.0.0.1",
      "type" : "string"
    } ]
  },
  "logs" : [ ],
  "references" : [ {
    "traceID" : "1",
    "spanID" : "2",
    "refType" : "follows_from"
  }, {
    "traceID" : "1",
    "spanID" : "2",
    "refType" : "child_of"
  } ],
  "operationName" : "test-general-conversion",
  "flags" : 1,
  "startTime" : 1485467191639875
}
```

Signed-off-by: Farid Zakaria <farid.m.zakaria@gmail.com>


jpkrohling:
I like the general idea of this PR, but looks like this includes only the data structures. Do you plan on working further on this PR, to integrate with the other modules? If so, please open a thread on the mailing list to discuss it before spending too much time on it :)

> You could imagine also other very simple use cases where an AsyncReporter
could also send directly to the datastore.

Am I right to assume that the idea is to skip the agent/collector?

pavolloffay:
@fzakaria thanks for the PR. I will close it as it hasn't been active for a long time.

PR #297, opened 2017-11-29
[apache] [jax-rs2] Instrumentation specific multiple span reporting Issue

dray92:
- Apache span interceptors have lighter dependency on threadlocal; newly created span is not pushed
- HttpContext is used to propagate the span between interceptors
- Similar changes in jax-rs2 instrumentation to reduce dependence on thread local, and instead use jax-rs2 specific `javax.ws.rs.client.ClientRequestContext`

### DIFF
--- a/jaeger-apachehttpclient/src/test/java/com/uber/jaeger/httpclient/JaegerRequestAndResponseInterceptorIntegrationTest.java
+++ b/jaeger-apachehttpclient/src/test/java/com/uber/jaeger/httpclient/JaegerRequestAndResponseInterceptorIntegrationTest.java
@@ -88,6 +88,9 @@ public class JaegerRequestAndResponseInterceptorIntegrationTest {
 
     client.start();
 
+    // Verify that parent span is on top of the stack _before_ request is made
+    assertEquals(parentSpan, TracingUtils.getTraceContext().getCurrentSpan());
+
     //Make a request to the async client and wait for response
     client
         .execute(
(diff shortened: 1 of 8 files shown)


jaeger-core/src/main/java/com/uber/jaeger/Span.java
```diff
@@ -147,6 +148,10 @@ public SpanContext context() {
 
   @Override
   public void finish() {
+    if (finished) {
```
yurishkuro: this logic should be in the `finishWithDuration` method, and inside the synchronized block

vprithvi:
Could you split this into two separate PRs by separating the instrumentation fixes from the changes in the Span class?

dray92:
@vprithvi definitely -- lemme just remove the span class changes

jaeger-apachehttpclient/src/test/java/com/uber/jaeger/httpclient/JaegerRequestAndResponseInterceptorIntegrationTest.java
```diff
@@ -124,7 +130,7 @@ public void testHttpClientTracing() throws Exception {
   private void verifyTracing(Span parentSpan) {
     //Assert that traces are correctly emitted by the client
     List<Span> spans = reporter.getSpans();
-    assertEquals(3, spans.size());
+    assertEquals(2, spans.size());
```
yurishkuro: what is the reason for this change?

dray92: Previously, when `pop()` was getting called, it would make another call to report the span. Hence the reporter was getting an extra span, and the last 2 spans were identical. This changes proves that the core issue is fixed.

PR #299, opened 2017-12-01
[jaeger-client-java] Span class does not report multiple time on multiple invocations of finish()

dray92:


jaeger-core/src/main/java/com/uber/jaeger/Span.java
```diff
@@ -162,6 +163,11 @@ public void finish(long finishMicros) {
 
   private void finishWithDuration(long durationMicros) {
     synchronized (this) {
+      if (finished) {
+        return;
```
yurishkuro: would it make sense to log a warning?

dray92: Yep - I think it does. It seems important to know that the subsequent calls to `finish()`, will have no real effect. Making the change.

PR #300, opened 2017-12-01
[jaeger-client-java] Reduce eagerness of Configuration.getTracer to register a global tracer

dray92:
With this change, we leave it up to the consumer to be the decider for registering a `GlobalTracer`. As much as I understand the reasoning for why it was put there in the first place, I think it's causing a bunch of confusion. 

There is the flag present if people want the call to `getTracer()` to go ahead and register the current tracer as the `GlobalTracer`. 

What do you all think about this?

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
@@ -126,7 +126,7 @@ public class Configuration {
   public static final String JAEGER_TAGS = JAEGER_PREFIX + "TAGS";
 
   /**
-   * Disables registration with {@link GlobalTracer}.
+   * Enables registration with {@link GlobalTracer}.
    */
   public static final String JAEGER_ENABLE_GLOBAL_TRACER = JAEGER_PREFIX + "ENABLE_GLOBAL_TRACER";
 
@@ -145,8 +145,7 @@ public class Configuration {
   private StatsFactory statsFactory;
 
   /**
-   * Don't use {@link GlobalTracer} to store the tracer.
-   * Use the local {@link #tracer} instead.
+   * Use {@link GlobalTracer} to store the tracer.
    */
   private final boolean enableGlobalTracer;
 
--- a/jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
@@ -128,7 +128,7 @@ public class Configuration {
   /**
    * Disables registration with {@link GlobalTracer}.
    */
-  public static final String JAEGER_DISABLE_GLOBAL_TRACER = JAEGER_PREFIX + "DISABLE_GLOBAL_TRACER";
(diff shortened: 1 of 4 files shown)


jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
```diff
@@ -189,15 +188,15 @@ private Configuration(
...
+    this.enableGlobalTracer = enableGlobalTracer;
   }
 
   public static Configuration fromEnv() {
     return new Configuration(
         getProperty(JAEGER_SERVICE_NAME),
         SamplerConfiguration.fromEnv(),
         ReporterConfiguration.fromEnv(),
-        getPropertyAsBool(JAEGER_DISABLE_GLOBAL_TRACER));
+        getPropertyAsBool(JAEGER_ENABLE_GLOBAL_TRACER));
```
isaachier: Why are we reversing the defaults?

debosmit: We'd rather have this not register the local tracer as the global tracer, so eagerly. If someone wishes to have this feature, they should be made aware of it, and they can set the config appropriately.

yurishkuro:
This will be a breaking change, because apps that are trying to access `TraceUtil.getCurrentSpan()` will start failing.

I agree that we may want to remove the dependency on the GlobalTracer. I would actually remove it completely. But we need to figure out what to do about TraceUtil.

debosmit:
@yurishkuro Yea, breaking change definitely. [Elaborating on reason, for readers of thread] Now people have to take up the work of registering the current tracer, as the global tracer, and then ensure that it is set for `TracingUtils` as well.

Yea, I totally agree. But, might be good to actually have a solid way to have access to current span. IIRC, in the present world, there's a bit of reflection going on when you try to obtain the active span from a jaeger tracer - since this is something that will be called in hot paths, we might want to figure a way around that.

yurishkuro:
Is there a way for our internal libs (like xtchannel) to register the tracer with TracingUtils?  If we can do that, then I would suggest the following plan:
  * change TracingUtils to require a call to `TracingUtils.setTracer()` before any other methods can be used. The exception should mention that `TracingUtils is going to be deprecated. To use its method please provide a tracer via TracingUtils.setTracer()`.
  * remove references to GlobalTracer from Configuration completely (no env vars or anything) - if we don't have internal dependencies on it then there is no reason to provide a way of registering it via config options, users can do it themselves.

debosmit:
@ys This definitely seems like a decent idea. Do you think it should be a part of this PR as well?

yurishkuro:
I think it should be _instead_ of this PR, because this PR introduced a new env var which I am suggesting to eliminate. But we need to validate the assumption "Is there a way for our internal libs (like xtchannel) to register the tracer with TracingUtils?"

debosmit:
From what I see, we can easily add it to both core-dw and xtchannel. We simply get the tracer from jaeger config, and then register it by calling `TracingUtils.setTracer(io.opentracing.Tracer)`.

yurishkuro:
ok let's do it then. We should do it while still on RC

dray92:
Synced offline with @yurishkuro 
Gonna be covered by #302

PR #302, opened 2017-12-07
Remove concept of registering a GlobalTracer when getting a new tracer

dray92:
TracingUtils classes have been deprecated. There is a new method called `setTracer(io.opentracing.Tracer)` that can be used to continue using the methods in that class. 

When trying to get a tracer from `jaeger.core.Configuration` object, it was previously registering that tracer as a global tracer. That functionality and the related environment variables have been removed. 

Users will have to be responsible a local tracer as a Global Tracer.

### DIFF
--- a/jaeger-apachehttpclient/src/test/java/com/uber/jaeger/httpclient/JaegerRequestAndResponseInterceptorIntegrationTest.java
+++ b/jaeger-apachehttpclient/src/test/java/com/uber/jaeger/httpclient/JaegerRequestAndResponseInterceptorIntegrationTest.java
@@ -22,8 +22,8 @@ import com.uber.jaeger.context.TracingUtils;
 import com.uber.jaeger.reporters.InMemoryReporter;
 import com.uber.jaeger.samplers.ConstSampler;
 import com.uber.jaeger.samplers.Sampler;
-import com.uber.jaeger.utils.TestUtils;
-import io.opentracing.util.GlobalTracer;
+import io.opentracing.NoopTracerFactory;
+import java.lang.reflect.Field;
 import java.util.List;
 import org.apache.http.HttpHost;
 import org.apache.http.concurrent.FutureCallback;
@@ -67,7 +67,7 @@ public class JaegerRequestAndResponseInterceptorIntegrationTest {
     reporter = new InMemoryReporter();
     Sampler sampler = new ConstSampler(true);
     tracer = new Tracer.Builder("test_service", reporter, sampler).build();
-    GlobalTracer.register(tracer);
+    TracingUtils.setTracer(tracer);
 
     parentSpan = (Span) tracer.buildSpan("parent_operation").startManual();
     parentSpan.setBaggageItem(BAGGAGE_KEY, BAGGAGE_VALUE);
@@ -78,7 +78,9 @@ public class JaegerRequestAndResponseInterceptorIntegrationTest {
 
   @After
   public void tearDown() throws Exception {
-    TestUtils.resetGlobalTracer();
+    Field field = TracingUtils.class.getDeclaredField("tracer");
+    field.setAccessible(true);
+    field.set(null, NoopTracerFactory.create());
   }
 
   @Test
--- a/jaeger-apachehttpclient/src/test/java/com/uber/jaeger/httpclient/JaegerRequestAndResponseInterceptorIntegrationTest.java
+++ b/jaeger-apachehttpclient/src/test/java/com/uber/jaeger/httpclient/JaegerRequestAndResponseInterceptorIntegrationTest.java
@@ -22,7 +22,6 @@ import com.uber.jaeger.context.TracingUtils;
 import com.uber.jaeger.reporters.InMemoryReporter;
 import com.uber.jaeger.samplers.ConstSampler;
 import com.uber.jaeger.samplers.Sampler;
-import io.opentracing.NoopTracerFactory;
 import java.lang.reflect.Field;
 import java.util.List;
 import org.apache.http.HttpHost;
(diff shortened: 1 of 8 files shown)


jaeger-context/src/main/java/com/uber/jaeger/context/TracingUtils.java
```diff
@@ -18,9 +18,18 @@
...
+/**
+ * TracingUtils is going to be deprecated. To use its method please provide a tracer via
+ * {@link com.uber.jaeger.context.TracingUtils#setTracer(io.opentracing.Tracer)}.
+ */
+@Deprecated
 public class TracingUtils {
   private static TraceContext traceContext = new ActiveSpanSourceTraceContext(GlobalTracer.get());
 
+  public static void setTracer(io.opentracing.Tracer tracer) {
+    GlobalTracer.register(tracer);
```
yurishkuro: This is not quite what I meant. Rather than depending on global tracer AT ALL, this could simply store the tracer and init traceContext with it. The other static methods check for traceContext to be initialized, if it's not they throw a descriptive error saying that `setTracer` must be called before using this class.

debosmit: Gotcha!

jaeger-context/src/main/java/com/uber/jaeger/context/TracingUtils.java
```diff
@@ -14,26 +14,37 @@
...
 import java.util.concurrent.ExecutorService;
 
+/**
+ * TracingUtils is going to be deprecated. To use its method please provide a tracer via
+ * {@link com.uber.jaeger.context.TracingUtils#setTracer(io.opentracing.Tracer)}.
+ */
+@Deprecated
 public class TracingUtils {
-  private static TraceContext traceContext = new ActiveSpanSourceTraceContext(GlobalTracer.get());
+  private static io.opentracing.Tracer tracer = NoopTracerFactory.create();
```
yurishkuro: I think `null` would do just as well here, even safer in case one of the business methods forgets to call assertTracerRegistered().

jaeger-context/src/main/java/com/uber/jaeger/context/TracingUtils.java
```diff
@@ -14,26 +14,37 @@
...
+  private static TraceContext traceContext;
+
+  public static void setTracer(io.opentracing.Tracer tracer) {
+    TracingUtils.tracer = tracer;
```
yurishkuro: maybe I am old school, but I was never comfortable with an assumption that assigning / accessing a reference is an atomic operation. I suggest adding `synchronized` to all public methods in this class.

jaeger-context/src/main/java/com/uber/jaeger/context/TracingUtils.java
dray92: +1

jaeger-context/src/main/java/com/uber/jaeger/context/TracingUtils.java
dray92: that's a very good point - and seems to be more solid design-wise 👍

dray92: i think this applies to `com.uber.jaeger.filters.jaxrs2.TracingUtils` as well, apart from `com.uber.jaeger.context.TracingUtils`

vprithvi: It is an assumption that the JVM adheres to the Java Language Spec.
https://docs.oracle.com/javase/specs/jls/se7/html/jls-17.html#jls-17.7

Could you elaborate on why one should be skeptical of a fundamental item in the language spec?

jaeger-apachehttpclient/src/test/java/com/uber/jaeger/httpclient/JaegerRequestAndResponseInterceptorIntegrationTest.java
```diff
@@ -78,7 +77,9 @@ public void setUp() {
 
   @After
   public void tearDown() throws Exception {
-    TestUtils.resetGlobalTracer();
+    Field field = TracingUtils.class.getDeclaredField("tracer");
+    field.setAccessible(true);
+    field.set(null, null);
```
vprithvi: This clean up is repeated in a few places, could you make move it to `TestUtils` and reuse?

jaeger-jaxrs2/src/main/java/com/uber/jaeger/filters/jaxrs2/TracingUtils.java
```diff
@@ -18,22 +18,32 @@
...
+@Deprecated
 public class TracingUtils {
+
+  public static synchronized void setTracer(io.opentracing.Tracer tracer) {
+    com.uber.jaeger.context.TracingUtils.setTracer(tracer);
+  }
+
   @Deprecated
-  public static TraceContext getTraceContext() {
+  public static synchronized TraceContext getTraceContext() {
```
vprithvi: These methods don't need to by synchronized because the underlying `context.TracingUtils` class is thread safe.

jaeger-apachehttpclient/src/test/java/com/uber/jaeger/httpclient/JaegerRequestAndResponseInterceptorIntegrationTest.java
dray92: Actually becoming a bit of a dep mess when I do that since `TestUtils` is a part of the core module

yurishkuro:
```
* What went wrong:
Execution failed for task ':jaeger-context:checkstyleMain'.
> Checkstyle rule violations were found. See the report at: file:///home/travis/build/jaegertracing/jaeger-client-java/jaeger-context/build/reports/checkstyle/main.html
```

dray92:
whoops! thanks @yurishkuro

dray92:
@yurishkuro can you drop me a small note when this is landed and a release is cut? I will need to bump the deps for xtchannel and a couple other libs. will be traveling tomorrow evening - will update you when I have internet connectivity.

PR #307, opened 2017-12-11
Fix currentSpan() when active span is null

### LINKED ISSUE #278: NPE when getCurrentSpan() called on empty trace context
Call to `TracingUtils.getTraceContext().getCurrentSpan()` when span stack is empty results in a NPE.

Method contract in [`TracingUtils.getCurrentSpan()`](https://github.com/jaegertracing/jaeger-client-java/blob/master/jaeger-context/src/main/java/com/uber/jaeger/context/TraceContext.java#L44) or [`ActiveSpanSourceTraceContext .getCurrentSpan()`](https://github.com/jaegertracing/jaeger-client-java/blob/master/jaeger-context/src/main/java/com/uber/jaeger/context/ActiveSpanSourceTraceContext.java#L71) does not specify this.

yurishkuro:
Resolves #278


vprithvi:
Could you also add javadocs?

PR #310, opened 2017-12-13
Make apache context span key public

dray92:
- apache http context key for span was not public
- using interfaces to store public constants to allow for easier static imports for both apache and jaxrs2


debosmit:
I think build needs to be restarted on this one

PR #314, opened 2017-12-28
Proof-of-concept: Jaeger configuration returns Jaeger Tracer

dray92:
Mostly c/p from opentracing.noop to show how something like this would work.

Ideally, just changing return type of `com.uber.jaeger.Configuration#getTracer` would do the trick. But, since `com.uber.jaeger.dropwizard.Configuration` sub-classes `com.uber.jaeger.Configuration` and also supports `disable` state, that returns an `NoopTracer` (which is an `io.opentracing.Tracer`), we can't change that return-type to `com.uber.jaeger.Tracer`.

### DIFF
--- a/jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/Configuration.java
+++ b/jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/Configuration.java
@@ -18,8 +18,11 @@ import com.codahale.metrics.MetricRegistry;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
-import io.opentracing.NoopTracerFactory;
-import io.opentracing.Tracer;
+import com.uber.jaeger.metrics.Metrics;
+import com.uber.jaeger.reporters.NoopReporter;
+import com.uber.jaeger.reporters.Reporter;
+import com.uber.jaeger.samplers.ConstSampler;
+import com.uber.jaeger.samplers.Sampler;
 
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class Configuration extends com.uber.jaeger.Configuration {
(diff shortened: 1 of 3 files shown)


yurishkuro:
Rather than faking noop impl, perhaps we should reconsider returning it in the first place. The whole "disable tracing" feature request was a knee-jerk reaction when we were first introducing tracing, in practice it has never been needed _for performance reasons_. And if we are not talking about performance reasons, then approximating noop-tracer with a real Jaeger tracer is pretty easy - just use ConstSampler(false) and NullReporter.

And most importantly, many new features are being built that require context propagation even when not tracing, so using the real tracer makes a lot more sense to me.

dray92:
@yurishkuro Ah! I we aren't tied to it, then its awesome.

`disabled` can still stay as it is, and simply return a Tracer with `ConstSampler(false)` and `NullReporter`, instead of the opentracing no-op tracer.
And as you said, the newer features that you mention, can go in without much resistance.

Let me know if my understanding is correct, and I can update to remove the fake no-op stuff, and simply return a "noop-configured" jaeger tracer.

yurishkuro:
go for it

jaeger-dropwizard/src/main/java/com/uber/jaeger/dropwizard/Configuration.java
```diff
@@ -39,9 +41,15 @@ public Configuration(
...
 
   @Override
-  public synchronized Tracer getTracer() {
+  public synchronized com.uber.jaeger.Tracer getTracer() {
     if (disable) {
-      return NoopTracerFactory.create();
+      Metrics metrics = new Metrics(statsFactory);
+      ConstSampler sampler = new ConstSampler(false);
+      Reporter reporter = new NoopReporter();
+      return new com.uber.jaeger.Tracer.Builder(serviceName, reporter, sampler)
```
dray92: @yurishkuro should we store a reference to this tracer (maybe call it no-op tracer), so that a new one does not have to be built on every call?

yurishkuro: yes, I think it should be stored the same way as if it was the regular tracer.

yurishkuro:
@dray92 we decided not to do this, correct?

dray92:
@yurishkuro Yup -- unnecessarily complex

PR #315, opened 2018-01-07
Allow passing debug span baggage from non-jaegerized clients

dray92:
Related to https://github.com/jaegertracing/jaeger/issues/165

Previously, presence of debug ID would not even consider baggage items populated. After I changed that behavior, I realized that the bare presence of the debug ID would cause child span to not get the baggage from list of references. Fixing both of these allowed `uberctx-*` to get populated as span baggage **iff** either ``uber-trace-id`` or ``jaeger-debug-id`` is present.
_**Awesome!**_ 🍺

Now, to maintain parity with the Go client, a new header called ``jaeger-baggage`` is being introduced. Communication to people can be that when using curl, etc to test, include both ``jaeger-debug-id`` and ``jaeger-baggage``. From my reading of the Go code, this seems to be the case (but I don't understand Go much, so I might be wrong).

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/Constants.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Constants.java
@@ -39,8 +39,15 @@ public class Constants {
    * An external process can inject baggage by setting the special HTTP Header jaeger-baggage
    * on a request. This is expected to be used by external processes to inject span baggage into
    * span context in cases where it is not created by a Jaeger-instrumented client/server.
+   *
+   * Debug jaeger baggage is loaded for a header item that looks like,
+   * <pre>
+   * {@code
+   *    "jaeger-baggage": "k1=v1,k2=v2"
+   * }
+   * </pre>
    */
-  public static final String DEBUG_JAEGER_SPAN_BAGGAGE_HEADER = "jaeger-baggage";
+  public static final String MANUAL_BAGGAGE_HEADER_KEY = "jaeger-baggage";
 
   /**
    * The name of the tag used to report client version.
--- a/jaeger-core/src/main/java/com/uber/jaeger/Constants.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Constants.java
@@ -36,6 +36,13 @@ public class Constants {
   public static final String DEBUG_ID_HEADER_KEY = "jaeger-debug-id";
 
   /**
+   * An external process can inject baggage by setting the special HTTP Header jaeger-baggage
+   * on a request. This is expected to be used by external processes to inject span baggage into
+   * span context in cases where it is not created by a Jaeger-instrumented client/server.
+   */
+  public static final String DEBUG_JAEGER_SPAN_BAGGAGE_HEADER = "jaeger-baggage";
+
+  /**
    * The name of the tag used to report client version.
    */
   public static final String JAEGER_CLIENT_VERSION_TAG_KEY = "jaeger.version";
--- a/jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
@@ -283,11 +283,11 @@ public class Tracer implements io.opentracing.Tracer, Closeable {
         }
       }
 
-      return new SpanContext(id, id, 0, flags);
+      return new SpanContext(id, id, 0, flags, createChildBaggage(), debugId);
     }
 
     private Map<String, String> createChildBaggage() {
-      Map<String, String> baggage = null;
+      Map<String, String> baggage = new HashMap();
 
       // optimization for 99% use cases, when there is only one parent
       if (references.size() == 1) {
@@ -296,9 +296,6 @@ public class Tracer implements io.opentracing.Tracer, Closeable {
 
       for (Reference reference: references) {
         if (reference.getSpanContext().baggage() != null) {
-          if (baggage == null) {
-            baggage = new HashMap<String, String>();
-          }
           baggage.putAll(reference.getSpanContext().baggage());
         }
       }
--- a/jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
@@ -31,14 +31,12 @@ import com.uber.jaeger.samplers.SamplingStatus;
 import com.uber.jaeger.utils.Clock;
 import com.uber.jaeger.utils.SystemClock;
 import com.uber.jaeger.utils.Utils;
-
 import io.opentracing.References;
 import io.opentracing.Scope;
 import io.opentracing.ScopeManager;
 import io.opentracing.propagation.Format;
 import io.opentracing.tag.Tags;
 import io.opentracing.util.ThreadLocalScopeManager;
-
 import java.io.Closeable;
 import java.io.InputStream;
 import java.net.Inet4Address;
@@ -287,13 +285,12 @@ public class Tracer implements io.opentracing.Tracer, Closeable {
     }
 
     private Map<String, String> createChildBaggage() {
-      Map<String, String> baggage = new HashMap();
-
       // optimization for 99% use cases, when there is only one parent
       if (references.size() == 1) {
         return references.get(0).getSpanContext().baggage();
       }
 
+      Map<String, String> baggage = new HashMap<String, String>();
       for (Reference reference: references) {
         if (reference.getSpanContext().baggage() != null) {
           baggage.putAll(reference.getSpanContext().baggage());
--- a/jaeger-core/src/main/java/com/uber/jaeger/propagation/TextMapCodec.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/propagation/TextMapCodec.java
@@ -14,8 +14,9 @@
 
 package com.uber.jaeger.propagation;
 
-import static com.uber.jaeger.Constants.DEBUG_JAEGER_SPAN_BAGGAGE_HEADER;
+import static com.uber.jaeger.Constants.MANUAL_BAGGAGE_HEADER_KEY;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.uber.jaeger.Constants;
 import com.uber.jaeger.SpanContext;
 import io.opentracing.propagation.TextMap;
@@ -25,6 +26,9 @@ import java.net.URLEncoder;
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
+import java.util.regex.Pattern;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 public class TextMapCodec implements Codec<TextMap> {
   /**
@@ -45,6 +49,10 @@ public class TextMapCodec implements Codec<TextMap> {
 
   private final boolean urlEncoding;
 
+  private static final Pattern commaPattern = Pattern.compile(",");
+
+  private static final Pattern equalsPattern = Pattern.compile("=");
+
   public TextMapCodec(boolean urlEncoding) {
     this(builder().withUrlEncoding(urlEncoding));
   }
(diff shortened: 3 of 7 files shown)


jaeger-core/src/main/java/com/uber/jaeger/propagation/TextMapCodec.java
```diff
@@ -64,7 +66,7 @@ public void inject(SpanContext spanContext, TextMap carrier) {
   @Override
   public SpanContext extract(TextMap carrier) {
     SpanContext context = null;
-    Map<String, String> baggage = null;
+    Map<String, String> baggage = new HashMap();
```
yurishkuro: this causes unnecessary allocation when there is no baggage

jaeger-core/src/main/java/com/uber/jaeger/propagation/TextMapCodec.java
```diff
@@ -74,20 +76,30 @@ public SpanContext extract(TextMap carrier) {
...
+         */
+        String decodedDebugBaggageItems = decodedValue(entry.getValue());
+        String[] debugBaggageItems = decodedDebugBaggageItems.split(",");
+        for (String debugBaggageItem: debugBaggageItems) {
+          String[] debugBaggageItemParts = debugBaggageItem.split("=");
+          baggage.put(debugBaggageItemParts[0], debugBaggageItemParts[1]);
+        }
       }
     }
+    // this pattern allows for presence of debug ID to trump everything else
```
yurishkuro: comment seems inaccurate - if the context was created because there was incoming trace, then debug id is ignored

jaeger-core/src/main/java/com/uber/jaeger/propagation/TextMapCodec.java
```diff
@@ -74,20 +76,30 @@ public SpanContext extract(TextMap carrier) {
...
+         * {@code
+         *    "jaeger-baggage": "k1=v1,k2=v2"
+         * }
+         * </pre>
+         */
+        String decodedDebugBaggageItems = decodedValue(entry.getValue());
+        String[] debugBaggageItems = decodedDebugBaggageItems.split(",");
+        for (String debugBaggageItem: debugBaggageItems) {
+          String[] debugBaggageItemParts = debugBaggageItem.split("=");
+          baggage.put(debugBaggageItemParts[0], debugBaggageItemParts[1]);
```
yurishkuro: must check for malformed headers, current code can cause IndexOutOfBounds exception

jaeger-core/src/main/java/com/uber/jaeger/Constants.java
```diff
@@ -35,6 +35,13 @@
    */
   public static final String DEBUG_ID_HEADER_KEY = "jaeger-debug-id";
 
+  /**
+   * An external process can inject baggage by setting the special HTTP Header jaeger-baggage
+   * on a request. This is expected to be used by external processes to inject span baggage into
+   * span context in cases where it is not created by a Jaeger-instrumented client/server.
+   */
+  public static final String DEBUG_JAEGER_SPAN_BAGGAGE_HEADER = "jaeger-baggage";
```
yurishkuro: MANUAL_BAGGAGE_HEADER_KEY

jaeger-core/src/main/java/com/uber/jaeger/propagation/TextMapCodec.java
```diff
@@ -74,20 +76,30 @@ public SpanContext extract(TextMap carrier) {
...
         baggage.put(keys.unprefixedKey(key, baggagePrefix), decodedValue(entry.getValue()));
+      } else if (key.equals(DEBUG_JAEGER_SPAN_BAGGAGE_HEADER)) {
+        /**
+         * debug jaeger baggage is loaded for a header item that looks like,
+         * <pre>
+         * {@code
+         *    "jaeger-baggage": "k1=v1,k2=v2"
+         * }
+         * </pre>
+         */
```
yurishkuro: I would move this comment to where the constant is defined

jaeger-jaxrs2/src/test/java/com/uber/jaeger/propagation/ServerFilterIntegrationTest.java
```diff
@@ -0,0 +1,197 @@
...
+import org.glassfish.jersey.server.ResourceConfig;
+import org.glassfish.jersey.test.JerseyTest;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public class ServerFilterIntegrationTest extends JerseyTest {
+
+  // Base URI the Grizzly HTTP server will listen on
+  private static final String BASE_URI = "http://localhost:8080/";
```
yurishkuro: just curious - is it possible to specify port 0 and let the OS pick an ephemeral port? This is how we write tests in Go, because hardcoding a specific port, especially a commonly used one like 8080, could make the tests unstable/sensitive to environment.

jaeger-jaxrs2/src/test/java/com/uber/jaeger/propagation/ServerFilterIntegrationTest.java
```diff
@@ -0,0 +1,197 @@
...
+
+  @Override
+  protected Application configure() {
+
+    io.opentracing.Tracer serverTracer =
+        new com.uber.jaeger.Tracer
+            .Builder("server-tracer", new NoopReporter(), new ConstSampler(true))
+            .build();
+
+    TraceContext serverTraceContext = new ActiveSpanSourceTraceContext(serverTracer);
```
yurishkuro: `ActiveSpanSourceTraceContext` - this class is going away in the next release, so let's either wait with this PR or make it off the next release branch

jaeger-core/src/main/java/com/uber/jaeger/propagation/TextMapCodec.java
```diff
@@ -74,20 +76,30 @@ public SpanContext extract(TextMap carrier) {
...
+      if (debugId == null) {
+        return null;
       }
-      return null;
-    }
-    if (baggage == null) {
-      return context;
+      context = SpanContext.withDebugId(debugId);
     }
     return context.withBaggage(baggage);
```
yurishkuro: not sure if this is changing the behavior from current code, but it seems the baggage from uberctx- headers will be accepted even when there is no trace id, is that correct?

jaeger-jaxrs2/src/test/java/com/uber/jaeger/propagation/ServerFilterIntegrationTest.java
```diff
@@ -0,0 +1,197 @@
...
+        .header("uberctx-x-test", "testing")
+        .get();
+    assertEquals(resp.getStatus(), 200);
+  }
+
+  @Test
+  public void testTracingUsingNonInstrumentedClientAndNoDebugId() {
+    /**
+     * this test shows that in case where there is no uber-trace-id (non-instrumented client),
+     * only way to force baggage to get populated is to include the ``"jaeger-debug-id"`` header.
```
yurishkuro: yes, but why do we want that? Let's come up with a minimum & consistent set of rules of what we want. For example (not necessarily that we MUST do it this way, but at least it has some internal consistency):
  * when trace ID is present
     * baggage from uberctx- is respected
     * baggage from jaeger-baggage is not respected
  * when trace ID is not present
     * baggage from uberctx- is not respected
     * baggage from jaeger-baggage is respected
  * debug-id is always respected regardless of other data

Another alternative is to always respect everything (probably the simplest).

jaeger-jaxrs2/src/test/java/com/uber/jaeger/propagation/ServerFilterIntegrationTest.java
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  @Test
+  public void testTracingUsingNonInstrumentedClientWithFakeBaggageHeader() {
+    Response resp = target()
+        .path("test")
+        .request()
+        .header(DEBUG_ID_HEADER_KEY, "12345")
+        .header(DEBUG_JAEGER_SPAN_BAGGAGE_HEADER, "foo=bar")
+        .header("uberctx-x-test", "testing")
```
yurishkuro: one convenient way to write these kinds of tests is to make the server simply return the values in a JSON, and assert them in the tests. For example, in this test it's not clear if the debug-id was actually processed, i.e. that the span debug flag was set and the header value was stored as a tag.

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -283,22 +281,18 @@ private SpanContext createNewContext(String debugId) {
         }
       }
 
-      return new SpanContext(id, id, 0, flags);
+      return new SpanContext(id, id, 0, flags, createChildBaggage(), debugId);
```
dray92: not a 100% whether this is correct approach

jaeger-core/src/main/java/com/uber/jaeger/propagation/TextMapCodec.java
```diff
@@ -61,39 +71,102 @@ public void inject(SpanContext spanContext, TextMap carrier) {
...
       } else if (key.equals(Constants.DEBUG_ID_HEADER_KEY)) {
         debugId = decodedValue(entry.getValue());
       } else if (key.startsWith(baggagePrefix)) {
         if (baggage == null) {
           baggage = new HashMap<String, String>();
         }
         baggage.put(keys.unprefixedKey(key, baggagePrefix), decodedValue(entry.getValue()));
+      } else if (key.equals(MANUAL_BAGGAGE_HEADER_KEY)) {
+        String decodedDebugBaggageItems = decodedValue(entry.getValue());
+        debugBaggage = getDebugHeadersAsBaggage(decodedDebugBaggageItems);
```
yurishkuro: note that the header values separated by comma can be presented as duplicate headers. I.e. if sender sets header `X: a=b, x=y`, the server might receive

```
X: a=b
X: x=y
```
So for correctness this "debugBaggage" needs to be cumulative. Also, manualBaggage would be a better name, given the constant name.

jaeger-core/src/main/java/com/uber/jaeger/propagation/TextMapCodec.java
```diff
@@ -61,39 +71,102 @@ public void inject(SpanContext spanContext, TextMap carrier) {
...
+  @VisibleForTesting protected @Nullable Map<String, String> getDebugHeadersAsBaggage(
+      @NotNull String decodedDebugBaggageItems
+  ) {
+    // strip spaces from both ends to not cause cases like ",,,= " to get picked up
+    decodedDebugBaggageItems = decodedDebugBaggageItems.trim();
+
+    Map<String, String> debugBaggage = null;
+    String[] debugBaggageItems = commaPattern.split(decodedDebugBaggageItems);
+    for (String debugBaggageItem: debugBaggageItems) {
+      String[] debugBaggageItemParts = equalsPattern.split(debugBaggageItem.trim());
```
yurishkuro: wouldn't a regular string split (non-regex based) be more efficient?

jaeger-core/src/main/java/com/uber/jaeger/propagation/TextMapCodec.java
dray92: Ooh - good point. Gonna make both changes

jaeger-core/src/main/java/com/uber/jaeger/propagation/TextMapCodec.java
dray92: actually yes - for shorter strings, a manual split is indeed faster. And I think that's the case we care about. Will make change.

dray92:
@yurishkuro Definitely - I wanted to have you take a look at the logic that came out of it. I thought it was a bit too complex as well. 

Another very important point: baggage value can contain equal signs (base64 is commonly padded with `=` {[ref](https://en.wikipedia.org/wiki/Base64#Output_padding)}). The split logic we have there will ***definitely*** mess with this. The following test case fails in the present state of the world (for obvious reasons):
```
value = "item=foosdferv3re==";
result = codec.getDebugHeadersAsBaggage(value);
assertNotNull(result);
assertThat(result, IsMapContaining.hasEntry("item", "foosdferv3re=="));
```
`item` is picked up as `foosdferv3re`, not `foosdferv3re==`

This actually leads me to question the requirement of having something like `jaeger-baggage` with is separate from `uberctx-..`, in the first place. Do you know why we had decided to go down that route for the Go client?

yurishkuro:
> `value = "item=foosdferv3re==";`

Strictly speaking the notation we support is that both key and value (but not the `=` separator) are url-encoded. Basically, the format is supposed to be the same as https://github.com/w3c/distributed-tracing/blob/master/correlation_context/HTTP_HEADER_FORMAT.md#examples-of-http-headers

> This actually leads me to question the requirement of having something like jaeger-baggage with is separate from uberctx-.., in the first place. Do you know why we had decided to go down that route for the Go client?

The `uberctx-` was used long before jaeger-baggage. It was picked to pair off with uber-trace-id. In retrospect, the format `uberctx-{key}: value` was a poor choice since it puts unnecessary restrictions on the `key` (pretty much it must be lower-case-snake-case). The jaeger-baggage was introduced much later because it was easier to explain that format (a stable header and key=value pairs) and did not have the restrictions on the keys. Once w3c headers are officially standardized we will be supporting them in Jaeger, but it would still be difficult to migrate to them internally at Uber because it would require a client version upgrade in all microservices.

I think it makes sense to treat uberctx and jaeger-baggage as identical, with the aim to eventually decommission the former.

pavolloffay:
Is this active or can I close it? It also needs to rebase.

yurishkuro:
this needs to be fixed consistently across all clients.

pavolloffay:
@yurishkuro Is there umbrella issue for this in the main repo?

yurishkuro:
I suppose this one https://github.com/jaegertracing/jaeger/issues/165

pavolloffay:
@yurishkuro thanks, I will close this one as it needs a major rebase.

@dray92 @yurishkuro  thanks

PR #320, opened 2018-01-20
[jaeger-client-java/apache instrumentation] Completely remove apache instrumentation dependence on jaeger-context

dray92:
opentracing-0.31 Tracer interface completely removes the requirement for this module to depend on TracingUtils


debosmit:
Related to #188

PR #322, opened 2018-01-25
set travis wait time to 25 mins

dray92:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,10 @@ clean:
 test:
 	./gradlew check
 
+.PHONY: test-travis
+test-travis:
+	./gradlew -is check --debug
+
 .PHONY: release
 release: clean
 	./gradlew -i uploadArchives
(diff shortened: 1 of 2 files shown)


dray92:
How does this change affect these files?

vprithvi:
It doesn't. However, the codecov only maintains the code coverage last
successful build. When there are build failures, the report becomes wacky.

dray92:
@vprithvi @yurishkuro does this current change make more sense? I tried `--debug` but that prints more than 4mb in logs

Makefile
```diff
@@ -8,6 +8,10 @@ clean:
 test:
 	./gradlew check
 
+.PHONY: test-travis
+test-travis:
+	./gradlew -is check --info
```
vprithvi: Why are you setting the `info` flag redundantly using `-i` and `--info` ?

Issue #321, opened 2018-01-23
Unit tests often stall in Travis


yurishkuro:
like this
```
:jaeger-core:license UP-TO-DATE
:jaeger-core:test
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

vprithvi:
Could we enable debug mode?

https://docs.travis-ci.com/user/running-build-in-debug-mode/

@yurishkuro you might have to do it because they require push access (not sure why)

yurishkuro:
You are assuming it's a travis issue, I think it is much more likely java/gradle issue.

vprithvi:
How do you propose we get visibility into the stalled build?

Are you able to reproduce this locally?

yurishkuro:
I was not able to reproduce locally. Best we can do is to crank verbosity of Gradle logs.

jpkrohling:
This should have been fixed by #341

yurishkuro:
Hm, #340 and #341 just disabled the test rather than fixing it. I think we should close this issue but reopen #340, because disabling is not fixing.

jpkrohling:
Right, sorry, shouldn't have said "fixed". Reopened #340, as a proper fix is still wanted. And as suggested, I'm closing this one.

PR #323, opened 2018-01-25
Enable info logs and stacktraces from gradle

vprithvi:
ref #321


dray92:
i'm gonna add a change to print debug logs

PR #324, opened 2018-01-30
Give Gradle 1gb of memory

yurishkuro:
Signed-off-by: Yuri Shkuro <ys@uber.com>


yurishkuro:
I don't think it actually helped because the previous build still got stuck, but at least how we have better logging, so merging.

PR #326, opened 2018-01-30
Fixes #272 - Updated thrift to 0.11.0/0.10.0

### LINKED ISSUE #272: old httpcore dependency
The transitive dependencies through libthrift are getting to be quite old and conflicting with other common APIs - e.g. S3's client lib needs a newer httpcore.

I'm not sure whats involved with upgrading thrift for Jaeger, but perhaps the shaded thrift could be built with newer httpcore or something?

For now we're working around it with 

```
<!-- Jaeger - OpenTracing -->
<dependency>
  <groupId>com.uber.jaeger</groupId>
  <artifactId>jaeger-core</artifactId>
  <version>${jaeger.version}</version>
  <exclusions>
  <!-- Current Jaeger transitively depends on httpcore 4.2.4, we want 4.4.4 -->
    <exclusion>
      <groupId>org.apache.httpcomponents</groupId>
      <artifactId>httpcore</artifactId>
    </exclusion>
  </exclusions>
</dependency>
 ```

Note that Java is a second language for me and its possible this is self-inflicted, in which case please close :)

### PR DESCRIPTION
**Pull request by jpkrohling**

### DIFF
--- a/travis/docker-thrift/thrift
+++ b/travis/docker-thrift/thrift
@@ -2,7 +2,7 @@
 
 pwd=$(dirname ${PWD})
 
-THRIFT_VER=0.9.2
+THRIFT_VER=0.10.0
 THRIFT_IMG=thrift:${THRIFT_VER}
 THRIFT="docker run -v ${pwd}:/data ${THRIFT_IMG} thrift"
 
--- a/build.gradle
+++ b/build.gradle
@@ -18,7 +18,7 @@ ext.developmentVersion = '0.24.1-SNAPSHOT'
 
 ext.opentracingVersion = '0.31.0'
 ext.guavaVersion = '18.0'
-ext.apacheThriftVersion = '0.9.2'
+ext.apacheThriftVersion = '0.11.0'
 ext.jerseyVersion = '2.22.2'
 ext.slf4jVersion = '1.7.25'
 ext.gsonVersion = '2.8.2'

### CODE REVIEW

File: travis/docker-thrift/thrift [hunk 1]
```diff
@@ -2,7 +2,7 @@
 
 pwd=$(dirname ${PWD})
 
-THRIFT_VER=0.9.2
+THRIFT_VER=0.10.0
```
jpkrohling: Looks inconsistent with the version above, but there's no Docker image for 0.11.0 yet. The idea is then to keep the Java dependency at the latest and update the docker image once it gets published.

pavolloffay: Container should be soon there https://github.com/ahawkins/docker-thrift/issues/10

jpkrohling: Do you want to wait for that to be available, or should we merge this one and send an update once that becomes available? My personal preference is to get this merged now, as we don't seem to require 0.11 vs. 0.10.

pavolloffay: We don't have to wait. 
 
Should we update the thrift in other repos to be consistent? Although, It seems that different thrift versions are compatible on wire format.

pavolloffay:
I am wondering whether we could exclude HC completely as we are using okhttp.

vprithvi:
I'm curious about why this is required? 
I understand that the HC pulled in by the thift 0.9.3 is old, but are there incompatible API changes in newer HC?

jpkrohling:
I just left a comment on #272, as I think the HC should be managed at the reporter's pom. In any case, updating libthrift sounds like a good idea: 0.9.2 (which is what is being used here) is from 2014.

http://archive.apache.org/dist/thrift/0.9.2/

yurishkuro:
Is the runtime of 10/11 compatible with code generated using 092?

jpkrohling:
> Is the runtime of 10/11 compatible with code generated using 092?

That's my understanding, yes. Are the unit tests sufficiently covering the thrift interface, ensuring that at least the parts we use are compatible?

yurishkuro:
Java is the only project where we generate thrift classes at build time, instead of checking them in source control (I would much prefer the latter). As such, the runtime always matches the generated code and no unit tests would catch issues

jpkrohling:
> As such, the runtime always matches the generated code and no unit tests would catch issues

Is there any way for us to assess the impact of this change on existing clients? Right now, I'm kinda divided on this. On one hand, we benefit from 4 years of bug fixes (possibly even some security ones?), on the other hand, we don't want to break existing clients...

jpkrohling:
Should this one be closed?

vprithvi:
>Should this one be closed? 

I vote for closing this. I've run into subtle thrift incpompatibilies in the past, and don't feel confident going forward with this change if there are no strong drivers for it.

donotello:
@vprithvi, @jpkrohling, @yurishkuro, @pavolloffay,

Hi,

Found this pull request during analysis of one of the reports from our Security team. Apache Thrift 0.9.2 contains high severity security vulnerability [CVE-2016-5397](https://nvd.nist.gov/vuln/detail/CVE-2016-5397) which was fixed in 0.10.0. In addition it contains medium severity [CVE-2015-3254](https://nvd.nist.gov/vuln/detail/CVE-2015-3254).

Not sure if this can be considered as a good driver for uplift.

yurishkuro:
Are these vulnerabilities for Java? Looks like Go and CPP.

If we upgrade thrift we need to make sure it's shaded.

donotello:
@yurishkuro,

Sorry you are correct about [CVE-2016-5397](https://nvd.nist.gov/vuln/detail/CVE-2016-5397), didn't notice this in the first place.

But I'm still not sure about [CVE-2015-3254](https://nvd.nist.gov/vuln/detail/CVE-2015-3254). It says "The client libraries in Apache Thrift before 0.9.3" without specifying the exact client. Also in [JIRA](https://issues.apache.org/jira/browse/THRIFT-3231) relates to [this pull request](https://github.com/apache/thrift/pull/546/files) with changes in compiler that per my understanding impact the java library as well.

PR #329, opened 2018-01-31
Adding annotated filter priorities for Jersey filters

dray92:
Some open bugs in Jersey ([#2912](https://github.com/jersey/jersey/issues/2912), [#3467](https://github.com/jersey/jersey/issues/3467)) cause the framework to not honor filter binding priorities. For our internal libraries, we have found that keeping the binding priorities in the [5001, ∞) range while registering the filters keeps the ordering intact.

This change attempts to ensure that span is created very early in the request life-cycle, while giving room for consumers to manually register filters to priorities greater than `javax.ws.rs.Priorities.USER`, and yet, come before the span injection stage.

Similar idea follows for the server side instrumentation.

### DIFF
--- a/jaeger-jaxrs2/build.gradle
+++ b/jaeger-jaxrs2/build.gradle
@@ -1,6 +1,7 @@
 description = 'Instrumentation library for jaeger-jaxrs 2.0'
 
 dependencies {
+    compile group: 'javax.annotation', name: 'javax.annotation-api', version: '1.2-b02'
     compile group: 'javax.ws.rs', name: 'javax.ws.rs-api', version: '2.0.1'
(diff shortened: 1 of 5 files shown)


yurishkuro:
hardcoding priorities seems like a bad thing. Can they not be passed to constructor or via some other mechanisms?

dray92:
@yurishkuro So, I get your point. But, while registering, a user *can* specify priorities. This is what we are doing in our internal libraries now. To keep the filters and interceptors as self-contained units, it makes sense to give it sane defaults such as the ones we have here - and the `@Priority` annotation allows us to do just that. So, users can just plug in these components without having to worry about priorities.

PR #333, opened 2018-02-02
Add micrometer StatsReporter implementation

NeoCN:
implementation `StatsReporter` interface with [Micrometer](https://micrometer.io/), as mentioned in
https://github.com/jaegertracing/jaeger/issues/572

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,6 +23,7 @@ ext.jerseyVersion = '2.22.2'
 ext.slf4jVersion = '1.7.25'
 ext.gsonVersion = '2.8.2'
 ext.tracerResolverVersion = '0.1.3'
+ext.micrometerVersion = '1.0.0-rc.9'
 
 ext.junitVersion = '4.12'
 ext.mockitoVersion = '2.12.0'
--- a/jaeger-core/build.gradle
+++ b/jaeger-core/build.gradle
@@ -7,7 +7,7 @@ dependencies {
     compile group: 'com.google.code.gson', name: 'gson', version: gsonVersion
     compile group: 'org.slf4j', name: 'slf4j-api', version: slf4jVersion
     compile group: 'com.squareup.okhttp3', name: 'okhttp', version: '3.9.0'
-    compile group: 'io.micrometer', name: 'micrometer-core', version: '1.0.0-rc.9'
+    compile group: 'io.micrometer', name: 'micrometer-core', version: micrometerVersion
 
     // Testing frameworks
     // Jersey dependencies for unit tests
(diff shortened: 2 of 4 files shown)


jpkrohling:
I tried restarting Travis a few times, but it always times out: 

> No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

I have also tried to run a [sample application](https://github.com/jpkrohling/jaeger-client-java-tester) with this, but I got the following exception. @pavolloffay also took a look and got the same problem:

> java.lang.IllegalArgumentException: Prometheus requires that all meters with the same name have the same set of tag keys. There is already an existing meter containing tag keys [group, sampled]. The meter you are attempting to register has keys [group, state].

Relevant piece of the stack trace:

```
	at io.micrometer.core.instrument.Metrics.counter(Metrics.java:65)
	at com.uber.jaeger.metrics.MicrometerStatsReporter.incCounter(MicrometerStatsReporter.java:42)
	at com.uber.jaeger.metrics.StatsFactoryImpl$1.inc(StatsFactoryImpl.java:31)
	at com.uber.jaeger.Tracer.reportSpan(Tracer.java:142)
	at com.uber.jaeger.Span.finishWithDuration(Span.java:180)
	at com.uber.jaeger.Span.finish(Span.java:157)
```

I do get *some* output, though, which confirms the integration has all the pieces in place:

```
$ curl  http://localhost:8081
# HELP jaeger_spans_total  
# TYPE jaeger_spans_total counter
jaeger_spans_total{group="sampling",sampled="y",} 2.0
# HELP jaeger_traces_total  
# TYPE jaeger_traces_total counter
jaeger_traces_total{sampled="y",state="started",} 2.0
```

Besides the points above, this feature could make use of some documentation. If I can get the sample application to work, I can take care of the documentation, though.

jpkrohling:
I created a branch on my fork with a new commit that would fix the issue I mentioned above:

https://github.com/jpkrohling/jaeger-client-java/tree/333-NormalizedTags

I've taken a very conservative approach and normalized all tags across all metrics, while the absolute necessary would be across metrics with the same name.

@NeoCN , would you prefer to pick my commit and update this PR, or would you prefer me to create a new PR similar to the branch I linked above? 

cc @pavolloffay , @objectiser (no need to review my branch yet, although I would appreciate if you'd have any concern with the approach I took).

yurishkuro:
I would like this to be a separate module, rather that increasing the core module's dependency footprint

jpkrohling:
> I would like this to be a separate module, rather that increasing the core module's dependency footprint

Agree. I just added a new commit on top of my branch with that. 

@NeoCN , if you don't have any objections, I'll close this PR and open a new one based on my branch. It will still include your commits, don't worry :)

NeoCN:
@jpkrohling Agree with you, just go ahead!

PR #336, opened 2018-02-09
Closes #258 - Added builder for Configuration

### LINKED ISSUE #258: Revisit construction API for all core public components

jpkrohling:

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
@@ -40,6 +40,7 @@ import io.opentracing.propagation.TextMap;
 import java.io.IOException;
 import java.text.NumberFormat;
 import java.text.ParseException;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
@@ -256,11 +257,153 @@ public class Configuration {
    */
   @Deprecated
   public void setStatsFactor(StatsFactory statsFactory) {
-    this.statsFactory = statsFactory;
+    setStatsFactory(statsFactory);
   }
 
   public void setStatsFactory(StatsFactory statsFactory) {
     this.statsFactory = statsFactory;
+  }
+
+  public String getServiceName() {
+    return serviceName;
+  }
+
+  public SamplerConfiguration getSamplerConfig() {
+    return samplerConfig;
+  }
+
+  public ReporterConfiguration getReporterConfig() {
+    return reporterConfig;
+  }
+
+  public CodecConfiguration getCodecConfig() {
+    return codecConfig;
+  }
+
+  public StatsFactory getStatsFactory() {
+    return statsFactory;
+  }
+
+  /**
+   * Starts a new builder with the given service name.
+   *
+   * @param serviceName the service name
+   * @return a properly started builder
+   */
+  public static Builder builder(String serviceName) {
+    return new Builder(serviceName);
+  }
+
+  /**
+   * Gathers the configuration options from environment variables, based on {@link Configuration#fromEnv()}, meaning
+   * that it requires the service name to be set as an environment variable
+   *
+   * @see Configuration#fromEnv()
+   * @return a new builder's instance
+   */
+  public static Builder builderFromEnv() {
+    Configuration configuration = Configuration.fromEnv();
+
+    return new Builder(configuration.serviceName)
+        .withSamplerConfiguration(configuration.samplerConfig)
+        .withReporterConfiguration(configuration.reporterConfig)
+        .withCodecConfiguration(configuration.codecConfig)
+        .withStatsFactory(configuration.statsFactory);
+  }
+
+  /**
+   * Allows the construction of a {@link Configuration} object using the builder pattern
+   */
+  public static class Builder {
+    private String serviceName;
+    private SamplerConfiguration samplerConfiguration;
+    private ReporterConfiguration reporterConfiguration;
+    private CodecConfiguration codecConfiguration;
+    private StatsFactory statsFactory;
+
+    /**
+     * Creates a new builder with the given name. Null values are allowed and will usually be the case when the service
+     * name comes from an env var.
+     *
+     * @param serviceName the service name
+     */
+    public Builder(String serviceName) {
+      this.serviceName = serviceName;
+    }
+
+    /**
+     * Sets the service name to use, overriding the one specified on the constructor
+     * @param serviceName the service name
+     * @return the same builder's instance
+     */
+    public Builder withServiceName(String serviceName) {
+      this.serviceName = serviceName;
+      return this;
+    }
+
+    /**
+     * Sets the {@link SamplerConfiguration} to use. If the sampler type or parameter have been set already, they
+     * will be overridden by the given SamplerConfiguration.
+     *
+     * @see SamplerConfiguration
+     * @param samplerConfiguration the SamplerConfiguration to use
+     * @return the same builder's instance
+     */
+    public Builder withSamplerConfiguration(SamplerConfiguration samplerConfiguration) {
+      this.samplerConfiguration = samplerConfiguration;
+      return this;
+    }
+
+    /**
+     * Sets the {@link ReporterConfiguration} to use
+     * @see ReporterConfiguration
+     * @param reporterConfiguration the ReporterConfiguration to use
+     * @return the same builder's instance
+     */
+    public Builder withReporterConfiguration(ReporterConfiguration reporterConfiguration) {
+      this.reporterConfiguration = reporterConfiguration;
+      return this;
+    }
+
+    /**
+     * Sets the {@link CodecConfiguration} to use
+     * @see CodecConfiguration
+     * @param codecConfiguration the CodecConfiguration to use
+     * @return the same builder's instance
+     */
+    public Builder withCodecConfiguration(CodecConfiguration codecConfiguration) {
+      this.codecConfiguration = codecConfiguration;
+      return this;
+    }
+
+    /**
+     * Sets the {@link StatsFactory} to use.
+     *
+     * @see StatsFactory
+     * @param statsFactory the StatsFactory to use
+     * @return the same builder's instance
+     */
+    public Builder withStatsFactory(StatsFactory statsFactory) {
+      this.statsFactory = statsFactory;
+      return this;
+    }
+
+    /**
+     * Constructs a new {@link Configuration object} based on the information provided to the builder
+     * @return a ready to use Configuration instance
+     */
+    public Configuration build() {
+      Configuration configuration = new Configuration(serviceName,
+            samplerConfiguration,
+            reporterConfiguration,
+            codecConfiguration
+      );
+      if (null != statsFactory) {
+        configuration.setStatsFactory(statsFactory);
+      }
+
+      return configuration;
+    }
   }
 
   /**
@@ -287,18 +430,27 @@ public class Configuration {
      */
     private final String managerHostPort;
 
+    /**
+     * A sampler to be used, instead of determining the sampler via the {@link #type}
+     */
+    private final Sampler sampler;
+
     public SamplerConfiguration(String type, Number param) {
       this(type, param, null);
     }
 
-    public SamplerConfiguration() {
-      this(null, null, null);
-    }
-
     public SamplerConfiguration(String type, Number param, String managerHostPort) {
+      this.sampler = null;
       this.type = type;
       this.param = param;
       this.managerHostPort = managerHostPort;
+    }
+
+    public SamplerConfiguration(Sampler sampler) {
+      this.sampler = sampler;
+      this.type = null;
+      this.param = null;
+      this.managerHostPort = null;
     }
 
     public static SamplerConfiguration fromEnv() {
@@ -310,6 +462,10 @@ public class Configuration {
 
 
     private Sampler createSampler(String serviceName, Metrics metrics) {
+      if (this.sampler != null) {
+        return this.sampler;
+      }
+
       String samplerType = stringOrDefault(this.getType(), RemoteControlledSampler.TYPE);
       Number samplerParam = numberOrDefault(this.getParam(), DEFAULT_SAMPLING_PROBABILITY);
       String hostPort = stringOrDefault(this.getManagerHostPort(), defaultManagerHostPort);
@@ -348,6 +504,64 @@ public class Configuration {
     public String getManagerHostPort() {
       return managerHostPort;
     }
+
+    public static Builder builder() {
+      return new Builder();
+    }
+
+    public static class Builder {
+      private String type;
+      private Number param;
+      private String managerHostPort;
+      private Sampler sampler;
+
+      public Builder builder() {
+        return new Builder();
+      }
+
+      /**
+       * Set the sampler type
+       *
+       * @see ConstSampler#TYPE
+       * @see ProbabilisticSampler#TYPE
+       * @see RateLimitingSampler#TYPE
+       * @see RemoteControlledSampler#TYPE
+       * @param type the sampler type
+       * @return this builder
+       */
+      public Builder withType(String type) {
+        this.type = type;
+        this.sampler = null;
+        return this;
+      }
+
+      public Builder withParam(Number param) {
+        this.param = param;
+        this.sampler = null;
+        return this;
+      }
+
+      public Builder withManagerHostPort(String managerHostPort) {
+        this.managerHostPort = managerHostPort;
+        this.sampler = null;
+        return this;
+      }
+
+      public Builder withSampler(Sampler sampler) {
+        this.managerHostPort = null;
+        this.type = null;
+        this.param = null;
+        this.sampler = sampler;
+        return this;
+      }
+
+      public SamplerConfiguration build() {
+        if (null != this.sampler) {
+          return new SamplerConfiguration(this.sampler);
+        }
+        return new SamplerConfiguration(type, param, managerHostPort);
+      }
+    }
   }
 
   /**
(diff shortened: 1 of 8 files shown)


jpkrohling:
@pavolloffay , @vprithvi I just drafted something to solve #258. Could you please take a look at the tests and see if that's the sort of API that you were thinking of? Do check the actual implementation too if you want/can, but that's not the important part at this point :)

If the API itself (as seen on the tests) looks OK, I'll add more tests and ping you again for a more comprehensive review.

jaeger-core/src/test/java/com/uber/jaeger/ConfigurationTest.java
```diff
@@ -320,6 +321,53 @@ public void testPropagationValidFormat() {
...
+   * @see #testPropagationValidFormat() 
+   */
+  @Test
+  public void testPropagationViaBuilder() {
+    TestTextMap textMap = new TestTextMap();
+    SpanContext spanContext = new SpanContext(1234, 5678, 0, (byte)0);
+
+    Configuration
+          .builder("Test")
+          .withPropagationFormat("jaeger,invalid")
```
jpkrohling: I found builders with the prefix `with` (`TracerBuilder`, for instance) and without (`SenderConfiguration.Builder`). Which one should I use ?

jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
```diff
@@ -263,6 +264,354 @@ public void setStatsFactory(StatsFactory statsFactory) {
...
+     * Sets the sampler type and parameter to use when constructing a new {@link SamplerConfiguration}. If a
+     * {@link SamplerConfiguration} has been set already, it will be overridden.
+     *
+     * @see SamplerConfiguration#param
+     * @see SamplerConfiguration#type
+     * @param type  the sampler type
+     * @param param the sampler parameter
+     * @return the same builder's instance
+     */
+    public Builder withSamplerTypeAndParam(String type, Number param) {
```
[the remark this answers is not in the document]
jpkrohling: There is a `withSamplerConfiguration` up there. This option is to allow simple callers to avoid creating an intermediate `SamplerConfiguration`, making it simpler for a consumer to create a "default" tracer, overriding only these two parameters.

jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
```diff
@@ -263,6 +264,354 @@ public void setStatsFactory(StatsFactory statsFactory) {
...
+
+    /**
+     * Sets the codecs based on the given propagation format. If a {@link CodecConfiguration} has been set already, it
+     * will be overridden by this.
+     *
+     * @see CodecConfiguration#fromPropagationString(String)
+     * @param propagationFormat the propagation format
+     * @return the same builder's instance
+     */
+    public Builder withPropagationFormat(String propagationFormat) {
```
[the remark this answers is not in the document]
jpkrohling: Just looked into adding this, but the code actually accepts *more than one* format. It can be specified as: "jaeger,b3", and it will add appropriate codecs for Jaeger and B3.

The formats are documented under usual place for the doc on env vars (https://github.com/jaegertracing/jaeger-client-java/tree/master/jaeger-core), but I'll add a javadoc to `CodecConfiguration#fromPropagationString` as well.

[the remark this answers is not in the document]
pavolloffay: >  It can be specified as: "jaeger,b3", and it will add appropriate codecs for Jaeger and B3.

I think this is mostly the case when the configuration is created from env vars. here it could accept a list or injectors/extractors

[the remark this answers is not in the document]
jpkrohling: I would rather keep it close to the underlying configuration. Also, it might be easier for the simple case to just accept a list of `Propagation`. Anything more complex can be done by using the `CodecConfiguration` + `withCodecConfiguration`.

jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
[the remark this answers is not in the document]
vprithvi: I agree with @pavolloffay here. 

I'd rather have these convenience option be owned by `SamplerConfiguration` or something similar. 

I think that the current approach of having to do the following is intuitive enough:

`Tracer.builder("serviceName").withSamplerConfiguration(new SamplerConfiguration(type, param))`.

I feel that the current Builder knows too much about the fields of the classes that its holding, leading to pain when those classes change.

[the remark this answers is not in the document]
jpkrohling: > I'd rather have these convenience option be owned by SamplerConfiguration or something similar.

I see those methods as the equivalent of env vars: a consumer setting the two env vars is not required to know about `SamplerConfiguration`. To me, the following example summarizes the advantages (from the readme):

Current: 
```java
Configuration.SenderConfiguration senderConfiguration = new Configuration.SenderConfiguration.Builder()
  .endpoint("https://jaeger-collector:14268/api/traces")
  .build();

ReporterConfiguration reporterConfiguration = new ReporterConfiguration(senderConfiguration.getSender());
Configuration configuration = new Configuration("service-name", SamplerConfiguration.fromEnv(), reporterConfiguration);
io.opentracing.Tracer tracer = configuration.getTracer();
```

Using `withSenderConfiguration`:
```java
Configuration configuration = Configuration
  .builder("Test")
  .withSenderConfiguration(
    new Configuration.SenderConfiguration.Builder()
      .endpoint("https://jaeger-collector-custom:14268/api/traces")
      .build()
  )
  .build();
io.opentracing.Tracer tracer = configuration.getTracer();
```

The way this PR is right now:
```java
Configuration configuration = Configuration.builder("service-name")
  .withCollectorEndpoint("https://jaeger-collector-custom:14268/api/traces")
  .build();
io.opentracing.Tracer tracer = configuration.getTracer();
```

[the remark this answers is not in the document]
jpkrohling: Note that the above assumes `withSenderConfiguration` as part of the main builder, and not part of the `ReporterConfiguration` as it would need to be.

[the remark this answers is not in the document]
vprithvi: Thank you for that example. 

>I see those methods as the equivalent of env vars: a consumer setting the two env vars is not required to know about SamplerConfiguration

This is true, the env variables are prefixed with `JAEGER_SAMPLER`, which does give an indication that they are modifying the sampler. 

I agree that your approach provides a simpler API to callers, but I'm concerned that the builder is tightly coupled with parameters it shouldn't concern itself with. 

That being said, perhaps the user benefit outweighs this cost. Ideally, we'd have a nested fluent builder, but I don't see how to do that in Java 6.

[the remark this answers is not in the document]
jpkrohling: How about we mark the builder as "experimental" and see how hard it is to maintain in the long run? If it's too much of a burden, we remove the "deep settings" from the main builder.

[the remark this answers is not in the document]
pavolloffay: I don't think it's good to mark this as experimental. It's a core configuration struct. 

Another thing, this API shouldn't be aligned with what is exposed via env vars. For example the signature of this method is defined as `public Builder withSamplerTypeAndParam(String type, Number param)` the caller usually does not know string type of the sampler, signature `withSampler(Sampler sampler)` is easier to consume via programmatic API.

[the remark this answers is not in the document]
pavolloffay: I think there is some repetition between `TracerBuilder` and `Configuration` classes. Ideally, builders would be tightly coupled with its objects, for example `ReporterBuilder` would be inside the `Reporter` class/package.

This configuration class would just construct these builders based on supplied configuration one being environmental variables.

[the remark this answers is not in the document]
jpkrohling: If not a single builder at the top-level `Configuration` object, how about a consumer API that would look like this?

```java
    Configuration
        .builder("the-service")
        .reporter()
          .withAgentHost("agenthost")
          .build()
        .build();
```

`reporter()` returns a `ReporterConfiguration.Builder`, whose `build()` returns a `Configuration.Builder`, allowing the consumer to return back to the original builder.

Another option is to have a `ReporterConfiguration.Builder` that returns a `ReporterConfiguration` on `build()`, allowing for a consumer API to look like this:

```java
    Configuration
        .builder("the-service")
        .withReporterConfiguration(new ReporterConfiguration.Builder()
            .withAgentHost("agenthost")
            .build()
        )
        .build();
```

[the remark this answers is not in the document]
jpkrohling: By the way, as a consumer of Jaeger, I still think it's overly complex to require me to use two builders just to override the agent host... Let's remember that we are just a lib being consumed by a real application.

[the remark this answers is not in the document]
pavolloffay: The second approach looks better. Although it does not solve the initial concern about adding delegating options to the main conf builder.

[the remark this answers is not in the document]
jpkrohling: > Although it does not solve the initial concern about adding delegating options to the main conf builder.

Then I'm really not understanding something... Could you provide an example of how the consumer API would look like?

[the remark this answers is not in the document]
pavolloffay: Just without options like this  `withSamplerTypeAndParam(String type, Number param)` and rather accept the builder for `SamplerConfiguration`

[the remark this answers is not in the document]
jpkrohling: Discussion moved to https://groups.google.com/d/msgid/jaeger-tracing/1518708828.4245.32.camel%40redhat.com

jpkrohling:
@pavolloffay, @vprithvi, @objectiser , based on the discussion from the mailing list, I changed this PR to use "nested" builders (in the style of `Configuration.Builder#withReporterConfiguration(ReporterConfiguration.Builder#build()`).

Would you please take a look and see if that matches the expectation? There might be some doc changes needed, but for now, I'm mostly interested in a review of the API, as seen in the unit tests.

jaeger-core/src/test/java/com/uber/jaeger/ConfigurationTest.java
```diff
@@ -320,6 +339,490 @@ public void testPropagationValidFormat() {
...
+  @Test
+  public void testCloseWhenNoTracerWasBuilt() {
+    // this is a noop, but we ensure there's no NPE going on
+    new Configuration("the-service").closeTracer();
+  }
+
+  @Test
+  public void testClose() throws SenderException {
+    Sender sender = mock(Sender.class);
+    ReporterConfiguration reporterConfiguration = new ReporterConfiguration(sender);
```
objectiser: Should the constructor be removed (deprecated) - so only use builders?

jaeger-core/src/test/java/com/uber/jaeger/ConfigurationTest.java
```diff
@@ -320,6 +339,490 @@ public void testPropagationValidFormat() {
...
+  @Test(expected = IllegalArgumentException.class)
+  public void testNoServiceNameFails() {
+    Configuration.builder(null).build().getTracer();
+  }
+
+  @Test
+  public void testOverrideServiceName() {
+    Configuration configuration = Configuration
+        .builder("the-service")
+        .withServiceName("another-service")
```
objectiser: Do we need to have both builder parameter and `withServiceName` method - would it not be simpler just to have the `with` method?

jaeger-core/src/test/java/com/uber/jaeger/ConfigurationTest.java
```diff
@@ -320,6 +339,490 @@ public void testPropagationValidFormat() {
...
+    Tracer tracer = (Tracer) configuration.getTracer();
+    assertEquals("Should have overridden the service name", "another-service", tracer.getServiceName());
+  }
+
+  @Test
+  public void testBuilderFromEnv() {
+    System.setProperty(Configuration.JAEGER_SERVICE_NAME, "the-service");
+    Tracer tracer = (Tracer) Configuration
+        .builder()
+        .fromEnv()
```
objectiser: Just to clarify the behaviour - so iany preceding configuration that may have been defined will be overwritten by the values obtained from env - and similarly it would be possible to override parts of the env config by using `with...` methods after the `fromEnv` but before the `build`?

jaeger-core/src/test/java/com/uber/jaeger/ConfigurationTest.java
yurishkuro: deprecated

jaeger-core/src/test/java/com/uber/jaeger/ConfigurationTest.java
yurishkuro: service name is not an optional  field, I would rather pass it to `.builder("service-name")`

yurishkuro:
High level question - why do we need to separate the concept of Configuration and Builder? I.e. why can't it be the same thing? Currently configuration "builds" the tracer and tracer subcomponents, so it is already a "builder".

jaeger-core/src/test/java/com/uber/jaeger/ConfigurationTest.java
jpkrohling: Should it be possible to be overridden, though?

jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
```diff
@@ -256,13 +257,168 @@ public synchronized void closeTracer() {
...
+  }
+
+  /**
+   * Starts a new builder. Note that a service name is required to obtain a Tracer. If the service name is provided
+   * via an environment variable, make sure to call {@link Builder#fromEnv()}.
+   *
+   * @return a builder that still requires a name to start
+   */
+  public static Builder builder() {
+    return new Builder(null);
```
pavolloffay: This looks a bit inconsistent. Builder constructor requires service name and this not.

Does it make sense to change `fromEnv()` to be static?

jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
```diff
@@ -348,6 +500,49 @@ public Number getParam() {
...
+       * Set the sampler type
+       *
+       * @see ConstSampler#TYPE
+       * @see ProbabilisticSampler#TYPE
+       * @see RateLimitingSampler#TYPE
+       * @see RemoteControlledSampler#TYPE
+       * @param type the sampler type
+       * @return this builder
+       */
+      public Builder withType(String type) {
```
pavolloffay: Do we need `withType` and `withParam` as programmatic API? It's easier to directly consume `withSampler(Sampler)`.

pavolloffay:
I think that there should be only one way to create these builders/configuration objects. Now the configuration constructors are public. It makes sense to deprecate and later restrict visibility so there is one and expected way of this public API.

jaeger-core/src/test/java/com/uber/jaeger/ConfigurationTest.java
jpkrohling: The idea is that people would be able to get everything from the env and tweak one or two values. It happens to work the other way around as well, but that's a nice side effect.

jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
jpkrohling: You mean, `Builder.fromEnv()` to be the equivalent of `Builder.builder()`, but requiring the service name to come from the env var? Sounds like a good solution to me.

jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
jpkrohling: I added a new `withSampler`, which resets any value that has been set for type/param/managerHostPort (and vice-versa: setting type/param/managerHostPort resets a previously given Sampler)

jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
objectiser: What about `Configuration.builderFromEnv().build()`?

Not sure I like the idea of `fromEnv()` overwriting any config previously setup - seems pointless. This would also mean that we only have the `builder(serviceName)` option.

jpkrohling:
> High level question - why do we need to separate the concept of Configuration and Builder?

I don't think we do, but that would require making breaking changes to the `Configuration` object. I was expecting to do a cleanup as part of #344, if it's desirable. 

For *this* PR, though, I don't want to make breaking changes.

jpkrohling:
> I think that there should be only one way to create these builders/configuration objects.

Not sure I agree 100%: when writing the tests, I saw that it was easier to use the constructor of the `ReporterConfig` on a situation than get a builder, set the option and build it. I'd document it as the *preferred* way.

We'll still need constructors for the builders to use, so, we might as well just expose them as public constructors.

jaeger-core/src/test/java/com/uber/jaeger/ConfigurationTest.java
jpkrohling: Added `@Deprecated` to this constructor

jpkrohling:
> Not sure about the CodecConfiguration approach as it is different to the other config objects (builders). But not sure what would be better, unless have a builder with addProprogation methods instead of with - so if call multiple times it would build the list.

> Other main comment is possibly better to remove any constructors on the config classes - and just configure through the with methods.

I'm not sure I like this idea, but @pavolloffay seems to agree with you. I guess it's easier to make the constructors public afterwards. Note that we can't *remove* existing constructors, just deprecate.

jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
pavolloffay: > You mean, Builder.fromEnv() to be the equivalent of Builder.builder(), but requiring the service name to come from the env var?

yes `Configuration.Builder = Configuration.Builder.fromEnv() ` (or just configuration if configuration is a builder)

jpkrohling: Even if it's inconsistent with the `Configuration.builder()` method that starts a builder?

```
Configuration configuration = Configuration.builder().with...().build();
```

vs.

```
Configuration configuration = Configuration.Builder.fromEnv().build();
```

Between this and @objectiser, I think his is more consistent.

pavolloffay: > Even if it's inconsistent with the Configuration.builder() method that starts a builder?

I am not sure if I understand you. I started this thread because I don't like that `Configuration.builder()` creates a builder with `null` service name. `Builder.fromEnv()` without param is fine because it expects service name to be specified in env. Or we can remove service name from builder constructor and use `unknown` (Zipkin does this).

objectiser:
@jpkrohling yes renaming to `withPropagations` may be better

jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
jpkrohling: As part of changing the `Configuration.builder(String)` to add `Configuration.builderFromEnv()`, I removed the `fromEnv()` from the builder.

About removing the service name requirement, making it default to `unknown`, I would rather have it as part of another PR, especially because it changes an existing behavior.

yurishkuro:
> > High level question - why do we need to separate the concept of Configuration and Builder?
> 
> I don't think we do, but that would require making breaking changes to the Configuration object. I was expecting to do a cleanup as part of #344, if it's desirable.

Sorry, I don't follow. This PR is adding the new Builder class for the main Configuration an other configs. Why is the Builder needed? If we want `with` methods, why can't they be added directly to Configuration class (instead of overloading constructor as was done previously)?

jpkrohling:
> why can't they be added directly to Configuration

I see now what you mean. That's a good question. Usually, builders are used when constructing complex objects without leaving the target object in an invalid intermediary state. As far as I can see, it's not the case for the `Configuration` object, so, we *could* probably add `with` methods, returning a `Configuration` and it wouldn't break the compatibility, but I think people are more familiar with the builder pattern as implemented here, allowing chain calls among all methods on the builder.

jpkrohling:
As discussed on gitter, this will be closed until #258 is better specified.

PR #349, opened 2018-02-23
Closes #338 - Deprecated StatsReporter

### LINKED ISSUE #338: Deprecate StatsReporter
From @yurishkuro :

> StatsReporter should be deprecated, because it's very inefficient. StatsFactory is the preferred method (which we may also want to deprecate and rename to MetricsFactory to be consistent with Go)

https://github.com/jaegertracing/jaeger-client-java/pull/335#discussion_r167641501

jpkrohling:

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
@@ -15,9 +15,9 @@
 package com.uber.jaeger;
 
 import com.uber.jaeger.metrics.Metrics;
-import com.uber.jaeger.metrics.NullStatsReporter;
+import com.uber.jaeger.metrics.MetricsFactory;
+import com.uber.jaeger.metrics.NoopMetricsFactory;
 import com.uber.jaeger.metrics.StatsFactory;
-import com.uber.jaeger.metrics.StatsFactoryImpl;
 import com.uber.jaeger.propagation.B3TextMapCodec;
 import com.uber.jaeger.propagation.Codec;
 import com.uber.jaeger.propagation.CompositeCodec;
@@ -165,7 +165,7 @@ public class Configuration {
   private SamplerConfiguration samplerConfig;
   private ReporterConfiguration reporterConfig;
   private CodecConfiguration codecConfig;
-  private StatsFactory statsFactory;
+  private MetricsFactory metricsFactory;
   private Map<String, String> tracerTags;
 
   /**
@@ -201,6 +201,7 @@ public class Configuration {
     this.samplerConfig = samplerConfig;
     this.reporterConfig = reporterConfig;
     this.codecConfig = codecConfig;
+    this.metricsFactory = new NoopMetricsFactory();
   }
 
   /**
@@ -224,10 +225,10 @@ public class Configuration {
     if (codecConfig == null) {
       codecConfig = new CodecConfiguration(Collections.<Format<?>, List<Codec<TextMap>>>emptyMap());
     }
-    if (statsFactory == null) {
-      statsFactory = new StatsFactoryImpl(new NullStatsReporter());
+    if (metricsFactory == null) {
+      metricsFactory = new NoopMetricsFactory();
     }
-    Metrics metrics = new Metrics(statsFactory);
+    Metrics metrics = new Metrics(metricsFactory);
     Reporter reporter = reporterConfig.getReporter(metrics);
     Sampler sampler = samplerConfig.createSampler(serviceName, metrics);
     Tracer.Builder builder = new Tracer.Builder(serviceName)
@@ -257,16 +258,30 @@ public class Configuration {
   }
 
   /**
-   * @param statsFactory the factory
-   * @deprecated Use {@link #setStatsFactory(StatsFactory)} instead
+   * @see #withMetricsFactory(MetricsFactory)
+   * @param statsFactory the StatsFactory to use on the Tracer to be built
+   * @deprecated Use {@link #withMetricsFactory(MetricsFactory)} instead
    */
   @Deprecated
   public void setStatsFactor(StatsFactory statsFactory) {
-    this.statsFactory = statsFactory;
+    this.metricsFactory = statsFactory;
   }
 
+  /**
+   * @see #withMetricsFactory(MetricsFactory)
+   * @param statsFactory the StatsFactory to use on the Tracer to be built
+   * @deprecated Use {@link #withMetricsFactory(MetricsFactory)} instead
+   */
+  @Deprecated
   public void setStatsFactory(StatsFactory statsFactory) {
-    this.statsFactory = statsFactory;
+    this.metricsFactory = statsFactory;
+  }
+
+  /**
+   * @param metricsFactory the MetricsFactory to use on the Tracer to be built
+   */
+  public void withMetricsFactory(MetricsFactory metricsFactory) {
+    this.metricsFactory = metricsFactory;
   }
 
   public Configuration withReporter(ReporterConfiguration reporterConfig) {
--- a/jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
@@ -19,7 +19,8 @@ import com.uber.jaeger.baggage.BaggageSetter;
 import com.uber.jaeger.baggage.DefaultBaggageRestrictionManager;
 import com.uber.jaeger.exceptions.UnsupportedFormatException;
 import com.uber.jaeger.metrics.Metrics;
-import com.uber.jaeger.metrics.NullStatsReporter;
+import com.uber.jaeger.metrics.MetricsFactory;
+import com.uber.jaeger.metrics.NoopMetricsFactory;
 import com.uber.jaeger.metrics.StatsFactoryImpl;
 import com.uber.jaeger.metrics.StatsReporter;
 import com.uber.jaeger.propagation.Extractor;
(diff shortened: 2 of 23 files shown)


jpkrohling:
@yurishkuro, was something like this that you had in mind? https://github.com/jaegertracing/jaeger-client-java/pull/335#discussion_r167641501

@objectiser, @pavolloffay: your feedback is also appreciated

jaeger-core/src/test/java/com/uber/jaeger/metrics/InMemoryMetricsFactoryTest.java
```diff
@@ -0,0 +1,120 @@
...
+import com.uber.jaeger.reporters.InMemoryReporter;
+import com.uber.jaeger.samplers.ConstSampler;
+import java.util.HashMap;
+import java.util.Map;
+import org.junit.Test;
+
+public class InMemoryMetricsFactoryTest {
+
+  @Test
+  public void metricNameIsUsedForCounter() {
```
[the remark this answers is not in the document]
jpkrohling: I'm calling "metric name" the "gauge name" + tags, as seen on `Metrics.addTagsToMetricName`, as technically, the "metric name" is the composite of both. I also think it's confusing, but that's what I understood from the current code :)

jaeger-core/src/main/java/com/uber/jaeger/metrics/Metrics.java
```diff
@@ -19,12 +19,18 @@
 import java.util.HashMap;
 import java.util.Map;
 import java.util.SortedMap;
-import java.util.SortedSet;
 import java.util.TreeMap;
-import java.util.TreeSet;
 
 public class Metrics {
   public Metrics(StatsFactory factory) {
```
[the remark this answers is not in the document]
jpkrohling: Done

jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
```diff
@@ -251,16 +248,30 @@ public synchronized void closeTracer() {
...
+  @Deprecated
   public void setStatsFactory(StatsFactory statsFactory) {
-    this.statsFactory = statsFactory;
+    this.metricsFactory = statsFactory;
+  }
+
+  /**
+   * @param metricsFactory the MetricsFactory to use on the Tracer to be built
+   */
+  public void setMetricsFactory(MetricsFactory metricsFactory) {
```
yurishkuro: nit: pending #346, the method name might change, so best to wait until we agree on the new naming pattern

jpkrohling: Other than that, is this PR good to be merged?

jaeger-core/src/main/java/com/uber/jaeger/metrics/InMemoryMetricsFactory.java
```diff
@@ -0,0 +1,113 @@
...
+
+  /**
+   * Returns the counter value information for the counter with the given metric name.
+   * Note that the metric name is not the counter name, as a metric name usually includes the tags.
+   *
+   * @see Metrics#addTagsToMetricName(String, Map)
+   * @param name the metric name, which includes the tags
+   * @return the counter value or -1, if no counter exists for the given metric name
+   */
+  public long getCounter(String name) {
```
yurishkuro: I would rather have name+tags signature for the getters. It would decouple unit tests from knowing how metric ID is constructed internally. We took a similar approach in Go metrics (jaeger-lib).

An even more convenient signature for unit tests would be `getCounter(String name, String ...tags)` where `tags` are alternating key,value,key,value

jpkrohling: I replaced this with a method that accepts a string and an overloaded that accepts a map, as there are tests benefiting from both approaches.

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
```diff
@@ -501,11 +500,25 @@ public Builder withSampler(Sampler sampler) {
...
       this.metrics = new Metrics(new StatsFactoryImpl(statsReporter));
       return this;
     }
 
+    /**
+     * Creates a new {@link Metrics} to be used with the tracer, backed by the given {@link MetricsFactory}
+     * @param metricsFactory the metrics factory to use
+     * @return this instance of the builder
+     */
+    public Builder withMetricsFactory(MetricsFactory metricsFactory) {
```
pavolloffay: There is also `withMetrics` do we want to deprecate it too?

jaeger-core/src/main/java/com/uber/jaeger/metrics/InMemoryMetricsFactory.java
```diff
@@ -0,0 +1,175 @@
...
+  private long getValue(Map<String, AtomicLong> collection, String name, String tags) {
+    if (null == tags || tags.isEmpty()) {
+      return getValue(collection, name);
+    }
+
+    String[] entries = tags.split(",");
+    Map<String, String> tagsAsMap = new HashMap<String, String>(entries.length);
+    for (String entry : entries) {
+      String[] keyValue = entry.split("=");
+      tagsAsMap.put(keyValue[0], keyValue[1]);
```
pavolloffay: There should be a check otherwise it will throw `IndexOutOfBoundsException`

jaeger-core/src/main/java/com/uber/jaeger/metrics/InMemoryMetricsFactory.java
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  /**
+   * Returns the current value for the gauge with the given metric name. Note that the metric name is not the gauge
+   * name, as a metric name usually includes the tags.
+   *
+   * @see Metrics#addTagsToMetricName(String, Map)
+   * @param name the metric name, which includes the tags
+   * @param tags the metric tags as comma separated list of entries, like "foo=bar,baz=qux"
+   * @return the gauge value or -1, if no gauge exists for the given metric name
```
pavolloffay: `-1` is a valid gauge value, returning null would be better in this case

jaeger-core/src/main/java/com/uber/jaeger/metrics/InMemoryMetricsFactory.java
```diff
@@ -0,0 +1,175 @@
...
+        value.addAndGet(amount);
+      }
+    };
+  }
+
+  /**
+   * Returns the counter value information for the counter with the given metric name.
+   * Note that the metric name is not the counter name, as a metric name usually includes the tags.
+   *
+   * @see Metrics#addTagsToMetricName(String, Map)
```
pavolloffay: Why should caller see this, this methods abstracts metric name construction.

jaeger-core/src/main/java/com/uber/jaeger/metrics/InMemoryMetricsFactory.java
```diff
@@ -0,0 +1,175 @@
...
+   */
+  public long getCounter(String name, String tags) {
+    return getValue(counters, name, tags);
+  }
+
+  /**
+   * Returns the counter value information for the counter with the given metric name.
+   * Note that the metric name is not the counter name, as a metric name usually includes the tags.
+   *
+   * @see Metrics#addTagsToMetricName(String, Map)
```
pavolloffay: Why should caller see this, this methods abstracts metric name construction.

jaeger-core/src/main/java/com/uber/jaeger/metrics/InMemoryMetricsFactory.java
jpkrohling: This class is replacing InMemoryStatsReporter and the `getValue()` is a new method, to help in the tests. The javadoc for the metrics factory clearly says it's not for production purposes, it doesn't make sense to implement a more complex solution for a case that isn't being used at all, and probably never will.

jaeger-core/src/main/java/com/uber/jaeger/metrics/InMemoryMetricsFactory.java
jpkrohling: I don't think it would matter that much, but I added this check anyway

jaeger-core/src/main/java/com/uber/jaeger/Tracer.java
jpkrohling: No, `Metrics` wasn't deprecated.

jaeger-core/src/main/java/com/uber/jaeger/metrics/InMemoryMetricsFactory.java
jpkrohling: It was relevant in a previous version of the PR, but not anymore.

jaeger-core/src/main/java/com/uber/jaeger/metrics/InMemoryMetricsFactory.java
pavolloffay: As a consumer of this API I would find it more intuitive to check on `null` than in `-1`. What if our test  will use `-1` as a valid number?

pavolloffay:
I will merge to unblock @jpkrohling on the next PR which is based on this. Any more changes can be done in a separate PR. My only concern was about this https://github.com/jaegertracing/jaeger-client-java/pull/349#discussion_r170917284.

jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
```diff
@@ -257,16 +258,30 @@ public synchronized void closeTracer() {
...
   public void setStatsFactory(StatsFactory statsFactory) {
-    this.statsFactory = statsFactory;
+    this.metricsFactory = statsFactory;
+  }
+
+  /**
+   * @param metricsFactory the MetricsFactory to use on the Tracer to be built
+   */
+  public void withMetricsFactory(MetricsFactory metricsFactory) {
+    this.metricsFactory = metricsFactory;
```
pavolloffay: It seems like we forgot to follow the builder pattern

PR #355, opened 2018-02-26
Closes #354 - Ignores B3 headers if invalid values are provided

### LINKED ISSUE #354: B3TextMapCodec & incorrect input.
How should codecs, in general, behave when incorrect (or even malicious) input is provided? 

For example: `B3TextMapCodec` fails with an exception when incorrect header values are used.

Shouldn't it be handled gracefully?

Here is example PR with tests: https://github.com/jaegertracing/jaeger-client-java/pull/353

jpkrohling:


jaeger-core/src/main/java/com/uber/jaeger/propagation/B3TextMapCodec.java
```diff
@@ -60,8 +60,8 @@ public void inject(SpanContext spanContext, TextMap carrier) {
 
   @Override
   public SpanContext extract(TextMap carrier) {
-    Long traceId = null;
-    Long spanId = null;
+    long traceId = -1;
```
pavolloffay: `HexCodec.lowerHexToUnsignedLong("ffffffffffffffffffffffffffffffff")` returns -1

jpkrohling: Why does a method named `lowerHexToUnsignedLong` returns a *signed* long as a valid response? O_O

jpkrohling:
@pavolloffay , I changed the PR to use `null` instead of `-1`, as it didn't really occur to me that they were fitting a 64 bit into Java's 32 bit long.

Could you please re-review and merge? Looks like the debug flag is safe, but I couldn't find where the baggage might be vulnerable (see https://github.com/jaegertracing/jaeger-client-java/issues/354#issuecomment-368800486)

jaeger-core/src/test/java/com/uber/jaeger/propagation/B3TextMapCodecTest.java
```diff
@@ -59,8 +60,9 @@ public void downgrades128BitTraceIdToLower64Bits() throws Exception {
 
     SpanContext context = b3Codec.extract(textMap);
 
-    assertEquals(HexCodec.lowerHexToUnsignedLong(lower64Bits), context.getTraceId());
-    assertEquals(HexCodec.lowerHexToUnsignedLong(lower64Bits), context.getSpanId());
+    assertNotNull(HexCodec.lowerHexToUnsignedLong(lower64Bits));
+    assertEquals((long) HexCodec.lowerHexToUnsignedLong(lower64Bits), context.getTraceId());
```
pavolloffay: nit there is `Long.longValue()`

jaeger-core/src/test/java/com/uber/jaeger/propagation/B3TextMapCodecResiliencyTest.java
```diff
@@ -0,0 +1,81 @@
...
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.uber.jaeger.propagation;
+
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertThat;
+import static org.mockito.ArgumentMatchers.isNull;
```
pavolloffay: nit unused and the above too

jaeger-core/src/main/java/com/uber/jaeger/propagation/HexCodec.java
```diff
@@ -20,11 +20,13 @@
   /**
    * Parses a 1 to 32 character lower-hex string with no prefix into an unsigned long, tossing any
    * bits higher than 64.
+   *
+   * @return a 64 bit long, meaning that negative values are the overflow of Java's 32 bit long
    */
-  static long lowerHexToUnsignedLong(String lowerHex) {
+  static Long lowerHexToUnsignedLong(String lowerHex) {
```
tmszdmsk: Generally, `Optional` would be more obvious. However I understand that some like to "optimise" code with `null`s

jaeger-core/src/test/java/com/uber/jaeger/propagation/B3TextMapCodecResiliencyTest.java
```diff
@@ -0,0 +1,66 @@
...
+  @UseDataProvider("maliciousInputs")
+  public void shouldFallbackWhenMaliciousInput(String maliciousInput) {
+    String validInput = "ffffffffffffffffffffffffffffffff";
+    TextMap maliciousCarrier = new B3TextMapCodecTest.DelegatingTextMap();
+    maliciousCarrier.put(B3TextMapCodec.TRACE_ID_NAME, validInput);
+    maliciousCarrier.put(B3TextMapCodec.SPAN_ID_NAME, validInput);
+    maliciousCarrier.put(B3TextMapCodec.PARENT_SPAN_ID_NAME, validInput);
+
+    // everything is valid, except for one of the headers at a time:
+    for (String header : Arrays.asList(
```
tmszdmsk: for test case failure readability it should be implemented as second test method parameter + cartesian product in `@DataProvidrer`

jaeger-core/src/main/java/com/uber/jaeger/propagation/HexCodec.java
vprithvi: I agree, this tradeoff was made to maintain Java 6 compatibility without adding/writing an optional back port.

tmszdmsk: I get it. I didn't expect to have different lang level in `test` and `main`. Makes sense ;)

jaeger-core/src/test/java/com/uber/jaeger/propagation/B3TextMapCodecResiliencyTest.java
tmszdmsk: I addressed that in https://github.com/jaegertracing/jaeger-client-java/pull/358

yurishkuro:
Why is this a good change? There's another PR #359 that catches and logs all exceptions in the codecs. I think it's better to log an error saying that the inbound trace context was malformed than to silently ignore it, which makes it much harder for the user to figure out what's going on.

PR #358, opened 2018-02-27
b3 codec test readability improvement.

tmszdmsk:
IMHO it is easier to reason about this test in this form. I don't know what's the policy for adding test dependencies but can be further optimized [with guava](https://google.github.io/guava/releases/21.0/api/docs/com/google/common/collect/Sets.html#cartesianProduct-java.util.List-)


tmszdmsk:
PR tests passed here but checks link still directs to old one. Probably some race condition: https://travis-ci.org/jaegertracing/jaeger-client-java/builds/346868442

jpkrohling:
> PR tests passed here but checks link still directs to old one

It's probably because there's a merge commit here. If you rebase and force push, it should work.

tmszdmsk:
> If you rebase and force push, it should work.
Done!

PR #359, opened 2018-02-27
Improving B3TextMapCodec by handling all unexpected RuntimeExceptions

tmszdmsk:
Relates to #354, #355 

Well behaved carriers can't be taken for granted. Some will misbehave and instrumentation code should take that into account.

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/propagation/B3TextMapCodec.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/propagation/B3TextMapCodec.java
@@ -16,6 +16,8 @@ package com.uber.jaeger.propagation;
 
 import com.uber.jaeger.SpanContext;
 import io.opentracing.propagation.TextMap;
+import lombok.extern.slf4j.Slf4j;
+
 import java.util.Map;
 
 /**
@@ -35,6 +37,7 @@ import java.util.Map;
  * <p>
  * See <a href="http://zipkin.io/pages/instrumenting.html">Instrumenting a Library</a>
  */
+@Slf4j
 public class B3TextMapCodec implements Codec<TextMap> {
   protected static final String TRACE_ID_NAME = "X-B3-TraceId";
   protected static final String SPAN_ID_NAME = "X-B3-SpanId";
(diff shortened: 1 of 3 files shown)


jaeger-core/src/main/java/com/uber/jaeger/propagation/B3TextMapCodec.java
```diff
@@ -47,44 +49,52 @@
...
+      if (spanContext.getParentId() != 0L) { // Conventionally, parent id == 0 means the root span
+        carrier.put(PARENT_SPAN_ID_NAME, HexCodec.toLowerHex(spanContext.getParentId()));
+      }
+      carrier.put(SPAN_ID_NAME, HexCodec.toLowerHex(spanContext.getSpanId()));
+      carrier.put(SAMPLED_NAME, spanContext.isSampled() ? "1" : "0");
+      if (spanContext.isDebug()) {
+        carrier.put(FLAGS_NAME, "1");
+      }
+    } catch (RuntimeException ex) {
+      log.error("Error when injecting SpanContext into carrier. Handling gracefully.", ex);
```
vprithvi: I like the idea of a try catch, I am slightly concerned about what happens when carrier.put somehow results in corrupt data being injected. For e.g., let's take line 60, in today's world, if an exception is thrown, the thread dies. 
After this change, the data is injected (potentially corrupted), and extract (on potentially a different application with an older version of Jaeger) would have to deal with it. 

As such, I don't think it's that big of a concern looking at the TextMap code.

adriancole: this can fill logs. I'd make it debug level

adriancole: to put in perspective. once there was a proxy which shipped with an encoding but (parentId = literally '-') it wasn't fixed for months. that would be a lot of error log statements!

jaeger-core/src/test/java/com/uber/jaeger/propagation/FaultyTextMap.java
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright (c) 2016, Uber Technologies, Inc
```
jpkrohling: For new files, the header should be `The Jaeger Authors` (the year is wrong as well, it seems).

jaeger-core/src/test/java/com/uber/jaeger/propagation/B3TextMapCodecResiliencyTest.java
```diff
@@ -66,4 +66,21 @@ private TextMap validHeaders() {
...
+    //then
+    assertNull(extract);
+  }
+
+  @Test
+  public void shouldFallbackWhenInjectingIntoFaultyTextMap() {
+    TextMap faultyTextMap = new FaultyTextMap();
+    SpanContext spanContext = SpanContext.contextFromString("a:b:c:1");
+    //expect to pass
+    sut.inject(spanContext, faultyTextMap);
```
jpkrohling: Would it be possible to have a test to ensure the state of the `SpanContext` is sane after the exception? Mainly to address @vprithvi's comment https://github.com/jaegertracing/jaeger-client-java/pull/359/files#r171026709

tmszdmsk: As I understand @vprithvi was concerned about polluting `TextMap` with partial data. Unfortunately, I don't see a way to prevent that as we don't have any "transactional" method to set all fields at once + there is no method to compensate the action.

jaeger-core/src/main/java/com/uber/jaeger/propagation/B3TextMapCodec.java
tmszdmsk: * it would have been probably logged anyway if it had had bubble up the stack
* there is no other way to notify the developer that something doesn't work well. Especially in case of Span injection which will be most probably caused by developers error.
* In cases like the one you've mentioned it can be always muted on a logger level with configuration.

tmszdmsk: There is no easy way to prevent partial injection.
* Tracing data shouldn't affect the business logic so we're safe here.
* I agree that it can possibly cause problems on the other side of the wire if it's not "resilient" enough, but still, it is improvement over dying right away IMHO :)

tmszdmsk:
> :x: @codecov codecov/patch — 84.37% of diff hit (target 84.41%)
I don't really understand why it fails after the last commit

jaeger-core/src/main/java/com/uber/jaeger/propagation/B3TextMapCodec.java
```diff
@@ -47,44 +49,52 @@
...
         }
       }
-    }
 
-    if (null != traceId && null != parentId && null != spanId) {
-      return new SpanContext(traceId, spanId, parentId, flags);
+      if (null != traceId && null != parentId && null != spanId) {
+        return new SpanContext(traceId, spanId, parentId, flags);
+      }
+    } catch (RuntimeException ex) {
```
pavolloffay: We might want to revert back https://github.com/jaegertracing/jaeger-client-java/pull/355 as it removed exceptions from  `lowerHexToUnsignedLong`

jaeger-core/src/main/java/com/uber/jaeger/propagation/B3TextMapCodec.java
```diff
@@ -47,44 +49,52 @@
...
+      Long spanId = null;
+      Long parentId = 0L; // Conventionally, parent id == 0 means the root span
+      byte flags = 0;
+      for (Map.Entry<String, String> entry : carrier) {
+        if (entry.getKey().equalsIgnoreCase(SAMPLED_NAME)) {
+          String value = entry.getValue();
+          if ("1".equals(value) || "true".equalsIgnoreCase(value)) {
+            flags |= SAMPLED_FLAG;
+          }
+        } else if (entry.getKey().equalsIgnoreCase(TRACE_ID_NAME)) {
```
pavolloffay: not related to this PR but there can be NPE if the key is null, better to write it as `TRACE_ID_NAME.equals()`

tmszdmsk: yep, I noticed that but didn't want to make uber(haha) request ;)

jaeger-core/src/main/java/com/uber/jaeger/propagation/B3TextMapCodec.java
tmszdmsk: I am not sure, I left my comment on that here: https://github.com/jaegertracing/jaeger-client-java/issues/360#issuecomment-369161607

pavolloffay: without throwing ex a corrupted context wouldn't be logged as we simply return null in that case.

tmszdmsk: that's true, but I would rather add a simple log message somewhere as we know the cause instead of logging full stack trace.

tmszdmsk:
OK, let's put it on hold, I think I've found a better way to do that!

tmszdmsk:
Please, have a look at #362

PR #362, opened 2018-02-28
codecs resiliency, another approach

tmszdmsk:
After some thinking about #359 I prefer it this way because:
* it is more composable than writing the same logic in every codec
* covers all built-in and 3rd party codecs

Questions:
* can we assume that this should be the behavior for all possible `Injectors` and `Extractors`? Can you imagine any exception to this rule?
* it would be great to hide it from the external world, i.e. make it at least package scoped. IMHO they "belong" to `propagation` package naturally but if we want to make it "invisible" we can:
  * place them the main package where the `Tracer` with package scope
  * place it inside `Tracer`, but it is quite large already.
  * extract `PropagationRegistry` to a separate file and move decorators inside of it keeping it in the main package with package visibility. *I prefer this one*

I am new to this project so please think if I've missed something important from the design perspective.
 
To do
------
- [ ] tests


jaeger-core/src/main/java/com/uber/jaeger/propagation/ExceptionCatchingExtractorDecorator.java
```diff
@@ -0,0 +1,22 @@
+package com.uber.jaeger.propagation;
+
+import com.uber.jaeger.SpanContext;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+
+@RequiredArgsConstructor
```
jpkrohling: Could you add a javadoc, stating why this exists and how it works?

jpkrohling:
The build failed with `License violations were found: /.../jaeger-core/src/main/java/com/uber/jaeger/propagation/ExceptionCatchingExtractorDecorator.java,/home/travis/build/jaegertracing/jaeger-client-java/jaeger-core/src/main/java/com/uber/jaeger/propagation/ExceptionCatchingInjectorDecorator.java`

If you want to catch such errors faster, you may want to run `./gradlew check` locally before pushing updates to the PR :)

jaeger-core/src/main/java/com/uber/jaeger/propagation/ExceptionCatchingExtractorDecorator.java
tmszdmsk: you mean https://projectlombok.org/features/constructor?

tmszdmsk:
@jpkrohling please, wait with low-level CR. I want to establish if it's an accepted way to go on a "design level" first.

jaeger-core/src/main/java/com/uber/jaeger/propagation/ExceptionCatchingExtractorDecorator.java
jpkrohling: No, sorry, I meant a class-level javadoc for the classes `ExceptionCatching*Decorator`

tmszdmsk: sure, I will do that although I think that  `Injector` and `Extractor` interfaces need more attention in this area.

tmszdmsk:
ok, I've extracted PropagationRegistry to separate class to hide exception-catching decorators from users. 

also, I've added javadoc proposition for `Extractor` which is based on my understanding of this topic. Please comment if I made wrong assumptions.

jaeger-core/src/main/java/com/uber/jaeger/PropagationRegistry.java
```diff
@@ -0,0 +1,65 @@
...
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+
+class PropagationRegistry {
+  private final Map<Format<?>, Injector<?>> injectors = new HashMap<Format<?>, Injector<?>>();
+  private final Map<Format<?>, Extractor<?>> extractors = new HashMap<Format<?>, Extractor<?>>();
+
+  @SuppressWarnings("unchecked")
+  <T> Injector<T> getInjector(Format<T> format) {
+    return new ExceptionCatchingInjectorDecorator<T>((Injector<T>) injectors.get(format));
```
yurishkuro: there's no reason to wrap the injector on every `get`, instead wrap it on `register`

jaeger-core/src/main/java/com/uber/jaeger/PropagationRegistry.java
```diff
@@ -0,0 +1,65 @@
...
+    injectors.put(format, injector);
+  }
+
+  public <T> void register(Format<T> format, Extractor<T> extractor) {
+    extractors.put(format, extractor);
+  }
+
+  @RequiredArgsConstructor
+  @Slf4j
+  public static class ExceptionCatchingExtractorDecorator<T> implements Extractor<T> {
```
yurishkuro: why public?

jaeger-core/src/main/java/com/uber/jaeger/PropagationRegistry.java
```diff
@@ -0,0 +1,65 @@
...
+      } catch (RuntimeException ex) {
+        ExceptionCatchingExtractorDecorator.log.warn("Error when extracting SpanContext from carrier. Handling gracefully.", ex);
+        return null;
+      }
+    }
+  }
+
+  @RequiredArgsConstructor
+  @Slf4j
+  public static class ExceptionCatchingInjectorDecorator<T> implements Injector<T> {
```
yurishkuro: why public?

jaeger-core/src/main/java/com/uber/jaeger/PropagationRegistry.java
tmszdmsk: yep, automatic refactor ;)

yurishkuro:
Thanks @tmszdmsk !

tmszdmsk:
no problem! ;)

PR #370, opened 2018-03-15
Updating the ReporterConfig to set Sender Correctly

natehart:
A small bug in the private getReporter(..) method was resulting in the sender
always being set to the default UdpSender. This is because the field
`senderConfiguration.sender` is null until the `getSender()` method is called on
it. Because the reporter builder was accessing the field directly rather than
through the getter, it was always UdpSen a null sender, thus defaulting to the
UdpSender even if the JAEGER_ENDPOINT property was set.

This commit fixes the problem, accessing the configured sender through the
appropriate getter, allowing the user configurations to propagate through to the
tracer retrieved via the `Configuration.fromEnv().getTracer()` method.

Signed-off-by: Nate Hart <nhart@tableau.com>

### DIFF
--- a/jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
+++ b/jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
@@ -570,7 +570,7 @@ public class Configuration {
     private Reporter getReporter(Metrics metrics) {
       Reporter reporter = new RemoteReporter.Builder()
           .withMetrics(metrics)
-          .withSender(senderConfiguration.sender)
+          .withSender(senderConfiguration.getSender())
           .withFlushInterval(numberOrDefault(this.flushIntervalMs, RemoteReporter.DEFAULT_FLUSH_INTERVAL_MS).intValue())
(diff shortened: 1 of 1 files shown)


jaeger-core/src/main/java/com/uber/jaeger/Configuration.java
```diff
@@ -570,7 +570,7 @@ public ReporterConfiguration withSender(SenderConfiguration senderConfiguration)
     private Reporter getReporter(Metrics metrics) {
       Reporter reporter = new RemoteReporter.Builder()
           .withMetrics(metrics)
-          .withSender(senderConfiguration.sender)
+          .withSender(senderConfiguration.getSender())
```
yurishkuro: oh, come on, Java, it's a private field, why is it visible to a class defined at the same level?

yurishkuro:
thanks @natehart 

cc @pavolloffay

pavolloffay:
@jpkrohling could you please cut a micro release with this change?

PR #372, opened 2018-03-19
Preparing release 0.26.0

jpkrohling:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,10 +1,11 @@
 Changes by Version
 ==================
 
-0.25.1 (unreleased)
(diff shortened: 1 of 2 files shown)


yurishkuro:
actually no, won't work even with `.md`, ex: https://github.com/jaegertracing/jaeger-client-go/blob/master/CHANGELOG.md

To me the main thing is to copy the release notes into the GitHub release, where PR links *will work* for sure.

pavolloffay:
Directly link should work too on the release page. We could also use `@mention` then it's easier to find the contributor and maybe he gets notified that he is included in the release page

Issue #384, opened 2018-04-10
Add automated tests for Java Client metrics being exposed via prometheus


kevinearls:
The main purpose of this issue is to ensure metrics are being exposed via prometheus endpoint.

However a separate aim could be to start testing that the correct metrics are being generated. This issue may only deal with a subset (to be defined) - but should provide a framework in which additional tests could be added to verify other metrics.

PR #386, opened 2018-04-11
ISSUE 384 add tests for java client metrics

kevinearls:
Signed-off-by: Kevin Earls <kearls@redhat.com>

### DIFF
--- a/jaeger-micrometer/src/test/java/io/jaegertracing/micrometer/MicrometerTest.java
+++ b/jaeger-micrometer/src/test/java/io/jaegertracing/micrometer/MicrometerTest.java
@@ -14,26 +14,61 @@
 
 package io.jaegertracing.micrometer;
 
+import static java.util.stream.Collectors.counting;
+import static java.util.stream.Collectors.groupingBy;
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
 
+import io.jaegertracing.Configuration;
+import io.jaegertracing.Tracer;
 import io.jaegertracing.metrics.Metrics;
 import io.jaegertracing.metrics.Timer;
+import io.jaegertracing.samplers.ConstSampler;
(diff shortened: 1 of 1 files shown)


jaeger-micrometer/src/test/java/io/jaegertracing/micrometer/MicrometerTest.java
```diff
@@ -85,4 +120,83 @@ public void testTimer() {
...
+    assertEquals("Wrong number of traces", 10.0, traces, assertDelta);
+  }
+
+  private void createSomeSpans(Tracer tracer) {
+    for (int i = 0; i < 10; i++) {
+      Span span = tracer.buildSpan("metricstest")
+              .withTag("foo", "bar" + i)
+              .start();
+
+      if (i % 3 == 0) {
```
jpkrohling: Can you add a comment explaining what this is for?

jpkrohling: I'm not quite sure I understand what is going on here: are you finishing only every third span? Would it cause any resource leaks?

kevinearls:
@jpkrohling I've changed the title and added a comment.  I'm not finishing those spans just so that finished spans and started spans counts will be different.  It's not really completely necessary, so I can remove it you think that would be better.

jpkrohling:
>  It's not really completely necessary, so I can remove it you think that would be better.

That's fine, your comment there should be enough. I'm just a bit concerned about possible side effects, but for now, let's just keep an eye on suspicious failures elsewhere.

yurishkuro:
@kevinearls a couple of stylistic comments about putting "ISSUE 384" into the title:
  * when the PR is merged, its title is usually used as the final commit message, but it does not match out commit message guidelines
  * putting issue reference in the title does not create any linkage in GitHub between the issue and the PR. The preferred way is to put "Resolves #384" in the PR description, which not only links them, but also auto-resolves the issue on PR merge.

PR #389, opened 2018-04-11
Use travis stages to simplify config

pavolloffay:
* uses more declarative approach
* split deployment from tests
* should fail if the deploy fails - which does not hold for the current config - for example publish here https://travis-ci.org/jaegertracing/jaeger-client-java/jobs/365086226#L1333 failed and build is green

Docs for if statement: https://docs.travis-ci.com/user/conditional-builds-stages-jobs/


travis/build-crossdock.sh
```diff
@@ -2,8 +2,6 @@
```
yurishkuro: we should probably rename this script to `publish-crossdock.sh`

.travis.yml
```diff
@@ -36,11 +29,24 @@ cache:
...
+
+    - env: [ NAME=crossdock ]
+      script: make crossdock-fresh
+
+    - stage: deploy
+      env: [ NAME=deploy-to-maven ]
+      script: ./gradlew upload
+
+    - env: [ NAME=deploy-to-dockerhub ]
+      script: ./travis/build-crossdock.sh
```
yurishkuro: not sure I follow this setup. The build shows only two steps now, both from the `test` stage. Where can we see the logs & status from the deploy stage?

image

.travis.yml
```diff
@@ -36,11 +29,24 @@ cache:
...
+    if: env(TRAVIS_PULL_REQUEST) = false
+
+jobs:
+  include:
+    - stage: test
+      env: [ NAME=unit-test ]
+      script: make test-travis && ./gradlew codeCoverageReport && bash <(curl -s https://codecov.io/bash)
+
+    - env: [ NAME=crossdock ]
+      script: make crossdock-fresh
```
yurishkuro: Is the indentation correct here? I was expecting something like
```yaml
  - stage: test
    - env:
      script: 
    - env:
      script: 
```

pavolloffay: yes it is correct

yurishkuro: why is it correct though? It seems order dependent. Can we add `stage: test` as the first line of the second group, to make it more obvious?

pavolloffay: can try, I have copied it from this repo https://github.com/opentracing-contrib/java-spring-cloud/blob/master/.travis.yml#L25 it uses even more stages

yurishkuro: it looks like Travis explicitly allows that (i.e. the order matters and stage name is inherited from the previous item). Feels like a bad practice to me.

pavolloffay: This works (done in the last commit): 
```yml
jobs:
  include:
    - stage: test
      env: [ NAME=unit-test ]
      script: make test-travis && ./gradlew codeCoverageReport && bash <(curl -s https://codecov.io/bash)

    - stage: test
      env: [ NAME=crossdock ]
      script: make crossdock-fresh

    - stage: deploy
      env: [ NAME=deploy-to-maven ]
      script: ./gradlew upload

    - stage: deploy
      env: [ NAME=deploy-to-dockerhub ]
      script: ./travis/publish-crossdock.sh
```

pavolloffay:
@yurishkuro PR rebased and updated

PR #399, opened 2018-04-17
Implement Zipkin 2 JSON Sender

### LINKED ISSUE #383: Zipkin v2 JSON Sender
I would like to be able to send spans to Zipkin's v2 JSON API. 
 
I've started on a new sender in the `com.uber.jaeger.senders.zipkin` package which I intend to contribute, but wanted to make sure I didn't overlook anything that already enables this or get any feedback on how I should do it.  Unfortunately Zipkin 1/2 use a different interface for zipkin Sender so I will have to either have some duplicated code or write adapters to make them interchangeable.

keitwb:
The existing ZipkinSender class only supported the Thrift v1 protocol,
but the new Zipkin2Sender supports both v1 and v2 JSON.  Otherwise,
functionality is mostly the same.

There is some conceptually duplicate code in the Sender classes due to
the use of different classes between Zipkin 1/2 but it seemed overly
complicated to adapt them to a common adapter interface.

Fixes #383.


jaeger-zipkin/src/main/java/io/jaegertracing/senders/zipkin/V2SpanConverter.java
```diff
@@ -0,0 +1,193 @@
...
+/**
+ * Converts a Jaeger span to a Zipkin2 span.
+ */
+public class V2SpanConverter {
+
+  private static final Gson gson = new Gson();
+
+  public static zipkin2.Span convertSpan(Span span) {
+    Tracer tracer = span.getTracer();
+    Endpoint host = zipkin2.Endpoint.newBuilder()
```
mpetazzoni: To avoid confusion I would remove the `import zipkin2.Endpoint` and use `zipkin2.Endpoint host = ...` here.

jaeger-zipkin/src/main/java/io/jaegertracing/senders/zipkin/ConverterUtil.java
```diff
@@ -0,0 +1,39 @@
...
+ * Logic that is common to both Thrift v1 and JSON v2 senders
+ */
+class ConverterUtil {
+  static boolean isRpcServer(Span span) {
+    return Tags.SPAN_KIND_SERVER.equals(span.getTags().get(Tags.SPAN_KIND.getKey()));
+  }
+
+  static boolean isRpc(Span span) {
+    Object spanKindValue = span.getTags().get(Tags.SPAN_KIND.getKey());
+    return Tags.SPAN_KIND_CLIENT.equals(spanKindValue) || Tags.SPAN_KIND_SERVER.equals(spanKindValue);
```
mpetazzoni: I know you're just moving code around here, but this could simply be `return isRpcServer(span) || isRpcClient(span)`. Also clean up the superfluous empty newlines here and at the end of the class.

jaeger-zipkin/src/main/java/io/jaegertracing/senders/zipkin/V2SpanConverter.java
```diff
@@ -0,0 +1,193 @@
...
+        .ip(convertIp(tracer.getIpv4()))
+        .port(0)
```
mpetazzoni: You can probably just omit that. Might be good to add a comment saying that we don't have a local endpoint port to put in at all.

jaeger-zipkin/src/main/java/io/jaegertracing/senders/zipkin/V2SpanConverter.java
```diff
@@ -0,0 +1,193 @@
...
+            .duration(span.getDuration());
+
+    buildAnnotations(span, builder);
+    buildTags(span, builder);
+
+    return builder.build();
+  }
+
+  private static zipkin2.Span.Kind convertKind(Object kind) {
+    if (kind == null) {
```
mpetazzoni: If you reverse your equals calls you don't need that initial null check and have a single location where you return null from this method

jaeger-zipkin/src/main/java/io/jaegertracing/senders/zipkin/V2SpanConverter.java
```diff
@@ -0,0 +1,193 @@
...
+        // Every value is converted to string because zipkin search doesn't
+        // work well with ints, and bytes.
+        Object tagValue = entry.getValue();
+        builder.putTag(tagKey, tagValue.toString());
+      }
+    }
+  }
+
+  private static InetAddress convertIp(Integer ip) {
+    byte[] bytes = ByteBuffer.allocate(4).putInt(ip).array();
```
mpetazzoni: Can `ip` be null here? If so, you probably need to handle that. If not, then you can change the method parameter to be the unboxed `int`

jpkrohling:
@keitwb, thank you for this PR and sorry for the delay in reviewing it. I've been AFK for most of last week and will be for most of this week again.

Perhaps @pavolloffay would have some time this week to review, otherwise, I'll review it next week.

pavolloffay:
Based on https://github.com/jaegertracing/jaeger-client-java/issues/383#issuecomment-380271254 I think it would be better to have an adapter for zipkin reporter. It's more flexible and allows us
* use any zipkin supported sender: okhttp, urlconnection, kafka...
* leverage zipkin reporters
*  less code to maintain?

keitwb:
@pavolloffay I added an abstract base class for the two Zipkin senders so that there is quite a bit less duplication now.  This always allowed using any kind of Zipkin sender (http clients, kafka, etc.), and I don't understand what you mean about a Zipkin reporter adapter.

pavolloffay:
I mean instead of implementing sender interface e.g. `ZipkinSender` we could implement `ZipkinReporter` 

```java
class ZipkinReporter implements io.jaegertracing.Reporter {
   private io.zipkin.Reporter delegate;

   public ZipkinReporter(io.zipkin.Reporter reporter) {
     this.delegate = reporter;
   {

   public report(io.jaegertracing.Span jaegerSpan) {
       io.zipkin.Span ipkinSpan  = convertToZipkin(span);
       delegate.report(span);
   }
}
```

pavolloffay:
Something like https://github.com/jaegertracing/jaeger-client-java/commit/431a558d6b939031b062bb67f5e484298f6255ab. 

But note that it deals only with jsonv1 and jsonv2. It should be also possible to add thrift.

One disadvantage is that flush is not supporter, but we could  bind to zipkin's `AsyncReporter` (instead to `zipkin.Reporter`) which implements flush.

keitwb:
@pavolloffay ok, yes I have started reimplementing this using reporters and it seems much cleaner and makes much better reuse of zipkin code instead of reinventing the wheel in this lib.  My main question at this point is whether there is anything special about the use of the older `com.twitter.zipkin.thriftjava.Span` class vs. just converting the jaeger spans to `zipkin.Span` and using the [Thrift encoding](https://github.com/openzipkin/zipkin-reporter-java/blob/release-1.1.1/core/src/main/java/zipkin/reporter/Encoder.java#L32) from the Zipkin v1 reporter.

pavolloffay:
I think those two should be compatible.

keitwb:
@pavolloffay  ok, I made an initial attempt at reusing Zipkin reporters as much as possible.  It is a bit tricky since there are Zipkin 1 and 2 reporters with a different Java interface, but I think I have simplified it as much as possible.  

This does break existing users of this lib since it totally removes ZipkinSender from this repo, but I am unsure if that is important or not.

jaeger-zipkin/build.gradle
```diff
@@ -1,12 +1,16 @@
 description = 'Integration library for Zipkin'
 
 dependencies {
-    compile group: 'io.zipkin.reporter', name: 'zipkin-reporter', version: '1.1.2'
-    compile group: 'io.zipkin.reporter', name: 'zipkin-sender-urlconnection', version: '1.1.2'
+    compile group: 'io.zipkin.reporter', name: 'zipkin-reporter', version: '1.0.4'
```
keitwb: I had to downgrade this because of some nasty classpath conflicts between the version of Zipkin that 1.1.2 and 2.6.0 of the reporter were using.  Is there some way to pin the version of Zipkin used by dependencies of this module?

jaeger-zipkin/src/main/java/io/jaegertracing/senders/zipkin/V2SpanConverter.java
```diff
@@ -0,0 +1,187 @@
...
+  private static void buildAnnotations(Span span, zipkin2.Span.Builder builder) {
+    if (ConverterUtil.isRpc(span)) {
+      String startLabel = zipkin.Constants.SERVER_RECV;
+      String endLabel = zipkin.Constants.SERVER_SEND;
+      if (ConverterUtil.isRpcClient(span)) {
+        startLabel = zipkin.Constants.CLIENT_SEND;
+        endLabel = zipkin.Constants.CLIENT_RECV;
+      }
+
+      builder.addAnnotation(span.getStart(), startLabel);
```
keitwb: Is this necessary in v2?

keitwb:
@pavolloffay How does it look now?

pavolloffay:
I think it's better to keep the zipkin sender and just deprecate it.

jaeger-zipkin/src/main/java/io/jaegertracing/senders/zipkin/V2SpanConverter.java
pavolloffay: I don't think so 

cs time - is kind=client and start time
cr time - is kind=client starttime+duration

the same for server side, but the kind is server

jaeger-zipkin/src/test/java/io/jaegertracing/senders/zipkin/V2SpanConverterTest.java
```diff
@@ -0,0 +1,310 @@
...
+import zipkin2.Annotation;
+
+@RunWith(DataProviderRunner.class)
+public class V2SpanConverterTest {
+  Tracer tracer;
+
+  @Before
+  public void setUp() {
+    tracer =
+        new Tracer.Builder("test-service-name", new InMemoryReporter(), new ConstSampler(true))
```
pavolloffay: please do not use deprecated APIs

jaeger-zipkin/build.gradle
pavolloffay: What If you exclude the old zipkin(2.2.1) from reporter 1.1.2?

keitwb: I'll see if it is still an issue after I revert the ZipkinSender and only implement the V2 Reporter.

keitwb:
@pavolloffay ok, I reverted the original ZipkinSender and rearranged things a bit.  I think I figured out how to exclude Zipkin 2.2.1 from polluting the classpath.  Let me know what you think.

pavolloffay:
@keitwb could you please keep `ThriftSpanConverter` in the old package? Just keep the changes minimal

jaeger-zipkin/README.md
```diff
@@ -21,12 +21,39 @@ tracer = new Tracer.Builder(serviceName, reporter, sampler)
...
+2 reporter to the Jaeger reporter interface and deals with converting Jaeger spans to the Zipkin 2 
+model.
+
+For example:
+```java
+import io.jaegertracing.zipkin.reporters.Zipkin2Reporter;
+import zipkin2.reporter.AsyncReporter;
+import zipkin2.reporter.urlconnection.URLConnectionSender;
+
+reporter = new Zipkin2Reporter(
```
pavolloffay: nit: could you please just rename it to `ZipkinV2Reporter`?

jaeger-zipkin/src/main/java/io/jaegertracing/zipkin/V2SpanConverter.java
```diff
@@ -0,0 +1,187 @@
...
+  private static void buildAnnotations(Span span, zipkin2.Span.Builder builder) {
+    if (ConverterUtil.isRpc(span)) {
+      String startLabel = zipkin.Constants.SERVER_RECV;
+      String endLabel = zipkin.Constants.SERVER_SEND;
+      if (ConverterUtil.isRpcClient(span)) {
+        startLabel = zipkin.Constants.CLIENT_SEND;
+        endLabel = zipkin.Constants.CLIENT_RECV;
+      }
+
+      builder.addAnnotation(span.getStart(), startLabel);
```
pavolloffay: From the previous review: these are not needed in V2

pavolloffay:
@keitwb the PR looks good, I had just a couple of minor comments.

pavolloffay:
@keitwb sorry It also needs a rebase due to my merges today :/

jaeger-zipkin/src/main/java/io/jaegertracing/zipkin/V2SpanConverter.java
```diff
@@ -0,0 +1,187 @@
...
+      String componentName;
+      Object componentTag = tags.get(Tags.COMPONENT.getKey());
+      if (componentTag instanceof String) {
+        componentName = componentTag.toString();
+      } else {
+        // spans always have associated tracers, and service names
+        componentName = span.getTracer().getServiceName();
+      }
+
+      builder.putTag(zipkin.Constants.LOCAL_COMPONENT, componentName);
```
keitwb: What about this one?  In Zipkin v2 spans, the local endpoint field contains the service name so this seems redundant to add this special tag.

I'm going to remove it for now but let me know if you think it should remain.

keitwb:
@pavolloffay ok, I moved the Thrift converter back to its original location and rebased.  Also, I removed some logic for the local component special tag since it was in the local endpoint object already (see comment).  I put the new stuff as a separate commit so you can see it clearly.  Thanks!

keitwb:
Ok, I did the same thing the Zipkin reporter tests do and it seemed to fix it.

pavolloffay:
@keitwb great work! Thanks for the PR

I will wait for 24h in case anybody else wants to review

jaeger-zipkin/build.gradle
```diff
@@ -1,12 +1,23 @@
...
 
 dependencies {
-    compile group: 'io.zipkin.reporter', name: 'zipkin-reporter', version: '1.1.2'
-    compile group: 'io.zipkin.reporter', name: 'zipkin-sender-urlconnection', version: '1.1.2'
+    compile(group: 'io.zipkin.reporter', name: 'zipkin-reporter', version: '1.1.2') {
+        exclude group:'io.zipkin.zipkin2'
+        exclude group:'io.zipkin.java'
+    }
+    compile(group: 'io.zipkin.reporter', name: 'zipkin-sender-urlconnection', version: '1.1.2') {
+        exclude group:'io.zipkin.zipkin2'
```
jpkrohling: Is this needed, if you are declaring it as a direct dependency of your project?

jaeger-zipkin/src/main/java/io/jaegertracing/zipkin/V2SpanConverter.java
```diff
@@ -0,0 +1,161 @@
...
+    if (Tags.SPAN_KIND_SERVER.equals(kind)) {
+      return zipkin2.Span.Kind.SERVER;
+    } else if (Tags.SPAN_KIND_CLIENT.equals(kind)) {
+      return zipkin2.Span.Kind.CLIENT;
+    } else if (Tags.SPAN_KIND_CONSUMER.equals(kind)) {
+      return zipkin2.Span.Kind.CONSUMER;
+    } else if (Tags.SPAN_KIND_PRODUCER.equals(kind)) {
+      return zipkin2.Span.Kind.PRODUCER;
+    } else {
+      return null;
```
jpkrohling: Is this an expected case? If not, a log statement should be produced, like, "Could not identify the span kind during conversion. Skipping."

jaeger-zipkin/src/main/java/io/jaegertracing/zipkin/V2SpanConverter.java
```diff
@@ -0,0 +1,161 @@
...
+      }
+    }
+  }
+
+  private static InetAddress convertIp(int ip) {
+    byte[] bytes = ByteBuffer.allocate(4).putInt(ip).array();
+    try {
+      return InetAddress.getByAddress(bytes);
+    } catch (UnknownHostException e) {
+      return null;
```
jpkrohling: IIRC, this *should* never happen but would be worth adding a log statement here.

jaeger-zipkin/src/test/java/io/jaegertracing/zipkin/V2SpanConverterTest.java
```diff
@@ -0,0 +1,266 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import zipkin2.Annotation;
+
+@RunWith(DataProviderRunner.class)
+public class V2SpanConverterTest {
+  Tracer tracer;
+
+  @Before
+  public void setUp() {
```
jpkrohling: `s/setUp/setup`

jaeger-zipkin/src/main/java/io/jaegertracing/zipkin/V2SpanConverter.java
pavolloffay: IIRC internal spans has a null kind.

pavolloffay: yes they have kind set to  `null`

jaeger-zipkin/build.gradle
keitwb: Tested it, and you are right, it doesn't appear to be necessary due to the explicit dependency.

pavolloffay:
I am just linking a similar PR in go https://github.com/jaegertracing/jaeger-client-go/pull/310. I have updated the branch and I will merge on green

PR #411, opened 2018-05-15
Polish configuration API

### LINKED ISSUE #394: Revisit Configuration API
There are a couple of inconsistencies in the configuration API.

The API should comply with this
* after creating the configuration from env it should be possible to override any value via programmatic API
* Configurations can be created via programmatic API - via constructor not only via `fromEnv`

* programmatic api does not allow to override set service name
* codec configuration can be created only via `.fromEnv()`
* most of the configurations provide getters but not for all properties

pavolloffay:
Resolves #394 

- [x] there are getters for all configuration properties
- [x] codec configuration can be created programmatically
- [x] allow overriding service name

Configuration can be created programmatically o via env. If it is created via env all properties can be changed programmatically.

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/Configuration.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/Configuration.java
@@ -161,7 +161,7 @@ public class Configuration {
   /**
    * The serviceName that the tracer will use
    */
-  private final String serviceName;
+  private String serviceName;
   private SamplerConfiguration samplerConfig;
   private ReporterConfiguration reporterConfig;
   private CodecConfiguration codecConfig;
@@ -284,6 +284,11 @@ public class Configuration {
     this.metricsFactory = metricsFactory;
   }
 
+  public Configuration withServiceName(String serviceName) {
+    this.serviceName = Tracer.Builder.checkValidServiceName(serviceName);
+    return this;
+  }
+
   public Configuration withReporter(ReporterConfiguration reporterConfig) {
     this.reporterConfig = reporterConfig;
     return this;
@@ -306,12 +311,24 @@ public class Configuration {
     return this;
   }
 
+  public String getServiceName() {
+    return serviceName;
+  }
+
   public ReporterConfiguration getReporter() {
     return reporterConfig;
   }
 
   public SamplerConfiguration getSampler() {
     return samplerConfig;
+  }
+
+  public CodecConfiguration getCodec() {
+    return codecConfig;
+  }
+
+  public MetricsFactory getMetricsFactory() {
+    return metricsFactory;
   }
 
   public Map<String, String> getTracerTags() {
@@ -430,6 +447,14 @@ public class Configuration {
   public static class CodecConfiguration {
     private Map<Format<?>, List<Codec<TextMap>>> codecs;
 
+    public CodecConfiguration() {
+    }
+
+    public CodecConfiguration withCodec(Format<?> format, Codec<TextMap> codec) {
+      addCodec(codecs, format, codec);
+      return this;
+    }
+
     private CodecConfiguration(Map<Format<?>, List<Codec<TextMap>>> codecs) {
       this.codecs = codecs;
     }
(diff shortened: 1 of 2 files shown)


jaeger-core/src/main/java/io/jaegertracing/Configuration.java
```diff
@@ -430,6 +447,14 @@ public SamplerConfiguration withManagerHostPort(String managerHostPort) {
   public static class CodecConfiguration {
     private Map<Format<?>, List<Codec<TextMap>>> codecs;
 
+    public CodecConfiguration() {
+    }
+
+    public CodecConfiguration withCodec(Format<?> format, Codec<TextMap> codec) {
```
aaronkorver: Yay!!

aaronkorver:
A note, not necessary on this PR as I can open a separate issue, the `withTags()` would be great if there was exposed a `withTagsFromEnv()` as well that could use the private method that is currently in the config.

pavolloffay:
@aaronkorver you can call ` c = Configuration.fromEnv()` then `c.getTags()` and then `c.withTags(tags)` to fully manipulate with tags.

aaronkorver:
right, but I can't call the parser function in Configuration as it is private right now.  I have some teams that want to take advantage of the `JAEGER_TAGS` environment.  My team has been providing a library for tracing and want them to use it as well, but have basically cut/paste that function into our library so we can configure certain things from the ENV, but override other pieces.  Hence the suggestion to do `withTagsFromEnv()`

pavolloffay:
I am not sure whether method `c.withTagsFromEnv()` makes sense as the tags from env are obtained during `Configuration.fromEnv()`.

We could perhaps expose `map = Configuration.tracerTagsFromEnv()` , then you can get tags from env and programmatically mix things together. Oh wait actually right now you can call `fromEnv()` and get parsed tags from the configuration.

jaeger-core/src/main/java/io/jaegertracing/Configuration.java
```diff
@@ -280,8 +280,14 @@ public void setStatsFactory(StatsFactory statsFactory) {
   /**
    * @param metricsFactory the MetricsFactory to use on the Tracer to be built
    */
-  public void withMetricsFactory(MetricsFactory metricsFactory) {
+  public Configuration withMetricsFactory(MetricsFactory metricsFactory) {
```
pavolloffay: This shouldn't break clients

aaronkorver:
So are you recommending that you should first do `myConfig = Configuration.fromEnv()` and then change things within `myConfig`?  The unfortunate part on that is the `getTracerTags()` returns an unmodifiableMap() that I can't merge things.

pavolloffay:
> So are you recommending that you should first do myConfig = Configuration.fromEnv() and then change things within myConfig

Yes, the map is unmodifiable for a purpose, you can do `m = new HasMap<>(unmodifiableMap)`

jaeger-core/src/test/java/io/jaegertracing/ConfigurationTest.java
```diff
@@ -328,6 +333,14 @@ public void testNoServiceName() {
     new Configuration(null);
   }
 
+  @Test
+  public void testOverrideServiceName() {
```
jpkrohling: Can you add a test using `withServiceName(null)`, and another with `withServiceName("")`? This should document what is expected to happen in such case.

pavolloffay: It's not strictly related to this PR, but I am happy to add it. 

Test for null service name is already there, I have added only for empty case.

jaeger-core/src/test/java/io/jaegertracing/ConfigurationTest.java
```diff
@@ -328,6 +333,19 @@ public void testNoServiceName() {
     new Configuration(null);
   }
 
+  @Test(expected = RuntimeException.class)
+  public void testEmptyServiceName() {
+    new Configuration("");
```
jpkrohling: I actually thought about having the `null` overriding a valid existing name:

```java
new Configuration("Test").withServiceName(null);
```

This way, you'd be documenting the expected behavior for this scenario, as it might be ambiguous. This is not required, so, I'm giving my +1 to this PR.

pavolloffay: The check is also in tracer constructor, but in this case we wanted to fail before tracer-init time

PR #412, opened 2018-05-15
Do not use deprecated APIs internally

pavolloffay:
This is not a breaking change! It does not introduces/removes any APIs or implementation. It only removes calls to deprecated APIs from internal packages.


pavolloffay:
The coverage is just wrong :). I haven't added new code

Issue #413, opened 2018-05-15
Do not deprecate Tracer.Builder(name, reporter, sampler)


pavolloffay:
We have deprecated `Tracer.Builder(name, reporter, sampler)`, but I think that constructor is useful and for the folks who would like to instantiated the tracer in a one-liner.

Related to #395 

cc @yurishkuro @jpkrohling @black-adder @vprithvi

jpkrohling:
I don't have a strong opinion on this, although I prefer we'd have one way of doing things. In the end, `new Tracer.Builder(serviceName, reporter, sampler)` isn't _that_ different from `Tracer.Builder(serviceName).withReporter(reporter).withSampler(sampler)`

pavolloffay:
I agree.., the main reason I have opened this is that we use that constructor internally a lot.

pavolloffay:
The constructor was removed in #414

PR #414, opened 2018-05-16
Remove deprecated APIs

### LINKED ISSUE #395: Remove deprecated API
see https://github.com/jaegertracing/jaeger-client-java/issues/253#issuecomment-380480924

pavolloffay:
Resolves #395 #413 

~I didn't remove `Tracer.Builder(name, reporter, samper)` #413~ I did in the second commit

Related PRs which removed usage of deprecated APIs from internal code
* https://github.com/jaegertracing/jaeger-client-java/pull/412
* https://github.com/jaegertracing/jaeger-client-java/pull/416 
* https://github.com/jaegertracing/jaeger-client-java/pull/417


pavolloffay:
> Are there only two tests affected by this change? O_O

Internal use of deprecated API has been removed in previous PRs - just to keep this clean

jpkrohling:
> just to keep this clean

Ok, I personally would have preferred to keep this here, so that the actual effects of this PR could be easily measured.

pavolloffay:
@yurishkuro could you please have a look?

jaeger-core/src/main/java/io/jaegertracing/reporters/RemoteReporter.java
```diff
@@ -46,17 +46,6 @@
...
-   * At this point in time the protocol (thrift) is tightly-coupled to  Reporter, but in the future it may not be.
-   *
-   * @deprecated use {@link RemoteReporter.Builder} with fluent API
-   **/
-  @Deprecated
-  public RemoteReporter(Sender sender, int flushInterval, int maxQueueSize, Metrics metrics) {
-    this(sender, flushInterval, maxQueueSize, DEFAULT_CLOSE_ENQUEUE_TIMEOUT_MILLIS, metrics);
-  }
-
   RemoteReporter(Sender sender, int flushInterval, int maxQueueSize, int closeEnqueueTimeout,
```
yurishkuro: should we remove this ctor as well? Having public ctors is what leads to proliferation of deprecated methods in the first place, it's better to use builder pattern for most core components. Builder can be overkill for smaller classes, but RemoteReporter is a critical component that can continue evolving.

E.g. I see that we already use builder for RemoteSampler.

jaeger-core/src/main/java/io/jaegertracing/senders/HttpSender.java
```diff
@@ -43,47 +43,12 @@
    * http://localhost:14268/api/traces
    *
    * Uses the default {@link okhttp3.OkHttpClient} which uses {@link okhttp3.ConnectionPool#ConnectionPool()}.
-   * Use {@link HttpSender#HttpSender(java.lang.String, int, okhttp3.OkHttpClient)} to adjust parameters.
+   * Use {@link HttpSender.Builder} if you need to add more parameters
    */
   public HttpSender(String endpoint) {
```
yurishkuro: any reason to keep this ctor?

jaeger-core/src/main/java/io/jaegertracing/reporters/RemoteReporter.java
pavolloffay: It is package private, so not exposed to the world.

I can make it private in this PR.

jaeger-core/src/main/java/io/jaegertracing/senders/HttpSender.java
pavolloffay: Not really, It would be better to remove it to make the API more consistent. We can do 0.28.0 with more deprecations ASAP, then remove it

pavolloffay:
@yurishkuro I will merge this and address other comments in a separate PR as it's not related to removing deprecated APIs.

PR #417, opened 2018-05-16
Do not use deprecated Tracer.Builder constructor internally

pavolloffay:
Related to #413 and  #395 

This does not change any APIs or implementation. The change is only in test classes - removed invocation of deprecated `Tracer.Builder(name, reporter, sampler)`


pavolloffay:
I will merge to be able to continue in #414.

PR #438, opened 2018-06-05
Add baggage to B3 codec

pavolloffay:
Related to https://github.com/jaegertracing/jaeger/issues/755#issuecomment-393929718

* by default uses baggage- prefix
* it's enabled by default

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/propagation/B3TextMapCodec.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/propagation/B3TextMapCodec.java
@@ -15,7 +15,10 @@
 package io.jaegertracing.propagation;
 
 import io.jaegertracing.SpanContext;
+import io.jaegertracing.Tracer;
+import io.jaegertracing.baggage.BaggageRestrictionManager;
 import io.opentracing.propagation.TextMap;
+import java.util.HashMap;
 import java.util.Map;
 
 /**
@@ -34,6 +37,11 @@ import java.util.Map;
  *
  * <p>
  * See <a href="http://zipkin.io/pages/instrumenting.html">Instrumenting a Library</a>
+ *
+ * Note that this codec automatically propagates baggage
+ * (with {@value io.jaegertracing.propagation.B3TextMapCodec#BAGGAGE_NAME} prefix).
+ * Baggage whitelisting can be configured in {@link BaggageRestrictionManager} and then
+ * passed to {@link Tracer.Builder#baggageRestrictionManager}
  */
 public class B3TextMapCodec implements Codec<TextMap> {
   protected static final String TRACE_ID_NAME = "X-B3-TraceId";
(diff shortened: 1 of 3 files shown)


pavolloffay:
we should decide whether baggage will be enabled by default or not. However our baggage management is different and separate from the codec. Which makes me think that we can remove `enableBaggage` from the codec and just use baggage manager. Note that baggage manager applies for all installed codecs. But doing this we are forced to by default enable baggage in B3. On the other hand it will be consistent with our codec.

Brave/Sleuth has to be configured explicitly which keys are whitelisted for propagation.

yurishkuro:
+1 to integrate with baggage manager.

pavolloffay:
Done in the last commit

pavolloffay:
@yurishkuro could you please review?

jaeger-core/src/main/java/io/jaegertracing/propagation/B3TextMapCodec.java
```diff
@@ -34,17 +37,34 @@
...
   protected static final String PARENT_SPAN_ID_NAME = "X-B3-ParentSpanId";
   protected static final String SAMPLED_NAME = "X-B3-Sampled";
   protected static final String FLAGS_NAME = "X-B3-Flags";
+  protected static final String BAGGAGE_NAME = "baggage-";
   // NOTE: uber's flags aren't the same as B3/Finagle ones
   protected static final byte SAMPLED_FLAG = 1;
   protected static final byte DEBUG_FLAG = 2;
 
+  private static final PrefixedKeys keys = new PrefixedKeys();
+  private final String baggageName;
```
black-adder: ditto call it prefix instead?

jaeger-core/src/main/java/io/jaegertracing/propagation/B3TextMapCodec.java
```diff
@@ -80,12 +104,37 @@ public SpanContext extract(TextMap carrier) {
         if (entry.getValue().equals("1")) {
           flags |= DEBUG_FLAG;
         }
+      } else if (entry.getKey().startsWith(BAGGAGE_NAME)) {
```
black-adder: should this be `this.baggageName`?

jaeger-core/src/main/java/io/jaegertracing/propagation/B3TextMapCodec.java
```diff
@@ -80,12 +104,37 @@ public SpanContext extract(TextMap carrier) {
...
+        if (baggage == null) {
+          baggage = new HashMap<String, String>();
+        }
+        baggage.put(keys.unprefixedKey(entry.getKey(), BAGGAGE_NAME), entry.getValue());
```
black-adder: ditto

pavolloffay:
@black-adder PR updated.

jaeger-core/src/main/java/io/jaegertracing/propagation/B3TextMapCodec.java
```diff
@@ -80,12 +104,37 @@ public SpanContext extract(TextMap carrier) {
...
+        }
+        baggage.put(keys.unprefixedKey(entry.getKey(), baggagePrefix), entry.getValue());
       }
     }
 
     if (null != traceId && null != parentId && null != spanId) {
-      return new SpanContext(traceId, spanId, parentId, flags);
+      SpanContext spanContext = new SpanContext(traceId, spanId, parentId, flags);
+      if (baggage != null) {
+        spanContext = spanContext.withBaggage(baggage);
```
black-adder: this is an interesting use case, what if only baggage is propagated without the rest of the trace context?

pavolloffay: At the moment it's an invalid span context - null

PR #445, opened 2018-06-08
Make flush method in RemoteReporter public

### LINKED ISSUE #444: Make flush method in RemoteReporter public
Sometimes it's useful to force the flush explicitly. My use case is mostly related to tests purposes when I need to flush and then query.

pavolloffay:
Resolves #444 

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/reporters/RemoteReporter.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/reporters/RemoteReporter.java
@@ -106,7 +106,7 @@ public class RemoteReporter implements Reporter {
     }
   }
 
-  void flush() {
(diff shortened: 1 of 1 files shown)


vprithvi:
Commented on #444

jaeger-core/src/main/java/io/jaegertracing/reporters/RemoteReporter.java
```diff
@@ -106,7 +106,7 @@ public void close() {
     }
   }
 
-  void flush() {
+  public void flush() {
```
jpkrohling: As this is now public, add a bit of javadoc: is it blocking? Can it fail? If it fails, what happens to the spans on the buffer?

PR #449, opened 2018-06-13
Remove dependency from jaeger-core to jaeger-thrift

### LINKED ISSUE #448: Remove dependency from jaeger-core to jaeger-thrift
Currently, the `jaeger-core` module depends on `jaeger-thrift`. With possible new transports coming as part of #407, it would be beneficial to have the `jaeger-core` not to depend on `jaeger-thrift`.

jpkrohling:
Resolves #448 - This PR removes the dependency from jaeger-core to jaeger-thrift by loading
the appropriate Sender via either environment variables or based on existing service providers
on the classpath, loaded via ServiceLoader.


jaeger-core/src/main/java/io/jaegertracing/Configuration.java
```diff
@@ -137,6 +136,11 @@
    */
   public static final String JAEGER_PROPAGATION = JAEGER_PREFIX + "PROPAGATION";
 
+  /**
+   * The fully qualified class name for the sender factory
```
objectiser: Just wondering if we could select the factory based on a simpler name, which may be maps onto a predefined class name - so the configuration by the user requires less implementation knowledge?

jaeger-core/src/main/java/io/jaegertracing/senders/SenderResolver.java
```diff
@@ -0,0 +1,90 @@
...
+            + " set one by passing the fully qualified class name via JAEGER_SENDER_FACTORY");
+      }
+
+      log.info(String.format("Found a sender factory: %s", senderFactory));
+      Sender sender = senderFactory.getSender(senderConfiguration);
+      log.info(String.format("Using sender %s", sender));
+      return sender;
+    }
+
+    return new NoopSender();
```
objectiser: Should also log a warning that no sender has been found.

jaeger-thrift/build.gradle
```diff
@@ -43,6 +48,11 @@ shadowJar {
     classifier 'thrift92'
 }
 
+shadowJar {
```
objectiser: Don't we just want the single shaded jar, with the dependencies relocated?

jaeger-thrift/src/test/java/io/jaegertracing/thrift/senders/ThriftSenderFactoryTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+  @Test
+  public void testSenderWithAllPropertiesReturnsHttpSender() {
+    System.setProperty(Configuration.JAEGER_ENDPOINT, "https://jaeger-collector:14268/api/traces");
+    System.setProperty(Configuration.JAEGER_AGENT_HOST, "jaeger-agent");
+    System.setProperty(Configuration.JAEGER_AGENT_PORT, "6832");
+
+    assertTrue(Configuration.SenderConfiguration.fromEnv().getSender() instanceof HttpSender);
+  }
+
```
objectiser: Shouldn't there be some tests for getting the UdpSender?

jaeger-zipkin/build.gradle
```diff
@@ -8,6 +8,7 @@ dependencies {
     compile group: 'io.zipkin.reporter2', name: 'zipkin-reporter', version: '2.6.0'
 
     compile project(path: ':jaeger-core', configuration: 'shadow')
```
objectiser: Is a shadow configuration still required here? It might need to be on the jaeger-thrift dependency?

jaeger-core/src/main/java/io/jaegertracing/Configuration.java
jpkrohling: This would be required only if there are multiple factories in the classpath. Do you think it would be a common use case to justify a map of "common" factories?

jaeger-thrift/build.gradle
jpkrohling: Should I just remove the one from jaeger-core and have then `jaeger-thrift-okhttp381`?

http://central.maven.org/maven2/io/jaegertracing/jaeger-core/0.28.0/

jpkrohling: Sorry, I think I get a better understanding now. Did you mean to have the shadowed okhttp on the main artifact as well? Or did you mean to have the optional okhttp shadow artifact for thrift, instead of for jaeger-core?

jpkrohling: I just changed this PR to include both shadowed thrift and okhttp on the same "shadow" artifact

jaeger-core/src/test/java/io/jaegertracing/senders/InMemorySender.java
```diff
@@ -34,33 +32,32 @@
   private Semaphore semaphore = new Semaphore(Integer.MAX_VALUE);
 
   public InMemorySender() {
```
yurishkuro: we used to have inMemoryReporter. What's the point of InMemorySender?

jaeger-core/src/test/java/io/jaegertracing/senders/InMemorySender.java
```diff
@@ -34,33 +32,32 @@
...
-    flushed = new ArrayList<Span>();
-    received = new ArrayList<Span>();
+    appended = new ArrayList<>();
+    flushed = new ArrayList<>();
+    received = new ArrayList<>();
   }
 
   public List<Span> getAppended() {
-    return new ArrayList<Span>(appended);
+    return new ArrayList<>(appended);
```
yurishkuro: what are the threading expectations?

jaeger-thrift/build.gradle
```diff
@@ -6,9 +6,14 @@ apply plugin: 'com.github.johnrengelman.shadow'
 description = 'Generated thrift code'
```
yurishkuro: ?

jaeger-core/src/test/java/io/jaegertracing/senders/InMemorySender.java
jpkrohling: The `InMemorySender` was in the `reporters` package. I moved it to this package, as it seems more appropriate.

jaeger-core/src/test/java/io/jaegertracing/senders/InMemorySender.java
jpkrohling: Given that it's used only for tests, I would say that we don't guarantee anything. Concurrent modification *will* cause a `ConcurrentModificationException`, for instance.

jaeger-core/src/test/java/io/jaegertracing/senders/InMemorySender.java
yurishkuro: ok, just for tests - nvm

jaeger-core/src/main/java/io/jaegertracing/Configuration.java
objectiser: Not necessary common, but just thinking of the detail the user would need to know (i.e. the fully qualified class name) to select a transport.

jaeger-core/src/main/java/io/jaegertracing/senders/SenderFactory.java
```diff
@@ -0,0 +1,30 @@
...
+ * Represents a class that knows how to select and build the appropriate {@link Sender} based on the given
+ * {@link Configuration.SenderConfiguration}
+ */
+public interface SenderFactory {
+  /**
+   * Builds and/or selects the appropriate sender based on the given {@link Configuration.SenderConfiguration}
+   * @param senderConfiguration the sender configuration
+   * @return an appropriate sender based on the configuration, or {@link NoopSender}.
+   */
+  Sender getSender(Configuration.SenderConfiguration senderConfiguration);
```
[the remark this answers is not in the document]
jpkrohling: Agree. This would require a bigger refactor than I'd like to focus on at this point. For consistency's sake, other configuration objects would also need such a change :/

jaeger-core/src/main/java/io/jaegertracing/Configuration.java
jpkrohling: In the usual case, they just need to have the transport on the classpath, there's no need to know the class name. If they end up somehow with two transports, the class name is used to disambiguate. 

How about we go without this map for now, and add later if the case is more common than we thought?

objectiser: Two other suggestions - (1) we treat multiple factories as an error - so if the case comes up to justify having two in the classpath, we can explore it further then, (2) add a type property on the factory, and that is what is specified in the environment variable?

jpkrohling: > add a type property on the factory, and that is what is specified in the environment variable?

I like this idea, as it's easily extensible.

jaeger-core/src/main/java/io/jaegertracing/senders/SenderFactory.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Represents a class that knows how to select and build the appropriate {@link Sender} based on the given
+ * {@link Configuration.SenderConfiguration}
+ */
+public interface SenderFactory {
+  /**
+   * Builds and/or selects the appropriate sender based on the given {@link Configuration.SenderConfiguration}
+   * @param senderConfiguration the sender configuration
+   * @return an appropriate sender based on the configuration, or {@link NoopSender}.
```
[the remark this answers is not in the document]
jpkrohling: Not really -- I expect concrete `SenderFactory` implementations to *never* return null. The contract is: either a real `Sender`, or `NoopSender`.

jaeger-core/src/main/java/io/jaegertracing/senders/SenderFactory.java
```diff
@@ -0,0 +1,30 @@
...
+ * the License.
+ */
+
+package io.jaegertracing.senders;
+
+import io.jaegertracing.Configuration;
+
+/**
+ * Represents a class that knows how to select and build the appropriate {@link Sender} based on the given
+ * {@link Configuration.SenderConfiguration}
```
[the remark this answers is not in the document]
jpkrohling: The resolver loads this via the service loader, but there's nothing preventing a consumer from calling a concrete implementation directly. Like: `SenderFactory f = new InMemorySenderFactory()`

jaeger-core/src/main/java/io/jaegertracing/senders/SenderResolver.java
```diff
@@ -0,0 +1,91 @@
...
+
+package io.jaegertracing.senders;
+
+import io.jaegertracing.Configuration;
+import java.util.Iterator;
+import java.util.ServiceLoader;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * Provides a way to resolve an appropriate {@link Sender}
```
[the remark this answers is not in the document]
jpkrohling: This is implementation-specific. The `ThriftSender`, for instance, builds an `UdpSender` or `HttpSender` based on the contents of the `SenderConfiguration`. More information is provided at the method-level javadoc.

jaeger-core/src/main/java/io/jaegertracing/senders/NoopSender.java
```diff
@@ -0,0 +1,40 @@
...
+@ToString
+public class NoopSender implements Sender {
+  @Override
+  public int append(Span span) {
+    return 1;
+  }
+
+  @Override
+  public int flush() {
+    return 1;
```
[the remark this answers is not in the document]
jpkrohling: Thought so too. But not for the `append(Span)`, right?

jaeger-core/src/main/java/io/jaegertracing/senders/SenderFactory.java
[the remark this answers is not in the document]
pavolloffay: What can't we return null? It's invoked only in `SenderFactory` which takes care about null/noop.

jaeger-core/src/main/java/io/jaegertracing/senders/SenderResolver.java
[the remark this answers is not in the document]
pavolloffay: Then how consumers of this API know how to provide their own sender?

jaeger-core/src/main/java/io/jaegertracing/senders/SenderFactory.java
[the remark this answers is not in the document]
jpkrohling: This is a contract between the implementation and the consumer, meaning that this documentation is a reference to both sides. By stating that it might return NoopSender, we are ruling out a null object, which makes it far easier for consuming code to not break (no null checks, no null pointer exceptions).

jaeger-core/src/main/java/io/jaegertracing/senders/SenderResolver.java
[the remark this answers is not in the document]
jpkrohling: Consumer of the API != Service Provider.

The Service Provider will know how to instantiate a new Sender and will return it as a result of this method call (see jaeger-thrift).

The consumer of the API (target application) will, at most, set the appropriate configuration options on `SenderConfiguration` and will get an appropriate sender in return.

jaeger-core/src/main/java/io/jaegertracing/Configuration.java
jpkrohling: > I think FQN is better than just a class name.

This is also available: if the env var contains a class name, we try to load it. Otherwise, we try to get a single factory from the service loader. If there are multiple, we try to match the type with the env var. If none matches, NoopSender is returned.

jaeger-core/src/main/java/io/jaegertracing/senders/SenderFactory.java
```diff
@@ -0,0 +1,37 @@
...
+public interface SenderFactory {
+  /**
+   * Builds and/or selects the appropriate sender based on the given {@link Configuration.SenderConfiguration}
+   * @param senderConfiguration the sender configuration
+   * @return an appropriate sender based on the configuration, or {@link NoopSender}.
+   */
+  Sender getSender(Configuration.SenderConfiguration senderConfiguration);
+
+  /**
+   * The Factory's name. Can specified via {@link Configuration#JAEGER_SENDER_FACTORY} to disambiguate
```
objectiser: nit: "Can be specified"

jaeger-core/src/main/java/io/jaegertracing/senders/SenderResolver.java
```diff
@@ -0,0 +1,113 @@
...
+   * JAEGER_SENDER_FACTORY or, in its absence or failure to deliver a {@link Sender}, via the {@link ServiceLoader}.
+   * If no factories are found, a {@link NoopSender} is returned. If multiple factories are available, the
+   * factory whose {@link SenderFactory#getType()} matches the JAEGER_SENDER_FACTORY env var is selected.
+   * If none matches, {@link NoopSender} is returned.
+   *
+   * @param senderConfiguration the configuration to pass down to the factory
+   * @return the resolved Sender, or NoopSender
+   */
+  public static Sender resolve(Configuration.SenderConfiguration senderConfiguration) {
+    String senderFactoryClass = System.getProperty(Configuration.JAEGER_SENDER_FACTORY);
```
objectiser: nit: `senderFactoryClass` -> `senderFactoryType` ?

objectiser: Isn't this still using the old mechanism - where the env var specifies a fq class name?

jpkrohling: It can be both. At this point in the code, we assume it's a class (we try to use it on a `Class.forName` later on).

jpkrohling: > Isn't this still using the old mechanism - where the env var specifies a fq class name?

Yes. I thought the type is only used to disambiguate in case there are multiple factories available? The previous resolution was:

- FQCN
- Service Loader (first service available)
- NoopSender

The current is:

- FQCN
- Service Loader (when only one service is available)
- Service Loader + Type (when multiple services are available)
- NoopSender

jaeger-thrift/build.gradle
```diff
@@ -40,7 +43,8 @@ checkstyleTest.enabled = false
 shadowJar {
     baseName = 'jaeger-thrift'
     relocate 'org.apache.thrift', 'org.shadow.apache.thrift92'
-    classifier 'thrift92'
+    relocate 'okhttp', 'jaeger.okhttp'
```
objectiser: nit: we should maybe have a consistent relocate pattern - so possibly prefix jaeger as here, so update the thrift one to `jaeger.org.apache.thrift` (without the 92)?
btw - might also be worth looking at the other dependencies included in this shaded jar, as they may be should be relocated aswell to avoid any potential clashes  - or possibly this should be a separate PR.

jpkrohling: > or possibly this should be a separate PR

That would be my preference.

jpkrohling: https://github.com/jaegertracing/jaeger-client-java/issues/450

jaeger-core/src/main/java/io/jaegertracing/senders/SenderFactory.java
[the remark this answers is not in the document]
pavolloffay: I still think the Noop here is just redundant. This API does not need it and it's just creating a dependency on specific sender impl

jaeger-core/src/main/java/io/jaegertracing/senders/SenderResolver.java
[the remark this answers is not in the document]
pavolloffay: I have my own sender impl. How do I use jaeger-core to load it from classpath? Where can I found more info about it? I just want to use `Configuration.fromEnv()`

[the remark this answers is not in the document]
jpkrohling: You implement the `SenderFactory`, which returns your own `Sender` on the `resolve()` method. See the unit tests and the jaeger-thrift for an example on how to do that.

jaeger-core/src/main/java/io/jaegertracing/senders/SenderResolver.java
```diff
@@ -0,0 +1,113 @@
...
+   * @return the resolved Sender, or NoopSender
+   */
+  public static Sender resolve() {
+    return resolve(Configuration.SenderConfiguration.fromEnv());
+  }
+
+  /**
+   * Resolves a sender by passing the given {@link Configuration.SenderConfiguration} down to the
+   * {@link SenderFactory}. The factory is loaded either based on the value from the environment variable
+   * JAEGER_SENDER_FACTORY or, in its absence or failure to deliver a {@link Sender}, via the {@link ServiceLoader}.
```
pavolloffay: JAEGER_SENDER_FACTORY make it a javadoc link.

jaeger-core/src/main/java/io/jaegertracing/senders/SenderResolver.java
pavolloffay: Type in `SenderFactory` seems redundant as you can still specify the wanted impl as FQCN or wouldn't it work?

jaeger-core/src/main/java/io/jaegertracing/senders/SenderResolver.java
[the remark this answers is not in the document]
pavolloffay: I implement `SenderFactory` in my app, add `jaeger-core`. then I use `Configuration.fromEnv()` without specifying any env var. Would be my sender factory used?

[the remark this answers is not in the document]
jpkrohling: Yes. The only thing missing in your example is the META-INF/services/io.jaegertracing.senders.SenderFactory . 

https://github.com/jpkrohling/jaeger-client-java/blob/6a83cd7c56877bc17aeb8f8fd432a00a2e2e99b7/jaeger-thrift/src/main/resources/META-INF/services/io.jaegertracing.senders.SenderFactory

jaeger-core/src/main/java/io/jaegertracing/senders/SenderResolver.java
jpkrohling: It would, and it's how it was being done before. Looks at this thread: https://github.com/jaegertracing/jaeger-client-java/pull/449#discussion_r195080992

Issue #406, opened 2018-04-26
HttpSender errors are not logged


black-adder:
If the http sender fails due to unauthorized access, we throw an exception: https://github.com/jaegertracing/jaeger-client-java/blob/2fa8f3ed17e88114f6cd7114747ed298933dabd8/jaeger-core/src/main/java/io/jaegertracing/senders/HttpSender.java#L120

However, this exception doesn't seem to be logged anywhere although we do emit a metric:
https://github.com/jaegertracing/jaeger-client-java/blob/2fa8f3ed17e88114f6cd7114747ed298933dabd8/jaeger-core/src/main/java/io/jaegertracing/reporters/RemoteReporter.java#L184

Either making the metric more explicit or logging the error will be apt here.

chandresh-pancholi:
Let me know what log statement you want me to add. 
I will include in the HttpSender.java

chandresh-pancholi:
@black-adder  Below code is throwing SenderException. It is not specific to unauthorized access though.
`

      String exceptionMessage = String.format("Could not send %d spans, response %d: %s",
          spans.size(), response.code(), responseBody);
      throw new SenderException(exceptionMessage, null, spans.size());

`

black-adder:
That's fine, we should log any exception. In line https://github.com/jaegertracing/jaeger-client-java/blob/2fa8f3ed17e88114f6cd7114747ed298933dabd8/jaeger-core/src/main/java/io/jaegertracing/reporters/RemoteReporter.java#L184 once we get the exception, we should log it, `log.error("QueueProcessor sender error:", e);`

chandresh-pancholi:
Is it right to log complete stack trace? Just the log message won't work?

jpkrohling:
Do pass the exception to `log.error`. The end user can configure the logger to suppress the stack trace if needed. Also, it would probably better to wait for #449 before doing this one, otherwise you might get a merge conflict.

chandresh-pancholi:
@jpkrohling  Sure, I will wait for #449 to close.

esukram:
Hi @jpkrohling 
As there has been some work already, do still qualify that to be done?

jpkrohling:
@chandresh-pancholi, are you still interested in working on this one? 

@esukram feel free to take it in a couple of days, in case @chandresh-pancholi isn't available or isn't interested. You might even be able to get two issues in one, as #653 is somewhat related to this one here.

[...]

PR #452, opened 2018-06-15
HttpSender errors are not logged #406

chandresh-pancholi:
Signed-off-by: chandresh-pancholi <chandreshpancholi007@gmail.com>

- HttpSender errors are not logged #406 

- Adding error log when the response  is not successful and can not read it

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/senders/HttpSender.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/senders/HttpSender.java
@@ -21,6 +21,7 @@ import io.jaegertracing.thriftjava.Span;
 import java.io.IOException;
 import java.util.List;
 import lombok.ToString;
+import lombok.extern.slf4j.Slf4j;
 import okhttp3.Credentials;
 import okhttp3.HttpUrl;
 import okhttp3.Interceptor;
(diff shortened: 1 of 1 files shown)


jaeger-core/src/main/java/io/jaegertracing/senders/HttpSender.java
```diff
@@ -31,6 +32,7 @@
 import okhttp3.Response;
 
 @ToString(exclude = {"httpClient", "requestBuilder"})
+@Slf4j
 public class HttpSender extends ThriftSender {
```
jpkrohling: Looks like your branch was still using an old master. As part of #449, this has been moved to `jaeger-thrift/src/main/java/io/jaegertracing/thrift/senders/HttpSender.java`

Please, rebase this to the latest master.

jaeger-core/src/main/java/io/jaegertracing/senders/HttpSender.java
```diff
@@ -74,6 +76,8 @@ public void send(Process process, List<Span> spans) throws SenderException {
         responseBody = response.body() != null ? response.body().string() : "null";
       } catch (IOException e) {
         responseBody = "unable to read response";
```
jpkrohling: How about something like this:

```java
responseBody = String.format("unable to read response: %s", e.getMessage());
log.error(responseBody, e);
```

PR #456, opened 2018-06-19
Upgrades internals to Zipkin v2 library

adriancole:
The current zipkin integration encourages use of a dead version of a library (io.zipkin.reporter)

Recent version of io.zipkin.zipkin2:zipkin has been backported to
support thrift encoding. This updates the corresponding sender here to
use that, notably to migrate off the io.zipkin.reporter libraries which
receive no updates.

### DIFF
--- a/jaeger-zipkin/build.gradle
+++ b/jaeger-zipkin/build.gradle
@@ -1,18 +1,12 @@
 description = 'Integration library for Zipkin'
 
 dependencies {
-    compile group: 'io.zipkin.reporter', name: 'zipkin-reporter', version: '1.1.2'
-    compile group: 'io.zipkin.reporter', name: 'zipkin-sender-urlconnection', version: '1.1.2'
-
-    compile group: 'io.zipkin.java', name: 'zipkin', version: '2.8.1'
-    compile group: 'io.zipkin.reporter2', name: 'zipkin-reporter', version: '2.6.0'
+    compile group: 'io.zipkin.reporter2', name: 'zipkin-sender-urlconnection', version: '2.7.6'
 
     compile project(':jaeger-core')
     compile project(':jaeger-thrift')
 
-    testCompile group: 'io.zipkin.reporter2', name: 'zipkin-sender-urlconnection', version: '2.6.0'
-    testCompile group: 'junit', name: 'junit', version: junitVersion
-    testCompile group: 'io.zipkin.java', name: 'zipkin-junit', version: '2.7.1'
+    testCompile group: 'io.zipkin.zipkin2', name: 'zipkin-junit', version: '2.9.4'
     testCompile group: 'com.tngtech.java', name: 'junit-dataprovider', version: junitDataProviderVersion
 
     signature 'org.codehaus.mojo.signature:java16:1.1@signature'
--- a/jaeger-zipkin/src/main/java/io/jaegertracing/senders/zipkin/ZipkinSender.java
+++ b/jaeger-zipkin/src/main/java/io/jaegertracing/senders/zipkin/ZipkinSender.java
@@ -14,6 +14,11 @@
 
 package io.jaegertracing.senders.zipkin;
 
+import static com.twitter.zipkin.thriftjava.zipkincoreConstants.CLIENT_RECV;
+import static com.twitter.zipkin.thriftjava.zipkincoreConstants.CLIENT_SEND;
+import static com.twitter.zipkin.thriftjava.zipkincoreConstants.SERVER_RECV;
+import static com.twitter.zipkin.thriftjava.zipkincoreConstants.SERVER_SEND;
+
 import com.twitter.zipkin.thriftjava.Annotation;
 import com.twitter.zipkin.thriftjava.BinaryAnnotation;
 import com.twitter.zipkin.thriftjava.Endpoint;
@@ -23,13 +28,14 @@ import io.jaegertracing.exceptions.SenderException;
 import io.jaegertracing.senders.Sender;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
+import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Set;
 import lombok.ToString;
-import zipkin.Constants;
-import zipkin.reporter.Encoding;
-import zipkin.reporter.internal.AwaitableCallback;
-import zipkin.reporter.urlconnection.URLConnectionSender;
+import zipkin2.codec.Encoding;
+import zipkin2.reporter.urlconnection.URLConnectionSender;
 
 /**
  * This sends (TBinaryProtocol big-endian) encoded spans to a Zipkin Collector (usually a
@@ -48,8 +54,10 @@ import zipkin.reporter.urlconnection.URLConnectionSender;
  * <p>
  * See https://github.com/openzipkin/zipkin/tree/master/zipkin-server
  */
-@ToString(exclude = "spanBuffer")
+@ToString(exclude = {"spanBuffer", "encoder"})
 public final class ZipkinSender implements Sender {
+  static final Set<String> CORE_ANNOTATIONS = new LinkedHashSet<String>(
+      Arrays.asList(CLIENT_SEND, CLIENT_RECV, SERVER_SEND, SERVER_RECV));
 
   /**
    * @param endpoint The POST URL for zipkin's <a href="http://zipkin.io/zipkin-api/#/">v1 api</a>,
@@ -57,7 +65,8 @@ public final class ZipkinSender implements Sender {
    * @return new sender
    */
   public static ZipkinSender create(String endpoint) {
-    return new ZipkinSender(URLConnectionSender.create(endpoint));
+    return new ZipkinSender(
+        URLConnectionSender.newBuilder().encoding(Encoding.THRIFT).endpoint(endpoint).build());
   }
 
   /**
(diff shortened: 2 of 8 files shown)


adriancole:
ps I plan to add a test for the v2 path, but only after figuring out the following:

1. would this be mergable in general?
2. since the version is <1.0 can we remove the old factory method now? If so, there will be less test work to do to achieve the same coverage.

adriancole:
The main change here is the library that sends the messages. both the new
and old factory method assume thrift. I think there is different code in
the same package which can use non-thrift. That hasnt changed as that side
is already using the zipkin2 reporter library.

on zipkin server side, v1 POST with thrift is available for the foreseeable
future and the kafka transport also still accepts thrift.

adriancole:
ok tests now use both paths.

I'm keen either way, retaining the deprecated factory method for a release, or cutting it now. If we do the former, we need to remember to remove it later.

pavolloffay:
As there is `ZipkinV2Reporter` (which can be used to sent thrift) can't we just deprecate the entire sender? Now we have to maintain two converters and sender only for thrift.

adriancole:
my goal is to drop the dependency on io.zipkin.java

dropping the thrift sender achieves that as long as dropping it doesn't
limit folks in some way I'm not aware of.

adriancole:
note even if you do drop the sender (now or later) the dependencies in
this change are cleaner, especially the tests. Please retain that part
if you decide to drop the sender.

jaeger-zipkin/src/main/java/io/jaegertracing/senders/zipkin/ZipkinSender.java
```diff
@@ -50,39 +59,58 @@
...
    *
    * <p>
-   * Ex. for Kafka ("io.zipkin.reporter:zipkin-sender-kafka08")
+   * Ex. for Kafka ("io.zipkin2.reporter:zipkin-sender-kafka11")
    *
    * <pre>{@code
-   * sender = ZipkinSender.create(KafkaSender.create("192.168.99.100:9092"));
+   * sender = ZipkinSender.create(
+   *     KafkaSender.newBuilder()
+   *                .encoding(Encoding.THRIFT)
```
pavolloffay: What happens if the encoding != thrift?

pavolloffay: Cool there is a check. nevermind

adriancole:
@yurishkuro etc I think the least tech debt way here is to remove the deprecated factory method which then removes a large part of this code which will restore coverage to where it was. This also allows a smooth transition for those who can't immediately switch off thrift (for example)

if you say go, I'll do that and hopefully we can get this merged.

yurishkuro:
@adriancole I have no problem with removing the factory, my question was about functional impact. I haven't followed zipkin v2 too closely, was it a change to JSON schema only, or did thrift model changed as well?  If the thrift model is the same (https://github.com/openzipkin/zipkin-api/blob/master/thrift/zipkinCore.thrift - this one?), then upgrading the library to send it makes total sense.

adriancole:
thrift is the same. we backported the encoding logic, so basically it takes
a v2 model and allocates it across a v1 (aka only) thrift struct when
writing.

will yank the old path which will be nicer I think in general.

adriancole:
ok ready to roll. I can squash or you can

jaeger-zipkin/src/main/java/io/jaegertracing/zipkin/V2SpanConverter.java
```diff
@@ -122,14 +119,13 @@ private static void buildTags(Span span, zipkin2.Span.Builder builder) {
     }
   }
 
-  private static InetAddress convertIp(int ip) {
```
adriancole: I fixed this as it was bothering me

adriancole:
sorry i answered wrong question. this change is about senders. no impact as
they take byte arrays and dont mess with them besides bundling

yurishkuro:
>  I can squash or you can

Our (CNCF's) contributing policy requires signed commits (the DCO check), so it's best if you squash them and sign as a single commit (git commit -s ...)

adriancole:
Signed with a Flourish

adriancole:
hopefully I've now addressed the severely annoying test coverage requirements

jaeger-zipkin/build.gradle
```diff
@@ -1,18 +1,15 @@
 description = 'Integration library for Zipkin'
 
 dependencies {
-    compile group: 'io.zipkin.reporter', name: 'zipkin-reporter', version: '1.1.2'
-    compile group: 'io.zipkin.reporter', name: 'zipkin-sender-urlconnection', version: '1.1.2'
+    implementation "io.zipkin.reporter2:zipkin-reporter-bom:2.7.5"
 
-    compile group: 'io.zipkin.java', name: 'zipkin', version: '2.8.1'
-    compile group: 'io.zipkin.reporter2', name: 'zipkin-reporter', version: '2.6.0'
+    implementation 'io.zipkin.reporter2:zipkin-reporter'
```
pavolloffay: shouldn't this be `api` scope?

jaeger-zipkin/build.gradle
```diff
@@ -1,18 +1,15 @@
...
+    implementation 'io.zipkin.reporter2:zipkin-sender-urlconnection'
```
pavolloffay: This is new to me, it's related to https://github.com/jaegertracing/jaeger-client-java/pull/456/files#diff-88b7c47e47b8ee65263b6784b86fa0a7R1

I got this when I included it into a different project

> [WARNING] The POM for io.jaegertracing:jaeger-zipkin:jar:0.29.1-SNAPSHOT is invalid, transitive dependencies (if any) will not be available, enable debug logging for more details


https://docs.gradle.org/4.6/userguide/java_library_plugin.html#sec:java_library_separation

> Dependencies appearing in the api configurations will be transitively exposed to consumers of the library, and as such will appear on the compile classpath of consumers. Dependencies found in the implementation configuration will, on the other hand, not be exposed to consumers, and therefore not leak into the consumers' compile classpath. This comes with several benefits:

We basically require our clients to explicitly add dependency on `zipkin-sender-urlconnection` to make this work. If am correct it will fail on `java.lang.NoClassDefFoundError: zipkin2/reporter/urlconnection/URLConnectionSender` even if you use different sender.

adriancole:
hmm maybe they should both be api I guess. Will fix!

pavolloffay:
+1, or just include the reporter and let users specify sender they want. But it's a bigger breaking change. I guess url sender does not pull in anything special so it's fine to include it.

adriancole:
hmm

```
> Could not find method api() for arguments [io.zipkin.reporter2:zipkin-reporter] on object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.
```

adriancole:
looks like api can work if I upgrade to gradle 4.8, adding that..

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.8-bin.zip
```
jpkrohling: It would be awesome if you could update the version specified under the Wrapper task as well: https://github.com/jaegertracing/jaeger-client-java/blob/master/build.gradle#L40

jaeger-zipkin/src/main/java/io/jaegertracing/senders/zipkin/ZipkinSender.java
```diff
@@ -160,9 +175,14 @@ public int close() throws SenderException {
     return n;
   }
 
+  @Override
+  public String toString() {
+    return "ZipkinSender{" + delegate + "}";
```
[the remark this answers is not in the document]
jpkrohling: Looks like this PR removes `@ToString` and implements its own `toString()`. Did you have problems with the resulting `toString()` from `@ToString`?

jaeger-zipkin/src/test/java/io/jaegertracing/senders/zipkin/ZipkinSenderTest.java
```diff
@@ -52,7 +53,7 @@
   Tracer tracer;
 
   @Before
-  public void setUp() throws Exception {
```
jpkrohling: +1

adriancole:
with regards to lombok toString. it was maintenance causing as someone
earlier forgot to exclude a field. I mentioned this in am earlier comment
on this PR

jpkrohling:
Sorry, missed your previous comment about the `toString()`. I'm OK with that but would prefer to keep the `@ToString` for consistency's sake.

adriancole:
as thinga dont seem to be working like I hoped and I am spending more time than I should.. i will revert the bom thing amd repush the change. feel free to re-bom later.

adriancole:
I will also revert the lombok tostring thing as that tool deserves no more time from me!

adriancole:
unless there's more karmatic debt to pay in the form of build failures, I'm done I think.

jaeger-zipkin/src/main/java/io/jaegertracing/senders/zipkin/ZipkinSender.java
```diff
@@ -48,41 +54,52 @@
  * <p>
  * See https://github.com/openzipkin/zipkin/tree/master/zipkin-server
  */
-@ToString(exclude = "spanBuffer")
+@ToString(exclude = {"spanBuffer", "encoder"})
```
jpkrohling: \o/

(not that I'm a fan of Lombok, btw)

adriancole:
thanks for getting this merged

PR #457, opened 2018-06-19
Make GuaranteedThroughputSampler public

quaff:


jpkrohling:
@quaff, do you have a reason to require this to be public? Also, looks like your commit is missing the DCO sign-off.

jaeger-core/src/main/java/io/jaegertracing/samplers/GuaranteedThroughputSampler.java
```diff
@@ -31,15 +30,15 @@
  */
 @ToString
 @EqualsAndHashCode
-@AllArgsConstructor (access = AccessLevel.PACKAGE) // Visible for testing
+@AllArgsConstructor
```
jpkrohling: Just noticed this change here: this exposes a *new* constructor, to initialize the private fields (`probabilisticSampler`, `lowerBoundSampler` and `tags`). 

I don't think we want that. In fact, I think this annotation can be removed altogether.

quaff:
@jpkrohling commit updated, since ProbabilisticSampler and RateLimitingSampler are public, why not GuaranteedThroughputSampler?

jpkrohling:
@quaff Looks like a recent change caused merge conflicts with this PR. Would you please update it? Once it's rebased to the latest master, I'll merge it.

quaff:
@jpkrohling updated.

jpkrohling:
Thanks, @quaff!

Issue #459, opened 2018-06-20
InMemoryMetricsFactory.createGauge considers amount as delta


mdouaihy:
Hello,

I am using the Java Jaeger client in my project. I stumbled upon `InMemoryMetricsFactory` and noticed a bug in the `createGauge` method.

In fact `createGauge` considers the amount as a delta whereas it's an absolute amount.

**Current Behavior**
```
return new Gauge() {
      @Override
      public void update(long amount) {
        value.addAndGet(amount);
      }
    };
```
**Suggested Behavior**
```
return new Gauge() {
      @Override
      public void update(long amount) {
        value.getAndSet(amount);
      }
    };
```

jpkrohling:
Would you like to contribute a fix? If you need help in submitting your first contribution, let me know.

mdouaihy:
Pull request #463 was added for that purpose.

PR #463, opened 2018-06-21
Fix bug described in (#459)

mdouaihy:
### Which problem is this PR solving?
InMemoryMetricsFactory gauge wrongly update its value.
The behavior is described in (#459)

### Description of the change
In InMemoryMetricsFactory, upon updating the gauge, the argument
should be considered as the new value and not a variation to
the previous one.

Signed-off-by: Mehrez Douaihy <mehrez.douaihy@gmail.com>


yurishkuro:
Thank you for your contribution.

PR #467, opened 2018-06-27
After finish, disallow setTag.

nikron:
If you set the tag after the span finishes, this can cause a race condition in the sender as it tries to iterate over the tag map.

To fix that, you really shouldn't even be able to set a tag after the span finishes.
Check finish on setTagAsObject and return if the span is already finished.

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/JaegerSpan.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/JaegerSpan.java
@@ -99,6 +99,10 @@ public class JaegerSpan implements Span {
   @Override
   public JaegerSpan setOperationName(String operationName) {
     synchronized (this) {
+      if (spanAlreadyFinished("set operation name")) {
+        return this;
+      }
+
       this.operationName = operationName;
     }
     return this;
@@ -132,6 +136,10 @@ public class JaegerSpan implements Span {
       return this;
     }
     synchronized (this) {
+      if (spanAlreadyFinished("set baggage item")) {
+        return this;
+      }
+
       context = tracer.setBaggage(this, key, value);
       return this;
     }
@@ -176,10 +184,10 @@ public class JaegerSpan implements Span {
 
   private void finishWithDuration(long durationMicros) {
     synchronized (this) {
-      if (finished) {
-        log.warn("Span has already been finished; will not be reported again.");
+      if (spanAlreadyFinished("finish")) {
         return;
       }
+
       finished = true;
 
       this.durationMicroseconds = durationMicros;
@@ -206,6 +214,10 @@ public class JaegerSpan implements Span {
   }
 
   private JaegerSpan setTagAsObject(String key, Object value) {
+    if (spanAlreadyFinished("set tag")) {
+      return this;
+    }
+
     if (key.equals(Tags.SAMPLING_PRIORITY.getKey()) && (value instanceof Number)) {
       int priority = ((Number) value).intValue();
       byte newFlags;
@@ -233,6 +245,10 @@ public class JaegerSpan implements Span {
   @Override
   public JaegerSpan log(long timestampMicroseconds, Map<String, ?> fields) {
     synchronized (this) {
+      if (spanAlreadyFinished("log")) {
+        return this;
+      }
+
       if (fields == null) {
         return this;
       }
@@ -257,6 +273,10 @@ public class JaegerSpan implements Span {
   @Override
   public JaegerSpan log(long timestampMicroseconds, String event) {
     synchronized (this) {
+      if (spanAlreadyFinished("log")) {
+        return this;
+      }
+
       if (event == null) {
         return this;
       }
@@ -268,6 +288,18 @@ public class JaegerSpan implements Span {
       }
       return this;
     }
+  }
+
+  private boolean spanAlreadyFinished(String methodName) {
+    if (finished) {
+      log.warn("Span has already been finished; cannot {}.", methodName);
+      if (log.isDebugEnabled()) {
+        log.debug("Generating stack trace because we cannot {}.", methodName, new RuntimeException("Called from:"));
+      }
+      return true;
+    }
+
+    return false;
   }
 
   /**
--- a/jaeger-core/src/main/java/io/jaegertracing/JaegerTracer.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/JaegerTracer.java
@@ -226,7 +226,7 @@ public class JaegerTracer implements Tracer, Closeable {
     private final Map<String, Object> tags = new HashMap<String, Object>();
     private boolean ignoreActiveSpan = false;
 
-    SpanBuilder(String operationName) {
+    private SpanBuilder(String operationName) {
       this.operationName = operationName;
     }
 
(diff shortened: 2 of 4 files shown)


jaeger-core/src/main/java/io/jaegertracing/JaegerSpan.java
```diff
@@ -206,6 +206,11 @@ public synchronized JaegerSpan setTag(String key, Number value) {
   }
 
   private JaegerSpan setTagAsObject(String key, Object value) {
+    if (finished) {
+      log.warn("Span has already been finished; cannot set tag.", new RuntimeException("Called from:"));
```
jpkrohling: Do you have a sample output for this log message? I'm not sure creating a new exception here (even if not throwing it) is what we want...

yurishkuro: Exception seems useful, but dual-finish doesn't use it currently, so if we want to have exception, I'd say it should be a tracer config switch to enable it.

If we are making this change for tags, it should also apply for all methods that write to the span, i.e. setOperation, log(), etc.

nikron:
Before I actually test the logging (because it's sort of annoying to test that...) does this look good?

jaeger-core/src/main/java/io/jaegertracing/JaegerTracer.java
```diff
@@ -536,6 +540,11 @@ public Builder withExpandExceptionLogs() {
       return this;
     }
 
+    public Builder withLoggingStackTraces() {
```
yurishkuro: LoggingStackTraces name might be confused with some more serious feature. I would rename to `withLogStackTraceOnWarnings`.

Also add javadoc explaining the option.

jaeger-core/src/main/java/io/jaegertracing/JaegerSpan.java
```diff
@@ -257,6 +297,15 @@ public JaegerSpan log(String event) {
...
   public JaegerSpan log(long timestampMicroseconds, String event) {
     synchronized (this) {
+      if (finished) {
+        if (tracer.isLoggingStackTraces()) {
+          log.warn("Span has already been finished; cannot log.", new RuntimeException("Called from:"));
+        } else {
+          log.warn("Span has already been finished; cannot log.");
+        }
+        return this;
+      }
```
yurishkuro: I suggest to DRY it by moving this to a method `spanAlreadyFinished(String funcName)` and then replacing the checks with 
```
if (spanAlreadyFinished("setTag")) {
    return this;
}
```

jaeger-core/src/main/java/io/jaegertracing/JaegerSpan.java
jpkrohling: I do agree that the exception seems useful, but I'd rather have two log entries then. One at warning level, with a message like "Cannot set tag, as the span is already closed. Set log level to debug for information about the caller". Then, the second would be in debug level, wrapped in `if (log.isDebugEnabled()) { logger.debug(...); }`, with the stack trace.

My concerns are:

1. Pollute the logs with stack traces
1. Confuse users into thinking these stack traces are actual exceptions
1. Wasted resources in collecting the stack trace for every single time the span is closed, as this information is useful only once per caller.

yurishkuro:
@jpkrohling suggestion is way better & simpler than mine of doing a custom flag.

nikron:
Changed it to do two lines when debug is on.

jaeger-core/src/main/java/io/jaegertracing/JaegerSpan.java
```diff
@@ -99,6 +99,10 @@ public JaegerTracer getTracer() {
   @Override
   public JaegerSpan setOperationName(String operationName) {
     synchronized (this) {
+      if (spanAlreadyFinished("set baggage item")) {
```
yurishkuro: set operation name

jaeger-core/src/main/java/io/jaegertracing/JaegerSpan.java
```diff
@@ -132,6 +136,10 @@ public JaegerSpan setBaggageItem(String key, String value) {
       return this;
     }
     synchronized (this) {
+      if (spanAlreadyFinished("set baggage item")) {
```
yurishkuro: strictly speaking setting baggage would work even after the span is finished

nikron: It seems unlikely that the user would intend to do this, but do you want me to remove it?

jaeger-core/src/test/java/io/jaegertracing/JaegerSpanTest.java
```diff
@@ -282,6 +313,32 @@ public void testLog() {
...
+    when(clock.currentTimeMicros()).thenReturn(expectedTimestamp);
+
+    jaegerSpan.finish();
+    jaegerSpan.log(expectedEvent);
+    assertNull(jaegerSpan.getLogs());
+  }
+
+  @Test
+  public void testLogWithTimestampAfterFinish() {
+    Map<String, String> unexpectedFields = new HashMap<String, String>() {
```
jpkrohling: It's cleaner to use [`Collections.singletonMap("foo", "bar")`](https://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#singletonMap(K,%20V)) in this case.

jaeger-core/src/test/java/io/jaegertracing/JaegerSpanTest.java
```diff
@@ -444,6 +509,38 @@ public void testNoExpandExceptionLogs() {
     assertEquals(ex, logData.get(0).getFields().get(Fields.ERROR_OBJECT));
   }
 
+  @Test
+  public void testNoLoggingingStackTracesOnDebug() {
+    log.setEnabledLevels(Level.WARN);
+    JaegerSpan jaegerSpan = (JaegerSpan)tracer.buildSpan("foo").start();
```
jpkrohling: I just got a PR merged yesterday that makes this cast unnecessary. Could you rebase your PR and check it?

jaeger-core/src/test/java/io/jaegertracing/JaegerSpanTest.java
```diff
@@ -444,6 +509,38 @@ public void testNoExpandExceptionLogs() {
...
+    LoggingEvent event = log.getLoggingEvents().get(0);
+    assertEquals("Span has already been finished; cannot {}.", event.getMessage());
+    assertEquals(Level.WARN, event.getLevel());
+    assertEquals(false, event.getThrowable().isPresent());
+  }
+
+  @Test
+  public void testLoggingingStackTracesOnDebug() {
+    log.setEnabledLevels(Level.WARN, Level.DEBUG);
+    JaegerSpan jaegerSpan = (JaegerSpan)tracer.buildSpan("foo").start();
```
jpkrohling: Same as the previous comment.

nikron: There have to be a few more changes for this to be possible (buildSpan has to return a `JaegerTracer.SpanBuilder`, and `SpanBuilder` needs to be public.

I'll make these changes, but I don't know if you want to do that in a separate diff

jpkrohling: No no, just rebase to the latest master. Those changes are already there :)

https://github.com/jaegertracing/jaeger-client-java/blob/master/jaeger-core/src/main/java/io/jaegertracing/JaegerTracer.java#L401

PR #480, opened 2018-07-06
Fix wrong package declaration

quaff:
- PR #470 brings inconsistent package name.

- rename package name.


jpkrohling:
Looks like there's someone being smart between the source code and the packaging:

```
$ jar tf ~/.m2/repository/io/jaegertracing/jaeger-core/0.30.0/jaeger-core-0.30.0.jar  | grep Clock
io/jaegertracing/internal/utils/Clock.class
io/jaegertracing/internal/utils/SystemClock.class
```

This prevented the problem from being seen at consumers like `JaegerTracer` or tests, but also prevented this issue from being found earlier.

So, thanks again @quaff for finding this one! You rock :-)

quaff:
It report compiler error by eclipse and vscode, I think gradle is not strict enough.

PR #481, opened 2018-07-06
Add "serialVersionUID" for exceptions to avoid InvalidClassException

quaff:
avoid potential InvalidClassException when perform deserialization.

add generated serialVersionUID for exceptions.


pavolloffay:
@quaff hi, can I ask you about your serialization use case?

quaff:
@pavolloffay please refer to https://stackoverflow.com/questions/285793/what-is-a-serialversionuid-and-why-should-i-use-it

jpkrohling:
I don't think we'll ever need to serialize the exception (or any other class in this client, for that matter), *but*, having a serial version UID for Exceptions is indeed a good practice.

quaff:
@jpkrohling It will used when perform RMI such as spring httpInvoker, exception will be serialized in server and deserialized in client, client rethrow it.

pavolloffay:
@quaff thanks for the link. I am not asking why we should add serialization id, but how you are using jaeger client so you need the serialization.

quaff:
@pavolloffay I'm using spring httpinvoker as microservice implementation, It use java serialization by default, exceptions will propagated from server to client.

jpkrohling:
> It will used when perform RMI such as spring httpInvoker

Interesting. I wouldn't expect the client to be remotely called, or that an exception on the client library to propagate way back to a remote caller, but it would certainly make sense.

I'm curious though: is this an actual scenario you faced, or is it theoretical? If you faced this issue, would you mind sharing more details (source code would be even better). We could then derive more tests from it (cc @jkandasa)

quaff:
The problem is not jaeger specific, but I'm using jaeger for demonstration
```
public interface EchoService {

	String echo(String message);

}

```
```
public class EchoServiceImpl implements EchoService {

	@Override
	public String echo(String message) {
		try (Scope scope = GlobalTracer.get().scopeManager().activate(GlobalTracer.get().buildSpan("echo").start(),
				true)) {
			return message;
		}
	}

}
```
```
@Component
public class Test {

	@Autowired
	private EchoService echoService;

	public void test() {
		echoService.echo("something"); // exception from server will be thrown
	}

}

```

quaff:
@jpkrohling It's not an actual scenario, I found it because eclipse reports warning.

jpkrohling:
Thanks @quaff !

PR #496, opened 2018-07-19
Change return type of Configuration.getTracer be opentracing.Tracer not JaegerTracer

### LINKED ISSUE #497: Bridging legacy API

Currently, we have [legacy-client-java](https://github.com/jaegertracing/legacy-client-java) that is entirely incompatible with the current API. This is more of an issue internally at Uber, but I can imagine the situation existing in open source as well. I am working on a bridge to allow users to upgrade the client to `io.jaegertracing` from `com.uber.jaeger` without the need to change their code (we have literally 100s of projects using the old code right now). Some API incompatibilities exist, but the basic use case is worth preserving where possible.

One problem is that package name change, but an intermediate library can resolve that with clever renaming. The bigger issue lies in the `Configuration.getTracer` method. As of [this commit](https://github.com/jaegertracing/jaeger-client-java/commit/46b64d5dab51447e17266b6de392a0d778285582), we return `io.jaegertracing.internal.JaegerTracer` instead of `io.opentracing.Tracer` (which in and of itself is a potentially breaking change). I suspect it was a simple mistake with find/replace in the file.

In practice, no one had issues with the change because even if the return type didn't match, it could be casted to the interface implicitly.

For example:

```
// Client code
io.opentracing.Tracer tracer = config.getTracer(); 
```

In the client code, both `io.opentracing.Tracer` and `io.jaegertracing.internal.JaegerTracer` are valid because they are both valid instances implementing `io.opentracing.Tracer`.

I would like to revert the change in the return type to use the OT tracer type. That way, I can take advantage of the interface to implement a wrapped tracer using the `com.uber.jaeger` package name, without breaking anyone's code.

isaachier:
Resolves #497.

### DIFF
--- a/jaeger-client/src/test/java/io/jaegertracing/client/VersionTest.java
+++ b/jaeger-client/src/test/java/io/jaegertracing/client/VersionTest.java
@@ -32,7 +32,7 @@ public class VersionTest {
 
     assertNotEquals(
         "The version from the tracer should not be the same string as Version.get()",
-        new Configuration("testVersionGet").getTracer().getVersion(),
+        ((JaegerTracer) new Configuration("testVersionGet").getTracer()).getVersion(),
         Version.get()
     );
   }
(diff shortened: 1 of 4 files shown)


isaachier:
Decided against this in #497.

PR #498, opened 2018-07-20
Fixes #494 - removed unwanted classes from jaeger-thrift shadowed jar

### LINKED ISSUE #494: jaeger-thrift JAR has bundled dependencies, breaking builds
If you look into 0.30.x JARs of jaeger-thrift (see here http://central.maven.org/maven2/io/jaegertracing/jaeger-thrift/0.30.2/jaeger-thrift-0.30.2.jar), you will see a lot of stuff that does not belong there. I guess this is done by `com.github.johnrengelman.shadow` gradle plugin.

Bundling dependencies in this jar breaks dependency management (it is transitive dependency of jaeger-client), and even java compiler - project, compiled with JDK 10, breaks with 0.30.2 version, because it uses old lombok - which i don't reference and don't have on my drive.

I'm using Maven 3.5.3, Spring Boot 2.0.3.RELEASE as parent, io.opentracing.contrib:opentracing-spring-jaeger-starter:0.2.0, which uses jaeger-client:0.30.2.

mdvorak:
Moved relocated packages to `io.jaegertracing.vendor`
Excluded lombok and animal-sniffer from shadow plugin in `jaeger-thrift`
Added slf4j as runtime non-relocated dependency
Removed unused dependencies from `jaeger-crossdock`

I really believe `slf4j-api` should be normal runtime dependency, having it shadowed is not a good idea - it depends on external loggers to work. Relocating slf4j-api breaks that at best, crashes at worst.
(note: `slf4j-api` is transitive dependency of `libthrift`)

### DIFF
--- a/jaeger-crossdock/build.gradle
+++ b/jaeger-crossdock/build.gradle
@@ -13,9 +13,6 @@ compileJava {
 dependencies {
     compile project(':jaeger-client')
 
-    compile group: 'org.apache.thrift', name: 'libthrift', version: apacheThriftVersion
-    compile group: 'com.squareup.okhttp3', name: 'okhttp', version: okhttpVersion
-
     compile group: 'io.opentracing.contrib', name: 'opentracing-jaxrs2', version: '0.1.4'
     compile group: 'com.fasterxml.jackson.core', name: 'jackson-databind', version: jacksonVersion
     compile group: 'org.glassfish.jersey.containers', name: 'jersey-container-grizzly2-http', version: jerseyVersion
--- a/jaeger-thrift/build.gradle
+++ b/jaeger-thrift/build.gradle
@@ -6,7 +6,7 @@ description = 'Library to send data to Jaeger backend components via Thrift'
 dependencies {
     compile project(':jaeger-core')
 
-    compile group: 'org.slf4j', name: 'slf4j-api'
+    compile group: 'org.slf4j', name: 'slf4j-api', version: slf4jVersion
 
     compileOnly group: 'org.apache.thrift', name: 'libthrift', version: apacheThriftVersion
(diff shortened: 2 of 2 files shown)


jaeger-thrift/build.gradle
```diff
@@ -6,6 +6,8 @@ description = 'Library to send data to Jaeger backend components via Thrift'
 dependencies {
     compile project(':jaeger-core')
 
+    compile group: 'org.slf4j', name: 'slf4j-api'
```
jpkrohling: Shouldn't it be `compileOnly`? Try to run a `mvn dependency:tree` (or `./gradlew dependencies`, if that's what your consumer application uses) and check if this version of `jaeger-thrift` (or `jaeger-client`) brings this dependency.

jaeger-crossdock/build.gradle
```diff
@@ -13,9 +13,6 @@ compileJava {
 dependencies {
     compile project(':jaeger-client')
 
-    compile group: 'org.apache.thrift', name: 'libthrift', version: apacheThriftVersion
```
jpkrohling: +1, I think I had to add this in some iteration of the previous work, to get the tests to compile/run. Apparently, it's not needed now :)

jpkrohling: Er... oops :)

https://github.com/jaegertracing/jaeger-client-java/issues/494#issuecomment-405883005

mdvorak:
I've forgotten slf4j version in the dependency, added.

`compile` type dependency should generate `<scope>compile</scope>`, which is what we want for slf4j. Same as jaeger-core.

jpkrohling:
Thanks!

PR #502, opened 2018-07-22
Move SamplingStatus to spi package

### LINKED ISSUE #501: Move SamplingStatus out of internal package

Implementing a custom sampler.

Implementing a sampler (`io.jaegertracing.spi.Sampler`) requires overriding the `sample` method. However, the return type of `SamplingStatus` is an internal class. Furthermore, the constructor for `SamplingStatus` is package local, so overriding it is basically impossible.

In order to make creating a custom sampler feasible, I propose to move `io.jaegertracing.internal.samplers.SamplingStatus` to `io.jaegertracing.spi.SamplingStatus` and to make the constructor public.

isaachier:
Signed-off-by: Isaac Hier <ihier@uber.com>

- Resolves #501.

- Moves SamplingStatus from internal package to spi package.


isaachier:
@yurishkuro @jpkrohling @vprithvi any thoughts about this change?

yurishkuro:
Long term I think this is the right change. Short term I'd prefer not to do it, because the whole point of package renaming was to give us the opportunity to revisit & clean up the interfaces. In particular, returning the status objects is not very efficient as it requires memory allocation. Also, Won and I discussed that we may want to change the sampler signature to be more efficient, in particular by accepting the current time stamp (used by the rate limiter), or the whole span that already contains the time stamp, the span id, and the operation name (but passing the span creates a dependency loop of sorts).

So I would've actually preferred not having the Sampler interface in the spi package just yet.

Why not making this change blocks you from implementing a Sampler? The internal package is still accessible.

isaachier:
I don't believe I need this at the moment. Just checking in. Should I close this or leave it open just in case we want this change later on?

PR #507, opened 2018-07-27
Support for 128 bits trace ids

### LINKED ISSUE #495: Support 128bit trace IDs
https://github.com/jaegertracing/jaeger/issues/858

We should do this work sooner rather than later.

hypnoce:
Fixes #495 : support for 128 bits trace ids.

Added environment variable JAEGER_TRACEID_128BIT
Added TraceId and SpanId classes
Added withTraceId128Bit method to Configuration
Added withTraceId128Bit method to Tracer.Builder
Added useTraceId128Bit property to Tracer

Inspired from https://github.com/jaegertracing/jaeger-client-csharp/pull/94

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/Configuration.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/Configuration.java
@@ -148,6 +148,11 @@ public class Configuration {
   public static final String JAEGER_SENDER_FACTORY = JAEGER_PREFIX + "SENDER_FACTORY";
 
   /**
+   *  Opt-in to use 128 bit traceIds. By default, uses 64 bits.
+   */
+  public static final String JAEGER_TRACEID_128BIT = JAEGER_PREFIX + "TRACEID_128BIT";
+
+  /**
    * The supported trace context propagation formats.
    */
   public enum Propagation {
@@ -172,6 +177,7 @@ public class Configuration {
   private CodecConfiguration codecConfig;
   private MetricsFactory metricsFactory;
   private Map<String, String> tracerTags;
+  private boolean useTraceId128Bit;
 
   /**
    * lazy singleton JaegerTracer initialized in getTracer() method.
@@ -193,6 +199,7 @@ public class Configuration {
   public static Configuration fromEnv(String serviceName) {
     return new Configuration(serviceName)
             .withTracerTags(tracerTagsFromEnv())
+            .withTraceId128Bit(getPropertyAsBool(JAEGER_TRACEID_128BIT))
             .withReporter(ReporterConfiguration.fromEnv())
             .withSampler(SamplerConfiguration.fromEnv())
             .withCodec(CodecConfiguration.fromEnv());
@@ -219,6 +226,9 @@ public class Configuration {
         .withReporter(reporter)
         .withMetrics(metrics)
         .withTags(tracerTags);
+    if (useTraceId128Bit) {
+      builder = builder.withTraceId128Bit();
+    }
     codecConfig.apply(builder);
     return builder;
   }
@@ -282,6 +292,11 @@ public class Configuration {
 
   public Configuration withCodec(CodecConfiguration codecConfig) {
     this.codecConfig = codecConfig;
+    return this;
+  }
+
+  public Configuration withTraceId128Bit(boolean useTraceId128Bit) {
+    this.useTraceId128Bit = useTraceId128Bit;
     return this;
   }
 
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerObjectFactory.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerObjectFactory.java
@@ -57,13 +57,14 @@ public class JaegerObjectFactory {
   }
 
   public JaegerSpanContext createSpanContext(
-      long traceId,
+      long traceIdHigh,
+      long traceIdLow,
       long spanId,
       long parentId,
       byte flags,
       Map<String, String> baggage,
       String debugId) {
-    return new JaegerSpanContext(traceId, spanId, parentId, flags, baggage, debugId, this);
+    return new JaegerSpanContext(traceIdHigh, traceIdLow, spanId, parentId, flags, baggage, debugId, this);
   }
 
   public JaegerTracer.SpanBuilder createSpanBuilder(JaegerTracer tracer, String operationName) {
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
@@ -17,6 +17,7 @@ package io.jaegertracing.internal;
 
 import io.jaegertracing.internal.propagation.TextMapCodec;
 import io.opentracing.SpanContext;
+
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
@@ -25,7 +26,8 @@ public class JaegerSpanContext implements SpanContext {
   protected static final byte flagSampled = 1;
   protected static final byte flagDebug = 2;
 
-  private final long traceId;
+  private final long traceIdLow;
+  private final long traceIdHigh;
   private final long spanId;
   private final long parentId;
   private final byte flags;
@@ -33,9 +35,10 @@ public class JaegerSpanContext implements SpanContext {
   private final String debugId;
   private final JaegerObjectFactory objectFactory;
 
-  public JaegerSpanContext(long traceId, long spanId, long parentId, byte flags) {
+  public JaegerSpanContext(long traceIdHigh, long traceIdLow, long spanId, long parentId, byte flags) {
     this(
-        traceId,
+        traceIdHigh,
+        traceIdLow,
         spanId,
         parentId,
         flags,
@@ -45,7 +48,8 @@ public class JaegerSpanContext implements SpanContext {
   }
 
   protected JaegerSpanContext(
-      long traceId,
+      long traceIdHigh,
+      long traceIdLow,
       long spanId,
       long parentId,
       byte flags,
@@ -55,7 +59,8 @@ public class JaegerSpanContext implements SpanContext {
     if (baggage == null) {
       baggage = Collections.<String, String>emptyMap();
     }
-    this.traceId = traceId;
+    this.traceIdLow = traceIdLow;
+    this.traceIdHigh = traceIdHigh;
     this.spanId = spanId;
     this.parentId = parentId;
     this.flags = flags;
@@ -77,8 +82,27 @@ public class JaegerSpanContext implements SpanContext {
     return this.baggage;
   }
 
-  public long getTraceId() {
-    return traceId;
+  public String getTraceId() {
+    if (traceIdHigh == 0L) {
+      return Long.toHexString(traceIdLow);
+    }
+    final String hexStringHigh = Long.toHexString(traceIdHigh);
+    final String hexStringLow = Long.toHexString(traceIdLow);
+    if (hexStringLow.length() < 16) {
+      // left pad low trace id with '0'.
+      // In theory, only 12.5% of all possible long values will be padded.
+      // In practice, using Random.nextLong(), only 6% will need padding
+      return hexStringHigh + "0000000000000000".substring(hexStringLow.length()) + hexStringLow;
+    }
+    return hexStringHigh + hexStringLow;
+  }
+
+  public long getTraceIdLow() {
+    return traceIdLow;
+  }
+
+  public long getTraceIdHigh() {
+    return traceIdHigh;
   }
 
   public long getSpanId() {
@@ -113,15 +137,15 @@ public class JaegerSpanContext implements SpanContext {
     } else {
       newBaggage.put(key, val);
     }
-    return objectFactory.createSpanContext(traceId, spanId, parentId, flags, newBaggage, debugId);
+    return objectFactory.createSpanContext(traceIdHigh, traceIdLow, spanId, parentId, flags, newBaggage, debugId);
   }
 
   public JaegerSpanContext withBaggage(Map<String, String> newBaggage) {
-    return objectFactory.createSpanContext(traceId, spanId, parentId, flags, newBaggage, debugId);
+    return objectFactory.createSpanContext(traceIdHigh, traceIdLow, spanId, parentId, flags, newBaggage, debugId);
   }
 
   public JaegerSpanContext withFlags(byte flags) {
-    return objectFactory.createSpanContext(traceId, spanId, parentId, flags, baggage, debugId);
+    return objectFactory.createSpanContext(traceIdHigh, traceIdLow, spanId, parentId, flags, baggage, debugId);
   }
 
   /**
@@ -133,7 +157,7 @@ public class JaegerSpanContext implements SpanContext {
    * @see Constants#BAGGAGE_HEADER_KEY
    */
   boolean hasTrace() {
-    return traceId != 0 && spanId != 0;
+    return (traceIdLow != 0 || traceIdHigh != 0) && spanId != 0;
   }
 
   /**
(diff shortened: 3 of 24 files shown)


yurishkuro:
one meta comment/question: is the encapsulation afforded by SpanId type worth an extra memory allocation? With TraceId is kind of unavoidable, since some sort of object is always required (even as `byte[]`)

hypnoce:
This is something that I though of. I followed the c# implementation (even though c# handles better 'value types' in terms of memory than java for now).
I have no opinion on this one. Just that it creates a semantic around ids and is in par with the implementation of trace id.

yurishkuro:
fwiw, census also uses a dedicated type https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/trace/SpanId.java

hypnoce:
@yurishkuro let me know if you find anything that must be adapted in this PR. Thanks !

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
```diff
@@ -68,15 +68,15 @@ public String getBaggageItem(String key) {
     return this.baggage;
   }
 
-  public long getTraceId() {
+  public TraceId getTraceId() {
```
isaachier: This probably breaks backwards-compatibility. Is there a way to preserve the return type? I realize for `TraceId` this may be difficult, but why change `long` to `SpanId` unnecessarily?

jaeger-core/src/test/java/io/jaegertracing/ConfigurationTest.java
```diff
@@ -184,11 +203,11 @@ public void testPropagationB3Only() {
     System.setProperty(Configuration.JAEGER_PROPAGATION, "b3");
     System.setProperty(Configuration.JAEGER_SERVICE_NAME, "Test");
 
-    long traceId = 1234;
-    long spanId = 5678;
+    TraceId traceId = new TraceId(1234);
```
isaachier: Please try a bigger TraceId value, preferably > 64 bits. I'd like to see propagation works for large numbers too.

jaeger-core/src/test/java/io/jaegertracing/internal/SpanIdTest.java
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright (c) 2016, Uber Technologies, Inc
```
isaachier: Please change year to `2018` seeing as this file is new. For other files that exist already, you can do `<year>-2018` based on what was there originally.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
hypnoce: I believe it makes the API cleaner to have SpanId and TraceId as returned type.
I think keeping long for span and TraceId for traces is a short term solution that might change if spanIds become 128 bits for any reason. I would rather encapsulate now than later. What do you think ?

We can also change the method name and have both.

isaachier:
I read the comments here and see @yurishkuro suggested the dedicate span id type. So I'll leave that up to him.

@yurishkuro are you sure Java always allocates new objects and never relies on escape analysis? I'm seeing online that it is used rarely, but this seems like an ideal use case.

hypnoce:
Escape analysis will only work in really special cases that I believe won't work here since the SpanId object is attached to a field of another object that is not itself escaped. Basically SpanId leaves the stack.

isaachier:
@hypnoce I get your point. I just mean that it is really just wrapping a `long` so it ought to be treated as one by the JIT compiler. I'd be surprised that it has to go to the heap each time that value is used.

hypnoce:
Totally see your point. I'm not really sure how the JVM is able to optimise such objects. For sure, 'value-types' (http://openjdk.java.net/jeps/169) will be a perfect solution for this in future releases of Java.

jaeger-core/src/test/java/io/jaegertracing/ConfigurationTest.java
hypnoce: I just noticed that we propagate 128 bits ids in B3 formats. Not sure how the receiver of the trace context (ie the brave library) will react to this.

yurishkuro:
The OT proposal to expose trace/span IDs as strings has been accepted and will be coming down the line as two new methods, asTraceId/asSpanId. So my preference would be to hide the internal trace/span IDs completely, that would allow us to pick the most efficient types, namely `long traceIdLow, traceIdHigh, spanId`

hypnoce:
So you are saying that the method should return strings and not internal id representations ?

hypnoce:
@yurishkuro 
Sorry for the delay. I had a busy month for a release.
Do you prefer to wait for the _new_ methods in the opentracing API and return a String ? Or merge as is and update once the OT api is updated ?

yurishkuro:
We do not have to wait for the new methods, they are not going to affect the internal representation. My main objection is to using heap-allocated types:
```
      TraceId traceId,
      SpanId spanId,
      SpanId parentId,
```
We should think about how we want to represent the IDs efficiently and strategically. The options are:

  * as in this PR (the most inefficient, I think)
  * as longs / pairs of longs
  * as strings
  * as byte arrays

They have their pros and cons:

  * longs require no memory allocations
  * strings are how we receive and send IDs over the wire most of the time
  * byte arrays will be mostly useful when spans are serialized to protobuf (in thrift we still need longs)

We can also realize that except for the root span, the rest of the spans will either get trace ID from the parent span in-process (i.e. as already allocated object), or from serialized representation (ie. a string or a byte array - doing longs requires unmarshalling).

I think to unblock this PR, longs are the simplest path forward. The only breaking change would be getTraceId() would need to return a string instead of long.

hypnoce:
So the getTraceId will allocate a String each time ?
Also, converting the span to thrift requires long. Either we need to expose the high/low ids  or convert back the string to high/low ids when serializing to thrift (longs => string => longs).

hypnoce:
@yurishkuro I've updated the PR according to our previous discussion. Let me know if you have any comment !

jaeger-thrift/src/test/java/io/jaegertracing/thrift/internal/senders/UdpSenderTest.java
```diff
@@ -144,7 +144,7 @@ public void testFlushSendsSpan() throws Exception {
     assertEquals(expectedNumSpans, batch.getSpans().size());
 
     io.jaegertracing.thriftjava.Span actualSpan = batch.getSpans().get(0);
-    assertEquals(expectedSpan.context().getTraceId(), actualSpan.getTraceIdLow());
+    assertEquals(expectedSpan.context().getTraceIdLow(), actualSpan.getTraceIdLow());
     assertEquals(0, actualSpan.getTraceIdHigh());
```
yurishkuro: perhaps this test should be testing High as well, now that we have this upgrade

jaeger-zipkin/src/main/java/io/jaegertracing/zipkin/internal/V2SpanConverter.java
```diff
@@ -45,7 +45,7 @@
     JaegerSpanContext context = span.context();
     zipkin2.Span.Builder builder = zipkin2.Span.newBuilder()
             .id(Long.toHexString(context.getSpanId()))
-            .traceId(Long.toHexString(context.getTraceId()))
+            .traceId(context.getTraceIdHigh(), context.getTraceIdLow())
```
yurishkuro: another example if (high, low)

jaeger-zipkin/src/test/java/io/jaegertracing/zipkin/ZipkinSenderTest.java
```diff
@@ -137,14 +141,35 @@ public void testFlushSendsSpan() throws Exception {
...
+    assertEquals(context.getTraceIdHigh(), 0L);
     assertEquals(context.getSpanId(), HexCodec.lowerHexToUnsignedLong(actualSpan.id()));
     assertEquals(context.getParentId(), 0L);
     assertNull(actualSpan.parentId());
     assertEquals(expectedSpan.getOperationName(), actualSpan.name());
     assertEquals(tracer.getServiceName(), actualSpan.localServiceName());
   }
 
+  @Test
+  public void testAppend128BitTraceId() throws Exception {
```
yurishkuro: It seems the only place 64bit logic should be tested is in the trace ID generation (i.e. tests for `tracer.startspan()`). All other tests, especially those that deal with external formats (codecs and senders) should probably be only operating on 128bit (since it implicitly tests 64bit conversion as well).

But it's probably going to involve a lot of test changes, please lmk.

hypnoce: The changes would have been to big. I prefered to isolate the testing of 128bit (sometimes duplicating the test of 64 bit). So that I'm sure that the 64 bit test coverage stays the same. At some point refactoring the tests might make sense (maybe if deprecating 64bits traceid).
But until 64 bit traceid are considered fully supported, I would rather continue testing everything on 64bit and test 128 support in other/more tests.

jaeger-core/src/main/java/io/jaegertracing/internal/exceptions/TraceIdOutOfBoundException.java
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright (c) 2016, Uber Technologies, Inc
```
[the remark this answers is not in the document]
hypnoce: Done

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerObjectFactory.java
```diff
@@ -57,13 +57,14 @@ public JaegerSpan createSpan(
...
-      long traceId,
+      long traceIdLow,
+      long traceIdHigh,
       long spanId,
       long parentId,
       byte flags,
       Map<String, String> baggage,
       String debugId) {
-    return new JaegerSpanContext(traceId, spanId, parentId, flags, baggage, debugId, this);
+    return new JaegerSpanContext(traceIdLow, traceIdHigh, spanId, parentId, flags, baggage, debugId, this);
```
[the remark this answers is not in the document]
hypnoce: Done

jaeger-thrift/src/test/java/io/jaegertracing/thrift/internal/senders/UdpSenderTest.java
hypnoce: Done

jaeger-core/src/main/java/io/jaegertracing/Configuration.java
```diff
@@ -147,6 +147,11 @@
    */
   public static final String JAEGER_SENDER_FACTORY = JAEGER_PREFIX + "SENDER_FACTORY";
 
+  /**
+   *  Opt-in to use 128 bit traceIds. By default, uses 64 bits.
+   */
+  public static final String JAEGER_TRACEID_128BIT = JAEGER_PREFIX + "TRACEID_128BIT";
```
pavolloffay: It should be documented in readme

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
```diff
@@ -77,8 +83,30 @@ public String getBaggageItem(String key) {
...
+    final String hexStringHigh = Long.toHexString(traceIdHigh);
+    final String hexStringLow = Long.toHexString(traceIdLow);
+    if (hexStringLow.length() < 16) {
+      // left pad low trace id with '0'.
+      // In theory, only 12.5% of all possible long values will be padded.
+      // In practice, using Random.nextLong(), only 6% will need padding
+      final int padLeftLength = 16 - hexStringLow.length();
+      char[] zeroPadLeft = new char[padLeftLength];
+      Arrays.fill(zeroPadLeft, '0');
+      return hexStringHigh + new String(zeroPadLeft) + hexStringLow;
```
jpkrohling: This doesn't look 100% right to me: you are allocating a new array of zeroes every time you need to get the trace ID and the low value is lower than 16. Better would be something like:

```java
"0000000000000000".substring(hexStringLow.length()) + hexStringLow;
```

The zeroes are then treated as constant, which should have a better performance than dynamically allocating an array every time.

jaeger-core/src/main/java/io/jaegertracing/internal/exceptions/TraceIdOutOfBoundException.java
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright (c) 2018, Uber Technologies, Inc
```
jpkrohling: If this is a new source file, this should be assigned to "The Jaeger Tracing Authors". See https://github.com/jaegertracing/jaeger-client-java/blob/master/CONTRIBUTING.md#license-and-certificate-of-origin

jaeger-core/src/main/java/io/jaegertracing/internal/propagation/HexCodec.java
```diff
@@ -32,7 +32,21 @@ static Long lowerHexToUnsignedLong(String lowerHex) {
...
+
+  /**
+   * Parses a 1 to 32 character higher-hex string with no prefix into an unsigned long, tossing any
+   * bits higher than 64.
+   *
+   * @return a 64 bit long, meaning that negative values are the overflow of Java's 32 bit long
+   */
+  static Long higherHexToUnsignedLong(String higherHex) {
+    if (higherHex.length() > 32 || higherHex.length() < 1) {
+      return null;
```
jpkrohling: I think it would be worth a log message somewhere about this...

jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TextMapCodec.java
```diff
@@ -80,14 +81,29 @@ static JaegerSpanContext contextFromString(String value)
...
     return new JaegerSpanContext(
-        new BigInteger(parts[0], 16).longValue(),
+        high(traceId),
+        new BigInteger(traceId, 16).longValue(),
         new BigInteger(parts[1], 16).longValue(),
         new BigInteger(parts[2], 16).longValue(),
         new BigInteger(parts[3], 16).byteValue());
   }
 
+  private static long high(String hexString) {
```
jpkrohling: It *looks* like you expect to receive a full id (high + low) in hex here and you are getting only the first 16 chars. If that's indeed the case, please add a javadoc. 

If you are instead expecting only the high to be in this `hexString`, then add also a log message in the condition below (> 16), as it might indicate that the data the tracer is receiving is bad in some way.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
hypnoce: An array is allocated in copyOfRange. But, I did not catch _copyOfRange was a java intrinsics :).

jaeger-core/src/main/java/io/jaegertracing/Configuration.java
hypnoce: There is currently no structured documentation in the README for the env variable. The only thing the current documentation : https://github.com/jaegertracing/documentation/pull/157 
Do you want to create a place where all the env variable are listed and documented ?

jaeger-core/src/main/java/io/jaegertracing/internal/exceptions/TraceIdOutOfBoundException.java
hypnoce: Done

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
hypnoce: Done

jaeger-core/src/main/java/io/jaegertracing/internal/propagation/HexCodec.java
hypnoce: Done at debug Level. I though warn would be also good since it's a situation that should arrive rarely or that can cause tracing to malfunction.
Let me know what you think.

jpkrohling:
Thank you!

PR #509, opened 2018-08-06
Allow subclassing of common classes

isaachier:
Give protected access to core Jaeger classes so that users may extend them. The main motive here is to create a bridge between the legacy `com.uber.jaeger` classes to the `io.jaegertracing` classes (equivalent to a `typedef` in C/C++). **Note, in general it would not be wise to implement a subclass of these classes because they are internal and may change without notice!**

Additionally, implement abstract factory pattern to create instances of spans, span contexts, and span builders. This allows the caller to provide subclasses of the aforementioned classes. For example, when the tracer needs to create a `JaegerSpanBuilder` object, it invokes `createSpanBuilder` instead. `createSpanBuilder` acts as a hook to instantiate a subclass instead of the `JaegerSpanBuilder` base class.

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/Configuration.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/Configuration.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2018, The Jaeger Authors
  * Copyright (c) 2016, Uber Technologies, Inc
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
@@ -213,13 +214,17 @@ public class Configuration {
     Metrics metrics = new Metrics(metricsFactory);
     Reporter reporter = reporterConfig.getReporter(metrics);
     Sampler sampler = samplerConfig.createSampler(serviceName, metrics);
-    JaegerTracer.Builder builder = new JaegerTracer.Builder(serviceName)
+    JaegerTracer.Builder builder = createTracerBuilder(serviceName)
         .withSampler(sampler)
         .withReporter(reporter)
         .withMetrics(metrics)
         .withTags(tracerTags);
     codecConfig.apply(builder);
     return builder;
+  }
+
+  protected JaegerTracer.Builder createTracerBuilder(String serviceName) {
+    return new JaegerTracer.Builder(serviceName);
   }
 
   public synchronized JaegerTracer getTracer() {
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerObjectFactory.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerObjectFactory.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright (c) 2018, The Jaeger Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.jaegertracing.internal;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Implements abstract factory pattern for creating spans, span contexts, and span builders. This
+ * pattern allows subclasses of JaegerSpan, JaegerSpanContext, and JaegerTracer.SpanBuilder to be
+ * used consistently in the trace instead of the base class.
+ *
+ * <p>Example usage:</p>
+ *
+ * <pre>{@code
+ *     public class CustomObjectFactory extends JaegerObjectFactory {
+ *         \@Override
+ *         public JaegerSpan createSpan(...) {
+ *           return new CustomSpan(...);
+ *         }
+ *
+ *         // Override other methods...
+ *     }
+ * }</pre>
+ */
+public class JaegerObjectFactory {
+  public JaegerSpan createSpan(
+      JaegerTracer tracer,
+      String operationName,
+      JaegerSpanContext context,
+      long startTimeMicroseconds,
+      long startTimeNanoTicks,
+      boolean computeDurationViaNanoTicks,
+      Map<String, Object> tags,
+      List<Reference> references) {
+    return new JaegerSpan(
+        tracer,
+        operationName,
+        context,
+        startTimeMicroseconds,
+        startTimeNanoTicks,
+        computeDurationViaNanoTicks,
+        tags,
+        references);
+  }
+
+  public JaegerSpanContext createSpanContext(
+      long traceId,
+      long spanId,
+      long parentId,
+      byte flags,
+      Map<String, String> baggage,
+      String debugId) {
+    return new JaegerSpanContext(traceId, spanId, parentId, flags, baggage, debugId, this);
+  }
+
+  public JaegerTracer.SpanBuilder createSpanBuilder(JaegerTracer tracer, String operationName) {
+    return tracer.new SpanBuilder(operationName);
+  }
+}
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpan.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpan.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2018, The Jaeger Authors
  * Copyright (c) 2016, Uber Technologies, Inc
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
@@ -46,7 +47,7 @@ public class JaegerSpan implements Span {
   private List<LogData> logs;
   private boolean finished = false; // to prevent the same span from getting reported multiple times
 
-  JaegerSpan(
+  protected JaegerSpan(
       JaegerTracer tracer,
       String operationName,
       JaegerSpanContext context,
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2018, The Jaeger Authors
  * Copyright (c) 2016, Uber Technologies, Inc
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
@@ -30,18 +31,27 @@ public class JaegerSpanContext implements SpanContext {
   private final byte flags;
   private final Map<String, String> baggage;
   private final String debugId;
+  private final JaegerObjectFactory objectFactory;
 
   public JaegerSpanContext(long traceId, long spanId, long parentId, byte flags) {
-    this(traceId, spanId, parentId, flags, Collections.<String, String>emptyMap(), null);
+    this(
+        traceId,
+        spanId,
+        parentId,
+        flags,
+        Collections.<String, String>emptyMap(),
+        null, // debugId
+        new JaegerObjectFactory());
   }
 
-  JaegerSpanContext(
+  protected JaegerSpanContext(
       long traceId,
       long spanId,
       long parentId,
       byte flags,
       Map<String, String> baggage,
-      String debugId) {
+      String debugId,
+      JaegerObjectFactory objectFactory) {
     if (baggage == null) {
       throw new NullPointerException();
     }
@@ -51,6 +61,7 @@ public class JaegerSpanContext implements SpanContext {
     this.flags = flags;
     this.baggage = baggage;
     this.debugId = debugId;
+    this.objectFactory = objectFactory;
   }
 
   @Override
@@ -102,15 +113,15 @@ public class JaegerSpanContext implements SpanContext {
     } else {
       newBaggage.put(key, val);
     }
-    return new JaegerSpanContext(traceId, spanId, parentId, flags, newBaggage, debugId);
+    return objectFactory.createSpanContext(traceId, spanId, parentId, flags, newBaggage, debugId);
   }
 
   public JaegerSpanContext withBaggage(Map<String, String> newBaggage) {
-    return new JaegerSpanContext(traceId, spanId, parentId, flags, newBaggage, debugId);
+    return objectFactory.createSpanContext(traceId, spanId, parentId, flags, newBaggage, debugId);
   }
 
   public JaegerSpanContext withFlags(byte flags) {
-    return new JaegerSpanContext(traceId, spanId, parentId, flags, baggage, debugId);
+    return objectFactory.createSpanContext(traceId, spanId, parentId, flags, baggage, debugId);
   }
 
   /**
@@ -127,21 +138,12 @@ public class JaegerSpanContext implements SpanContext {
   }
 
   /**
-   * Create a new dummy JaegerSpanContext as a container for debugId string. This is used when
-   * "jaeger-debug-id" header is passed in the request headers and forces the trace to be sampled as
-   * debug trace, and the value of header recorded as a span tag to serve as a searchable
-   * correlation ID.
-   *
-   * @param debugId arbitrary string used as correlation ID
-   *
-   * @return new dummy JaegerSpanContext that serves as a container for debugId only.
+   * debugId is used when "jaeger-debug-id" header is passed in the request headers and forces the
+   * trace to be sampled as debug trace, and the value of header recorded as a span tag to serve as
+   * a searchable correlation ID.
    *
    * @see Constants#DEBUG_ID_HEADER_KEY
    */
-  public static JaegerSpanContext withDebugId(String debugId) {
-    return new JaegerSpanContext(0, 0, 0, (byte) 0, Collections.<String, String>emptyMap(), debugId);
-  }
-
   String getDebugId() {
     return debugId;
   }
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2018, The Jaeger Authors
  * Copyright (c) 2016, Uber Technologies, Inc
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
@@ -74,32 +75,24 @@ public class JaegerTracer implements Tracer, Closeable {
   private final ScopeManager scopeManager;
   private final BaggageSetter baggageSetter;
   private final boolean expandExceptionLogs;
+  private final JaegerObjectFactory objectFactory;
 
-  private JaegerTracer(
-      String serviceName,
-      Reporter reporter,
-      Sampler sampler,
-      PropagationRegistry registry,
-      Clock clock,
-      Metrics metrics,
-      Map<String, Object> tags,
-      boolean zipkinSharedRpcSpan,
-      ScopeManager scopeManager,
-      BaggageRestrictionManager baggageRestrictionManager,
-      boolean expandExceptionLogs) {
-    this.serviceName = serviceName;
-    this.reporter = reporter;
-    this.sampler = sampler;
-    this.registry = registry;
-    this.clock = clock;
-    this.metrics = metrics;
-    this.zipkinSharedRpcSpan = zipkinSharedRpcSpan;
-    this.scopeManager = scopeManager;
-    this.baggageSetter = new BaggageSetter(baggageRestrictionManager, metrics);
-    this.expandExceptionLogs = expandExceptionLogs;
+  protected JaegerTracer(JaegerTracer.Builder builder) {
+    this.serviceName = builder.serviceName;
+    this.reporter = builder.reporter;
+    this.sampler = builder.sampler;
+    this.registry = builder.registry;
+    this.clock = builder.clock;
+    this.metrics = builder.metrics;
+    this.zipkinSharedRpcSpan = builder.zipkinSharedRpcSpan;
+    this.scopeManager = builder.scopeManager;
+    this.baggageSetter = new BaggageSetter(builder.baggageRestrictionManager, metrics);
+    this.expandExceptionLogs = builder.expandExceptionLogs;
+    this.objectFactory = builder.objectFactory;
 
     this.version = loadVersion();
 
+    Map<String, Object> tags = new HashMap<String, Object>(builder.tags);
     tags.put(Constants.JAEGER_CLIENT_VERSION_TAG_KEY, this.version);
     if (tags.get(Constants.TRACER_HOSTNAME_TAG_KEY) == null) {
       String hostname = getHostName();
@@ -184,7 +177,7 @@ public class JaegerTracer implements Tracer, Closeable {
 
   @Override
   public JaegerTracer.SpanBuilder buildSpan(String operationName) {
-    return new SpanBuilder(operationName);
+    return objectFactory.createSpanBuilder(this, operationName);
   }
 
   @Override
@@ -223,10 +216,11 @@ public class JaegerTracer implements Tracer, Closeable {
      * a collection of references.
      */
     private List<Reference> references = Collections.emptyList();
+
     private final Map<String, Object> tags = new HashMap<String, Object>();
     private boolean ignoreActiveSpan = false;
 
-    SpanBuilder(String operationName) {
+    protected SpanBuilder(String operationName) {
       this.operationName = operationName;
     }
 
@@ -315,7 +309,13 @@ public class JaegerTracer implements Tracer, Closeable {
         }
       }
 
-      return new JaegerSpanContext(id, id, 0, flags);
+      return getObjectFactory().createSpanContext(
+          id,
+          id,
+          0,
+          flags,
+          Collections.<String, String>emptyMap(),
+          debugId);
     }
 
     private Map<String, String> createChildBaggage() {
@@ -354,7 +354,7 @@ public class JaegerTracer implements Tracer, Closeable {
         }
       }
 
-      return new JaegerSpanContext(
+      return getObjectFactory().createSpanContext(
           preferredReference.getTraceId(),
           Utils.uniqueId(),
           preferredReference.getSpanId(),
@@ -429,7 +429,7 @@ public class JaegerTracer implements Tracer, Closeable {
         }
       }
 
-      JaegerSpan jaegerSpan = new JaegerSpan(
+      JaegerSpan jaegerSpan = getObjectFactory().createSpan(
               JaegerTracer.this,
               operationName,
               context,
@@ -462,12 +462,16 @@ public class JaegerTracer implements Tracer, Closeable {
     public JaegerSpan startManual() {
       return start();
     }
+
+    private JaegerObjectFactory getObjectFactory() {
+      return JaegerTracer.this.objectFactory;
+    }
   }
 
   /**
    * Builds a {@link JaegerTracer} with options.
    */
-  public static final class Builder {
+  public static class Builder {
     private Sampler sampler;
     private Reporter reporter;
     private final PropagationRegistry registry = new PropagationRegistry();
@@ -479,13 +483,28 @@ public class JaegerTracer implements Tracer, Closeable {
     private ScopeManager scopeManager = new ThreadLocalScopeManager();
     private BaggageRestrictionManager baggageRestrictionManager = new DefaultBaggageRestrictionManager();
     private boolean expandExceptionLogs;
+    private final JaegerObjectFactory objectFactory;
 
     public Builder(String serviceName) {
+      this(serviceName, new JaegerObjectFactory());
+    }
+
+    protected Builder(String serviceName, JaegerObjectFactory objectFactory) {
       this.serviceName = checkValidServiceName(serviceName);
-      TextMapCodec textMapCodec = new TextMapCodec(false);
+      this.objectFactory = objectFactory;
+
+      TextMapCodec textMapCodec =
+          TextMapCodec.builder()
+              .withUrlEncoding(false)
+              .withObjectFactory(this.objectFactory)
+              .build();
       this.registerInjector(Format.Builtin.TEXT_MAP, textMapCodec);
       this.registerExtractor(Format.Builtin.TEXT_MAP, textMapCodec);
-      TextMapCodec httpCodec = new TextMapCodec(true);
+      TextMapCodec httpCodec =
+          TextMapCodec.builder()
+              .withUrlEncoding(true)
+              .withObjectFactory(this.objectFactory)
+              .build();
       this.registerInjector(Format.Builtin.HTTP_HEADERS, httpCodec);
       this.registerExtractor(Format.Builtin.HTTP_HEADERS, httpCodec);
       // TODO binary codec not implemented
@@ -590,8 +609,11 @@ public class JaegerTracer implements Tracer, Closeable {
             .withMetrics(metrics)
             .build();
       }
-      return new JaegerTracer(serviceName, reporter, sampler, registry, clock, metrics, tags,
-          zipkinSharedRpcSpan, scopeManager, baggageRestrictionManager, expandExceptionLogs);
+      return createTracer();
+    }
+
+    protected JaegerTracer createTracer() {
+      return new JaegerTracer(this);
     }
 
     public static String checkValidServiceName(String serviceName) {
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/propagation/B3TextMapCodec.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/propagation/B3TextMapCodec.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2018, The Jaeger Authors
  * Copyright (c) 2017, Uber Technologies, Inc
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
@@ -14,11 +15,13 @@
 
 package io.jaegertracing.internal.propagation;
 
+import io.jaegertracing.internal.JaegerObjectFactory;
 import io.jaegertracing.internal.JaegerSpanContext;
 import io.jaegertracing.internal.JaegerTracer;
 import io.jaegertracing.spi.BaggageRestrictionManager;
 import io.jaegertracing.spi.Codec;
 import io.opentracing.propagation.TextMap;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -57,6 +60,7 @@ public class B3TextMapCodec implements Codec<TextMap> {
 
   private static final PrefixedKeys keys = new PrefixedKeys();
   private final String baggagePrefix;
+  private final JaegerObjectFactory objectFactory;
 
   /**
    * @deprecated use {@link Builder} instead
@@ -68,6 +72,7 @@ public class B3TextMapCodec implements Codec<TextMap> {
 
   private B3TextMapCodec(Builder builder) {
     this.baggagePrefix = builder.baggagePrefix;
+    this.objectFactory = builder.objectFactory;
   }
 
   @Override
@@ -118,7 +123,14 @@ public class B3TextMapCodec implements Codec<TextMap> {
     }
 
     if (null != traceId && null != parentId && null != spanId) {
-      JaegerSpanContext spanContext = new JaegerSpanContext(traceId, spanId, parentId, flags);
+      JaegerSpanContext spanContext = objectFactory.createSpanContext(
+          traceId,
+          spanId,
+          parentId,
+          flags,
+          Collections.<String, String>emptyMap(),
+          null // debugId
+          );
       if (baggage != null) {
         spanContext = spanContext.withBaggage(baggage);
       }
@@ -129,12 +141,22 @@ public class B3TextMapCodec implements Codec<TextMap> {
 
   public static class Builder {
     private String baggagePrefix = BAGGAGE_PREFIX;
+    private JaegerObjectFactory objectFactory = new JaegerObjectFactory();
 
     /**
      * Specify baggage prefix. The default is {@value B3TextMapCodec#BAGGAGE_PREFIX}
      */
     public Builder withBaggagePrefix(String baggagePrefix) {
       this.baggagePrefix = baggagePrefix;
+      return this;
+    }
+
+    /**
+     * Specify JaegerSpanContext factory. Used for creating new span contexts. The default factory
+     * is an instance of {@link JaegerObjectFactory}.
+     */
+    public Builder withObjectFactory(JaegerObjectFactory objectFactory) {
+      this.objectFactory = objectFactory;
       return this;
     }
 
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TextMapCodec.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TextMapCodec.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2018, The Jaeger Authors
  * Copyright (c) 2016, Uber Technologies, Inc
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
@@ -15,6 +16,7 @@
 package io.jaegertracing.internal.propagation;
 
 import io.jaegertracing.internal.Constants;
+import io.jaegertracing.internal.JaegerObjectFactory;
 import io.jaegertracing.internal.JaegerSpanContext;
 import io.jaegertracing.internal.exceptions.EmptyTracerStateStringException;
 import io.jaegertracing.internal.exceptions.MalformedTracerStateStringException;
@@ -24,6 +26,7 @@ import java.io.UnsupportedEncodingException;
 import java.math.BigInteger;
 import java.net.URLDecoder;
 import java.net.URLEncoder;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
@@ -47,6 +50,11 @@ public class TextMapCodec implements Codec<TextMap> {
 
   private final boolean urlEncoding;
 
+  /**
+   * Object factory used to construct JaegerSpanContext subclass instances.
+   */
+  private final JaegerObjectFactory objectFactory;
+
   public TextMapCodec(boolean urlEncoding) {
     this(builder().withUrlEncoding(urlEncoding));
   }
@@ -55,6 +63,7 @@ public class TextMapCodec implements Codec<TextMap> {
     this.urlEncoding = builder.urlEncoding;
     this.contextKey = builder.spanContextKey;
     this.baggagePrefix = builder.baggagePrefix;
+    this.objectFactory = builder.objectFactory;
   }
 
   static JaegerSpanContext contextFromString(String value)
@@ -120,7 +129,13 @@ public class TextMapCodec implements Codec<TextMap> {
     }
     if (context == null) {
       if (debugId != null) {
-        return JaegerSpanContext.withDebugId(debugId);
+        return objectFactory.createSpanContext(
+            0,
+            0,
+            0,
+            (byte) 0,
+            Collections.<String, String>emptyMap(),
+            debugId);
       }
       return null;
     }
@@ -185,6 +200,7 @@ public class TextMapCodec implements Codec<TextMap> {
     private boolean urlEncoding;
     private String spanContextKey = SPAN_CONTEXT_KEY;
     private String baggagePrefix = BAGGAGE_KEY_PREFIX;
+    private JaegerObjectFactory objectFactory = new JaegerObjectFactory();
 
     public Builder withUrlEncoding(boolean urlEncoding) {
       this.urlEncoding = urlEncoding;
@@ -198,6 +214,16 @@ public class TextMapCodec implements Codec<TextMap> {
 
     public Builder withBaggagePrefix(String baggagePrefix) {
       this.baggagePrefix = baggagePrefix;
+      return this;
+    }
+
+    /**
+     * Set object factory to use for construction of JaegerSpanContext subclass instances.
+     *
+     * @param objectFactory JaegerObjectFactory subclass instance.
+     */
+    public Builder withObjectFactory(JaegerObjectFactory objectFactory) {
+      this.objectFactory = objectFactory;
       return this;
     }
 
--- a/jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
+++ b/jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
@@ -1,0 +1,174 @@
+/*
+ * Copyright (c) 2018, The Jaeger Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.jaegertracing.internal;
+
+import io.jaegertracing.Configuration;
+import io.opentracing.Scope;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class JaegerSubclassTest {
+  private static class CustomConfiguration extends Configuration {
+    private CustomConfiguration(String serviceName) {
+      super(serviceName);
+    }
+
+    @Override
+    public CustomTracer.CustomBuilder getTracerBuilder() {
+      return (CustomTracer.CustomBuilder) super.getTracerBuilder();
+    }
+
+    @Override
+    public synchronized CustomTracer getTracer() {
+      return (CustomTracer) super.getTracer();
+    }
+
+    @Override
+    protected CustomTracer.CustomBuilder createTracerBuilder(String serviceName) {
+      return new CustomTracer.CustomBuilder(serviceName);
+    }
+  }
+
+  private static class CustomTracer extends JaegerTracer {
+    public static class CustomBuilder extends JaegerTracer.Builder {
+      private CustomBuilder(String serviceName) {
+        super(serviceName, new CustomObjectFactory());
+      }
+
+      @Override
+      public CustomTracer build() {
+        return (CustomTracer) super.build();
+      }
+
+      @Override
+      protected JaegerTracer createTracer() {
+        return new CustomTracer(this);
+      }
+    }
+
+    public class CustomSpanBuilder extends JaegerTracer.SpanBuilder {
+      protected CustomSpanBuilder(String operationName) {
+        super(operationName);
+      }
+
+      @Override
+      public CustomSpan start() {
+        return (CustomSpan) super.start();
+      }
+    }
+
+    private CustomTracer(CustomBuilder builder) {
+      super(builder);
+    }
+  }
+
+  private static class CustomSpan extends JaegerSpan {
+    private CustomSpan(
+        CustomTracer tracer,
+        String operationName,
+        CustomSpanContext context,
+        long startTimeMicroseconds,
+        long startTimeNanoTicks,
+        boolean computeDurationViaNanoTicks,
+        Map<String, Object> tags,
+        List<Reference> references) {
+      super(
+          tracer,
+          operationName,
+          context,
+          startTimeMicroseconds,
+          startTimeNanoTicks,
+          computeDurationViaNanoTicks,
+          tags,
+          references);
+    }
+
+    @Override
+    public CustomSpanContext context() {
+      return (CustomSpanContext) super.context();
+    }
+  }
+
+  private static class CustomSpanContext extends JaegerSpanContext {
+    private CustomSpanContext(
+        long traceId,
+        long spanId,
+        long parentId,
+        byte flags,
+        Map<String, String> baggage,
+        String debugId,
+        CustomObjectFactory objectFactory) {
+      super(traceId, spanId, parentId, flags, baggage, debugId, objectFactory);
+    }
+  }
+
+  private static class CustomObjectFactory extends JaegerObjectFactory {
+    @Override
+    public CustomSpan createSpan(
+        JaegerTracer tracer,
+        String operationName,
+        JaegerSpanContext context,
+        long startTimeMicroseconds,
+        long startTimeNanoTicks,
+        boolean computeDurationViaNanoTicks,
+        Map<String, Object> tags,
+        List<Reference> references) {
+      return new CustomSpan(
+          (CustomTracer) tracer,
+          operationName,
+          (CustomSpanContext) context,
+          startTimeMicroseconds,
+          startTimeNanoTicks,
+          computeDurationViaNanoTicks,
+          tags,
+          references);
+    }
+
+    @Override
+    public CustomSpanContext createSpanContext(
+        long traceId,
+        long spanId,
+        long parentId,
+        byte flags,
+        Map<String, String> baggage,
+        String debugId) {
+      return new CustomSpanContext(traceId, spanId, parentId, flags, baggage, debugId, this);
+    }
+
+    @Override
+    public CustomTracer.CustomSpanBuilder createSpanBuilder(
+        JaegerTracer tracer, String operationName) {
+      return ((CustomTracer) tracer).new CustomSpanBuilder(operationName);
+    }
+  }
+
+  @Test
+  public void testTracer() {
+    final CustomConfiguration config = new CustomConfiguration("test-service");
+    final CustomTracer.CustomBuilder builder = config.getTracerBuilder();
+    final CustomTracer tracer = builder.build();
+    final Scope scope = tracer.buildSpan("test-operation").startActive(true);
+    Assert.assertNotNull(tracer.scopeManager().active());
+    Assert.assertTrue(tracer.scopeManager().active().span() instanceof CustomSpan);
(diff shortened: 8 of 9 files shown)


jaeger-core/src/main/java/io/jaegertracing/Configuration.java
```diff
@@ -462,7 +461,7 @@ public void apply(JaegerTracer.Builder builder) {
       registerCodec(builder, Format.Builtin.TEXT_MAP);
     }
 
-    protected void registerCodec(JaegerTracer.Builder builder, Format<TextMap> format) {
+    private void registerCodec(JaegerTracer.Builder builder, Format<TextMap> format) {
```
yurishkuro: why this?

isaachier: Find/replace mistake. Will remove.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpan.java
```diff
@@ -46,7 +46,7 @@
   private List<LogData> logs;
   private boolean finished = false; // to prevent the same span from getting reported multiple times
 
-  JaegerSpan(
+  protected JaegerSpan(
```
yurishkuro: @isaachier I don't think this will be enough, is it?  I.e. you'd also need an overridable method that actually creates the object.

yurishkuro: I would prefer if you accompany this with a unit test showing the working subclassing.

isaachier: I added a lot more. The test definitely helped me determine what was necessary.

isaachier:
Lol I'm not trying to. Travis CI kept failing the build because of the formatting and licensing issues.

yurishkuro:
If you need to fix formatting, then let's do it in a separate PR.

isaachier:
Undid the formatting fix for now.

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    @Override
+    public SpanContext context() {
+      return ctx;
+    }
+  }
+
+  private static class SpanContext extends JaegerSpanContext {
+    SpanContext(JaegerSpanContext ctx) {
```
yurishkuro: That's not what I thought you were going after. If you extend the context, why do you need to accept it as agrument? The oss tracer currently does "new SpanContext". You can move that call into a function that you would override in the subclassed tracer and return your own subclass of SpanContext.

isaachier: I could redo a lot of code. The idea is to avoid code duplication by allowing base classes to do 99% of the work. Imagine the case of a JaegerTracer starting a JaegerSpan. There is a lot of logic dedicated to determining the span parent, etc. Now I could override `start` because Tracer needs to return `Span` instead of `JaegerSpan`, but that's going to result in a huge amount of copy-paste. Isn't it easier to use a constructor `Span(JaegerSpan span)` (note package visibility, not public)?

yurishkuro: I am not suggesting redoing any code. What I am saying is that, for example, Trace has this code in `builder.createNewContext()`:
```
return new JaegerSpanContext(id, id, 0, flags);
```

`createNewContext` itself is a large method we don't want to replicate. But we can change the above line to
```
return instantiateNewSpanContext(id, id, 0, flags);
```

and add method:
```
protected JaegerSpanContext instantiateNewSpanContext(...) {
  return new JaegerSpanContext(id, id, 0, flags);
}
```

Then in your subclassed Tracer you just need to override this method
```
@Override
protected JaegerSpanContext instantiateNewSpanContext(...) {
  return new SubclassedJaegerSpanContext(id, id, 0, flags);
}
```

Your subclassed tracer will have a total of 3 methods: constructor that directly delegates to `super(....same args.....)`, and two instantiate methods, for span context and the Span that create 
subclassed instances.

jaeger-core/src/main/java/io/jaegertracing/Configuration.java
```diff
@@ -653,19 +652,19 @@ public static SenderConfiguration fromEnv() {
     }
   }
 
-  private static String stringOrDefault(String value, String defaultValue) {
+  protected static String stringOrDefault(String value, String defaultValue) {
```
jpkrohling: Are you making them `protected` so that you can *use* them in subclasses? Those are not part of the business of this class, I'd rather see them in a, say, `StringUtils` class (as much as a I dislike `utils` classes).

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
```diff
@@ -0,0 +1,179 @@
...
+import io.opentracing.ScopeManager;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class JaegerSubclassTest {
+  private static class Configuration extends io.jaegertracing.Configuration {
```
jpkrohling: `s/Configuration/CustomConfiguration/` (or anything else that won't class with i.j.Configuration)

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
```diff
@@ -0,0 +1,179 @@
...
+      return new Configuration(io.jaegertracing.Configuration.fromEnv());
+    }
+
+    @Override
+    public synchronized Tracer getTracer() {
+      return new Tracer(super.getTracer());
+    }
+  }
+
+  private static class Tracer extends JaegerTracer {
```
jpkrohling: `s/Tracer/CustomTracer/` (or anything else that won't clash with io.opentracing.Tracer, which is part of the hierarchy)

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
```diff
@@ -0,0 +1,179 @@
...
+        return new Span(super.start());
+      }
+
+      @Override
+      public Scope startActive(boolean finishSpanOnClose) {
+        return scopeManager().activate(start(), finishSpanOnClose);
+      }
+    }
+
+    Tracer(io.jaegertracing.internal.JaegerTracer tracer) {
```
jpkrohling: You probably don't need the fully qualified class name here, do you?

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
```diff
@@ -0,0 +1,179 @@
...
+          expandExceptionLogs);
+    }
+
+    @Override
+    public SpanBuilder buildSpan(String operationName) {
+      return new SpanBuilder(operationName);
+    }
+  }
+
+  private static class Span extends JaegerSpan {
```
jpkrohling: `s/Span/CustomSpan`

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
```diff
@@ -0,0 +1,179 @@
...
+      super(traceId, spanId, parentId, flags, baggage, debugId);
+    }
+  }
+
+  @Test
+  public void testTracer() {
+    final Configuration config = new Configuration("test-service");
+    final Tracer tracer = config.getTracer();
+    final Scope scope = tracer.buildSpan("test-operation").startActive(true);
+    Assert.assertNotNull(tracer.scopeManager().active());
```
jpkrohling: I understand that if this assertion works, then a bunch of other stuff has also worked, but it would be nice to have your "base" assertions made explicit here.

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
isaachier: OK now I understand. I think this is called the [abstract factory pattern](https://en.wikipedia.org/wiki/Abstract_factory_pattern). That definitely makes sense. I am usually hesitant to change so much of the existing code, but I guess this is a better design.

jaeger-core/src/main/java/io/jaegertracing/Configuration.java
isaachier: Honestly, my subclass doesn't use these. I thought if they were `static private` utility methods here, then they ought to be `static protected` for subclasses. Nonetheless, I was considering adding a new class called `ConfigUtils` or something just for that purpose.

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
isaachier: Right good idea.

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
isaachier: Definitely

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
isaachier: Good catch.

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
isaachier: Ya this test isn't perfect. Needs work.

isaachier: Should be a bit better now.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
```diff
@@ -122,11 +126,17 @@ public static JaegerSpanContext contextFromString(String value)
...
         new BigInteger(parts[1], 16).longValue(),
         new BigInteger(parts[2], 16).longValue(),
-        new BigInteger(parts[3], 16).byteValue());
+        new BigInteger(parts[3], 16).byteValue(),
+        Collections.<String, String>emptyMap(),
+        null);
+  }
+
+  protected TracingFactory tracingFactory() {
+    return new TracingFactory();
```
yurishkuro: factory should be instantiated only once

jaeger-core/src/main/java/io/jaegertracing/Configuration.java
```diff
@@ -190,11 +190,20 @@ public static Configuration fromEnv() {
...
-            .withTracerTags(tracerTagsFromEnv())
-            .withReporter(ReporterConfiguration.fromEnv())
-            .withSampler(SamplerConfiguration.fromEnv())
-            .withCodec(CodecConfiguration.fromEnv());
+    Configuration config = new Configuration(serviceName);
+    config.initFromEnv();
+    return config;
+  }
+
+  protected void initFromEnv() {
```
yurishkuro: why? this is not the code we want to repeat, so why is it factored out into a method?

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
isaachier: I was thinking about using a singleton. I guess I should.

jaeger-core/src/main/java/io/jaegertracing/Configuration.java
isaachier: This is because `Configuration` has no factory method at the moment. All these static methods are problematic because they construct the concrete type instead of the subclass. In general, I was able to solve by passing or overriding the factory. I can try doing the same here.

yurishkuro: just an idea - while subclassing tracer / span is necessary, I don't think the same holds for the configuration class. It's fine to wrap it, and not provide all the functionality that exists in the oss. For example, internally we don't init from env.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
yurishkuro: not a singleton in the sense of a private static member, but just an externally provided object.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
```diff
@@ -32,18 +32,30 @@
...
   private final Map<String, String> baggage;
   private final String debugId;
+  private final TracingFactory tracingFactory;
 
   public JaegerSpanContext(long traceId, long spanId, long parentId, byte flags) {
     this(traceId, spanId, parentId, flags, Collections.<String, String>emptyMap(), null);
   }
 
-  JaegerSpanContext(
+  protected JaegerSpanContext(
```
yurishkuro: who would use this ctor?

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
```diff
@@ -107,7 +120,11 @@ public String toString() {
     return contextAsString();
   }
 
-  public static JaegerSpanContext contextFromString(String value)
+  public static JaegerSpanContext contextFromString(String value) {
```
yurishkuro: who uses this?

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
```diff
@@ -72,10 +73,12 @@
   private final Map<String, ?> tags;
   private final boolean zipkinSharedRpcSpan;
   private final ScopeManager scopeManager;
+  private final BaggageRestrictionManager baggageRestrictionManager;
   private final BaggageSetter baggageSetter;
   private final boolean expandExceptionLogs;
+  private final TracingFactory tracingFactory;
 
-  private JaegerTracer(
+  protected JaegerTracer(
```
yurishkuro: I don't see why you need to duplicate constructors - it was private previously, nobody called it externally, so it should be ok to change its API

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
isaachier: Well it is `public` and this isn't the legacy client, so I feel uncomfortable deleting it or adding a required argument. It is used internally to decode a span context from a `TextMap`.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
isaachier: Good point. I'll remove this one and any others I find that were originally `private` or package scope.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
isaachier: It isn't exactly popular, but it is the more complete constructor. If the subclass wants to override the other constructor, it would need to use this constructor and pass in a custom factory for the last argument.

isaachier: Sorry, now that I see the constructor I understand what you mean. Will remove.

jaeger-core/src/main/java/io/jaegertracing/Configuration.java
```diff
@@ -177,10 +179,14 @@
    */
   private JaegerTracer tracer;
 
-  public Configuration(String serviceName) {
+  public Configuration(String serviceName, TracingFactory tracingFactory) {
```
yurishkuro: did you see my comment? https://github.com/jaegertracing/jaeger-client-java/pull/509#discussion_r208400650 ?

jaeger-core/src/main/java/io/jaegertracing/Configuration.java
```diff
@@ -190,11 +196,11 @@ public static Configuration fromEnv() {
...
-    return new Configuration(serviceName)
-            .withTracerTags(tracerTagsFromEnv())
-            .withReporter(ReporterConfiguration.fromEnv())
-            .withSampler(SamplerConfiguration.fromEnv())
-            .withCodec(CodecConfiguration.fromEnv());
+    return new Configuration(serviceName).
+        withTracerTags(tracerTagsFromEnv())
+        .withReporter(ReporterConfiguration.fromEnv())
+        .withSampler(SamplerConfiguration.fromEnv())
+        .withCodec(CodecConfiguration.fromEnv());
```
yurishkuro: doesn't look like there's a change here, please undo

jaeger-core/src/main/java/io/jaegertracing/Configuration.java
isaachier: I did. I'll comment more there. Need the tracing factory to construct the `JaegerTracer.Builder` subclass used in the config to create the tracer.

jaeger-core/src/main/java/io/jaegertracing/Configuration.java
isaachier: Good idea.

jaeger-core/src/main/java/io/jaegertracing/Configuration.java
isaachier: Agreed. I reverted the protected methods back to private.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
```diff
@@ -122,11 +129,13 @@ public static JaegerSpanContext contextFromString(String value)
...
     */
-    return new JaegerSpanContext(
+    return factory.createSpanContext(
         new BigInteger(parts[0], 16).longValue(),
         new BigInteger(parts[1], 16).longValue(),
         new BigInteger(parts[2], 16).longValue(),
-        new BigInteger(parts[3], 16).byteValue());
+        new BigInteger(parts[3], 16).byteValue(),
+        Collections.<String, String>emptyMap(),
+        null);
```
yurishkuro: what is null?  should it be `factory`?

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
```diff
@@ -166,14 +175,19 @@ boolean isDebugIdContainerOnly() {
    * debug trace, and the value of header recorded as a span tag to serve as a searchable
    * correlation ID.
    *
-   * @param debugId arbitrary string used as correlation ID
+   * @param debugId arbitrary string used as correlation ID.
+   * @param factory tracing object factory.
```
yurishkuro: The name TracingFactory was making me uncomfortable, I was thinking ObjectFactory, but it too was too generic. I think TracingObjectFactory is much cleaner name

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
```diff
@@ -95,8 +99,10 @@ private JaegerTracer(
     this.metrics = metrics;
     this.zipkinSharedRpcSpan = zipkinSharedRpcSpan;
     this.scopeManager = scopeManager;
+    this.baggageRestrictionManager = baggageRestrictionManager;
```
yurishkuro: so, like, a bug? 👍

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
```diff
@@ -315,7 +321,7 @@ private JaegerSpanContext createNewContext(String debugId) {
         }
       }
 
-      return new JaegerSpanContext(id, id, 0, flags);
+      return JaegerTracer.this.tracingFactory.createSpanContext(id, id, 0, flags, Collections.<String, String>emptyMap(), debugId);
```
yurishkuro: I am seeing a mix of `factory` and `tracingFactory`. If you agree with my above comment to call it `TracingObjectFactory`, then I'd suggest field name `objectFactory`, to be used consistently

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
```diff
@@ -354,7 +360,7 @@ private JaegerSpanContext createChildContext() {
         }
       }
 
-      return new JaegerSpanContext(
+      return JaegerTracer.this.tracingFactory.createSpanContext(
```
yurishkuro: just curious, why fully qualified accessor instead of jus `tracingFactory.createSpanContext` ?  I think the latter is also visible.

jaeger-core/src/main/java/io/jaegertracing/internal/TracingFactory.java
```diff
@@ -0,0 +1,77 @@
...
+import io.jaegertracing.internal.metrics.Metrics;
+import io.jaegertracing.spi.BaggageRestrictionManager;
+import io.jaegertracing.spi.Reporter;
+import io.jaegertracing.spi.Sampler;
+import io.opentracing.ScopeManager;
+
+import java.util.List;
+import java.util.Map;
+
+public class TracingFactory {
```
yurishkuro: add javadoc

jaeger-core/src/main/java/io/jaegertracing/internal/TracingFactory.java
```diff
@@ -0,0 +1,77 @@
...
+      Sampler sampler,
+      PropagationRegistry registry,
+      Clock clock,
+      Metrics metrics,
+      Map<String, Object> tags,
+      boolean zipkinSharedRpcSpan,
+      ScopeManager scopeManager,
+      BaggageRestrictionManager baggageRestrictionManager,
+      boolean expandExceptionLogs) {
+    return new JaegerTracer(
```
yurishkuro: per my comment on Configuration, I don't think this method needs to be in the factory. Here's what you can do instead, to reduce the surface area:
```java
class Configuration {
  public JaegerTracer.Builder getTracerBuilder() {
    - JaegerTracer.Builder builder = tracingFactory.createTracerBuilder(serviceName)
    + JaegerTracer.Builder builder = this.createTracerBuilder(serviceName)
  }

  protected createTracerBuilder() // overridable method, no more dependency on the factory
}

similar thing with JaegerTracer.Builder - just implement a protected method to create the tracer.
```

The only time where you need the abstract factory pattern is for creating SpanBuilder/Span/SpanContext, because those are done all the time from many places. But Tracer.Builder and Tracer are only created once, there's no need to complicate that with abstract factory pattern.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
yurishkuro: I would deprecate this method. I think the only reason it is public is because Codecs are in a different package. Since you already have `factory.createSpanContext()` method, the only thing this is doing is parsing a string representation of the span context - arguably a function that should live in the codecs themselves, since there can be different representations like the upcoming W3C Trace Context spec. So I suggest refactoring this and moving the main logic to TextMapCodec (it can be a public static method there, always taking the factory).

jaeger-core/src/main/java/io/jaegertracing/internal/TracingFactory.java
isaachier: I thought about doing that, but sort of wanted to do everything in factory or everything as subclass methods. This makes sense as a compromise.

jaeger-core/src/main/java/io/jaegertracing/internal/TracingFactory.java
isaachier: Ya. Just need to stabilize the API, then will do.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
isaachier: This is actually a reference to `debugId`. But this should definitely take a `TracingFactory` argument as well. Will fix.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
isaachier: OK I struggled with the names. JaegerFactory or JaegerObjectFactory could work too.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
isaachier: No actually just wasn't needed outside of constructor. Originally needed this when I had constructor `CustomTracer(JaegerTracer tracer)`, but don't need this anymore. Will remove.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
isaachier: I thought I renamed all `factory` references to `tracingFactory`. Will change to `objectFactory` throughout.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
isaachier: `JaegerTracer.SpanBuilder` is a non-static nested class. I am able to access the tracer's `tracingFactory` from the inner class. Personally, not a huge fan of non-static nested classes (prefer explicitly passing instance to static nested class instead).

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
isaachier: Never mind. The factory passes `this` internally to the `JaegerSpanContext` constructor.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
yurishkuro: +1 JaegerObjectFactory

jaeger-core/src/main/java/io/jaegertracing/Configuration.java
```diff
@@ -15,6 +15,7 @@
 package io.jaegertracing;
 
 import io.jaegertracing.internal.JaegerTracer;
+import io.jaegertracing.internal.JaegerObjectFactory;
```
yurishkuro: shouldn't need this

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerObjectFactory.java
```diff
@@ -0,0 +1,46 @@
...
+        operationName,
+        context,
+        startTimeMicroseconds,
+        startTimeNanoTicks,
+        computeDurationViaNanoTicks,
+        tags,
+        references);
+  }
+
+  public JaegerSpanContext createSpanContext(long traceId,
```
yurishkuro: formatting inconsistent with the previous method

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
```diff
@@ -166,14 +151,19 @@ boolean isDebugIdContainerOnly() {
...
    *
    * @see Constants#DEBUG_ID_HEADER_KEY
    */
+  public static JaegerSpanContext withDebugId(String debugId, JaegerObjectFactory factory) {
+    return factory.createSpanContext(0, 0, 0, (byte) 0, Collections.<String, String>emptyMap(), debugId);
+  }
+
   public static JaegerSpanContext withDebugId(String debugId) {
-    return new JaegerSpanContext(0, 0, 0, (byte) 0, Collections.<String, String>emptyMap(), debugId);
+    return withDebugId(debugId, new JaegerObjectFactory());
```
yurishkuro: pretty sure we can drop this method - it's `internal` package after all

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
yurishkuro: remove

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
```diff
@@ -107,26 +109,9 @@ public String toString() {
...
-    /*
-      oibe: because java doesn't like to convert large hex strings to longs
-      we should write this manually instead of using BigInteger.
-    */
-    return new JaegerSpanContext(
-        new BigInteger(parts[0], 16).longValue(),
-        new BigInteger(parts[1], 16).longValue(),
-        new BigInteger(parts[2], 16).longValue(),
-        new BigInteger(parts[3], 16).byteValue());
+  @Deprecated
```
yurishkuro: since this is `internal` package, can probably just remove this altogether

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
```diff
@@ -166,14 +151,19 @@ boolean isDebugIdContainerOnly() {
...
+   * @param debugId arbitrary string used as correlation ID.
+   * @param factory tracing object factory.
    *
    * @return new dummy JaegerSpanContext that serves as a container for debugId only.
    *
    * @see Constants#DEBUG_ID_HEADER_KEY
    */
+  public static JaegerSpanContext withDebugId(String debugId, JaegerObjectFactory factory) {
+    return factory.createSpanContext(0, 0, 0, (byte) 0, Collections.<String, String>emptyMap(), debugId);
+  }
```
yurishkuro: who is calling it?  I assume codecs only, so same fate as with the other public static methods.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
```diff
@@ -315,7 +320,7 @@ private JaegerSpanContext createNewContext(String debugId) {
         }
       }
 
-      return new JaegerSpanContext(id, id, 0, flags);
+      return JaegerTracer.this.objectFactory.createSpanContext(id, id, 0, flags, Collections.<String, String>emptyMap(), debugId);
```
yurishkuro: feels like it could use a helper function, if only to avoid this `JaegerTracer.this.objectFactory.createSpanContext` verbosity

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
```diff
@@ -479,18 +484,26 @@ public JaegerSpan startManual() {
...
+      TextMapCodec httpCodec = TextMapCodec.builder().withUrlEncoding(true).withTracingFactory(this.objectFactory).build();
       this.registerInjector(Format.Builtin.HTTP_HEADERS, httpCodec);
       this.registerExtractor(Format.Builtin.HTTP_HEADERS, httpCodec);
       // TODO binary codec not implemented
     }
 
+    @Deprecated
+    public Builder(String serviceName) {
+      this(serviceName, new JaegerObjectFactory());
+    }
```
yurishkuro: again, internal package, do we need this?

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
```diff
@@ -590,8 +603,24 @@ public JaegerTracer build() {
             .withMetrics(metrics)
             .build();
       }
+      return createTracer(serviceName, reporter, sampler, registry, clock, metrics, tags,
+          zipkinSharedRpcSpan, scopeManager, baggageRestrictionManager, expandExceptionLogs, objectFactory);
+    }
+
+    protected JaegerTracer createTracer(String serviceName,
```
yurishkuro: We used to have an auto-formatted in this repository, not sure it's still in effect. But pretty sure this style is ugly as hell - either inline them, or start from a new line so that they are only slightly indented (the way go fmt does)

jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TextMapCodec.java
```diff
@@ -83,7 +91,7 @@ public JaegerSpanContext extract(TextMap carrier) {
     }
     if (context == null) {
       if (debugId != null) {
-        return JaegerSpanContext.withDebugId(debugId);
```
yurishkuro: so you're not even using that withDebugId method! Kill it.

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
```diff
@@ -0,0 +1,231 @@
...
+      }
+
+      @Override
+      public CustomSpan start() {
+        return (CustomSpan) super.start();
+      }
+    }
+
+    public CustomTracer(
+        String serviceName,
```
yurishkuro: this! let's use this formatting.

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
```diff
@@ -0,0 +1,231 @@
...
+          expandExceptionLogs,
+          objectFactory);
+    }
+  }
+
+  private static class CustomSpan extends JaegerSpan {
+    public CustomSpan(
+        JaegerTracer tracer,
+        String operationName,
+        JaegerSpanContext context,
```
yurishkuro: nit: this should be `CustomSpanContext`

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
```diff
@@ -0,0 +1,231 @@
...
+
+  private static class CustomSpanContext extends JaegerSpanContext {
+    public CustomSpanContext(
+        long traceId,
+        long spanId,
+        long parentId,
+        byte flags,
+        Map<String, String> baggage,
+        String debugId,
+        JaegerObjectFactory objectFactory) {
```
yurishkuro: CustomObjectFactory

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
```diff
@@ -0,0 +1,231 @@
...
+      super(traceId, spanId, parentId, flags, baggage, debugId, objectFactory);
+    }
+
+    @Override
+    public CustomSpanContext withFlags(byte flags) {
```
yurishkuro: why do you need to override these?

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    @Override
+    public CustomSpanContext context() {
+      return (CustomSpanContext) super.context();
+    }
+  }
+
+  private static class CustomSpanContext extends JaegerSpanContext {
+    public CustomSpanContext(
```
yurishkuro: could it be private? go for minimum visibility

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
```diff
@@ -0,0 +1,231 @@
...
+                                               long parentId,
+                                               byte flags,
+                                               Map<String, String> baggage,
+                                               String debugId) {
+      return new CustomSpanContext(traceId, spanId, parentId, flags, baggage, debugId, this);
+    }
+
+    @Override
+    public CustomTracer.CustomSpanBuilder createSpanBuilder(JaegerTracer tracer, String operationName) {
+      return ((CustomTracer)tracer).new CustomSpanBuilder(operationName);
```
yurishkuro: Do you actually need to override span builder? Wouldn't the default one do the trick?

yurishkuro:
looking much better!

jaeger-core/src/main/java/io/jaegertracing/Configuration.java
isaachier: Ya I use the Google formatting plugin that will adjust the import and formatting at once, so I held off to avoid the formatting issues from earlier.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerObjectFactory.java
isaachier: Ya will fix formatting after main code is good to go.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
isaachier: OK will remove.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
isaachier: Lol same explanation as above.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
isaachier: Nope.

jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TextMapCodec.java
isaachier: Definitely.

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
isaachier: haha same

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
isaachier: Good idea.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerObjectFactory.java
```diff
@@ -0,0 +1,46 @@
...
+import io.jaegertracing.internal.metrics.Metrics;
+import io.jaegertracing.spi.BaggageRestrictionManager;
+import io.jaegertracing.spi.Reporter;
+import io.jaegertracing.spi.Sampler;
+import io.opentracing.ScopeManager;
+
+import java.util.List;
+import java.util.Map;
+
+public class JaegerObjectFactory {
```
pavolloffay: What is a purpose of this class? Can you please document it here?

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
isaachier: Sadly the covariant return type adds a lot of boilerplate where I have to cast things.

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
isaachier: I think so. Didn't want to change visibility from base class. Constructors should be okay with that.

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
isaachier: I wish... I have not found a way to do it in Java. Maybe by somehow (ab)using a generic helper class to generate bridge methods that are JVM generated casts.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerObjectFactory.java
isaachier: I will very soon.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
isaachier: OK makes sense.

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
isaachier: For a practical example, here is what happens if I don't override them:

```
CustomSpan customSpan = anotherCustomSpan.withFlags(flags);  // Error: cannot assign JaegerSpan to CustomSpan.
```

It will probably break user code without overriding.

yurishkuro: Are you worried that someone would write
```
com.uber.jaeger.SpanContext _ = spanCtx.withFlags()
```
?  I highly doubt it, but ok, we can leave these as a form of example.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
```diff
@@ -107,44 +114,22 @@ public String toString() {
     return contextAsString();
   }
 
-  public static JaegerSpanContext contextFromString(String value)
```
yurishkuro: may want to move this particular change into a separate PR.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
```diff
@@ -479,13 +493,24 @@ public JaegerSpan startManual() {
     private ScopeManager scopeManager = new ThreadLocalScopeManager();
     private BaggageRestrictionManager baggageRestrictionManager = new DefaultBaggageRestrictionManager();
     private boolean expandExceptionLogs;
+    private JaegerObjectFactory objectFactory;
 
     public Builder(String serviceName) {
       this.serviceName = checkValidServiceName(serviceName);
-      TextMapCodec textMapCodec = new TextMapCodec(false);
+      this.objectFactory = createObjectFactory();
```
yurishkuro: hmm, my old (mostly dead) C++ skills turn in their grave on this one. Calling virtual method from a constructor? Even if not a problem in Java it feels very non-intuitive. Maybe this?
```
public Builder(String serviceName)  {
    this(serviceName, new JaegerObjectFactory());
}

protected Builder(String serviceName, JaegerObjectFactory objectFactory)  {
   // actual ctor
}
```
Then you only need to override `createTracer` (and don't need to pass it the factory)

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
```diff
@@ -479,13 +493,24 @@ public JaegerSpan startManual() {
     private ScopeManager scopeManager = new ThreadLocalScopeManager();
     private BaggageRestrictionManager baggageRestrictionManager = new DefaultBaggageRestrictionManager();
     private boolean expandExceptionLogs;
+    private JaegerObjectFactory objectFactory;
```
yurishkuro: `protected final` ?

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
```diff
@@ -590,8 +615,51 @@ public JaegerTracer build() {
...
+          sampler,
+          registry,
+          clock,
+          metrics,
+          tags,
+          zipkinSharedRpcSpan,
+          scopeManager,
+          baggageRestrictionManager,
+          expandExceptionLogs,
+          objectFactory);
```
yurishkuro: torn between this and just passing the Builder. It will need some accessors (maybe getters), but would reduce the toil of 10+ parameters.

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
```diff
@@ -0,0 +1,254 @@
...
+        byte flags,
+        Map<String, String> baggage,
+        String debugId) {
+      return new CustomSpanContext(traceId, spanId, parentId, flags, baggage, debugId, this);
+    }
+
+    @Override
+    public CustomTracer.CustomSpanBuilder createSpanBuilder(
+        JaegerTracer tracer, String operationName) {
+      return ((CustomTracer) tracer).new CustomSpanBuilder(operationName);
```
yurishkuro: `(CustomTracer) tracer)` - would `CustomTracer.this` work?

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
isaachier: Sure.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
isaachier: Haha. Indeed you are correct! Good catch. I definitely like the idea of less additional virtual methods.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
isaachier: Maybe. `final` is fine. This might only be in C++, but I've heard to avoid protected fields. The logic is that you might want to remove the field later and will definitely break subclasses. If you use a protected setter/getter, then you can write functions to maintain some sort of compatibility.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
isaachier: I 100% agree that this is beyond OK. In [Clean Code](http://a.co/bfhJe1V), Robert C. Martin suggests:

clean-code-args

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
isaachier: In this context, `this` refers to the `CustomObjectFactory` instance, so I doubt it.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
isaachier: Luckily, C++ will be much easier to change down the road if I need to. I seriously miss typedefs, templates, and even the dark magic of implicit cast operators now that I have seen how tedious it is to migrate Java code.

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
isaachier: For `SpanContext`, I found 19 cases in internal Uber code. For `Span` I found 281.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
```diff
@@ -32,18 +29,27 @@
...
 
   public JaegerSpanContext(long traceId, long spanId, long parentId, byte flags) {
-    this(traceId, spanId, parentId, flags, Collections.<String, String>emptyMap(), null);
+    this(
+        traceId,
+        spanId,
+        parentId,
+        flags,
+        Collections.<String, String>emptyMap(),
+        null,
```
yurishkuro: if this is debugId, add a comment `// debugId`

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
yurishkuro: why is this still here? Didn't you merge it as a separate PR already? Maybe rebase from master?

isaachier: No just haven't gotten to fixing it yet. Sorry.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
```diff
@@ -74,44 +77,35 @@
...
+    this.scopeManager = builder.scopeManager;
+    this.baggageSetter = new BaggageSetter(builder.baggageRestrictionManager, metrics);
+    this.expandExceptionLogs = builder.expandExceptionLogs;
+    this.objectFactory = builder.objectFactory;
 
     this.version = loadVersion();
 
-    tags.put(Constants.JAEGER_CLIENT_VERSION_TAG_KEY, this.version);
-    if (tags.get(Constants.TRACER_HOSTNAME_TAG_KEY) == null) {
+    builder.tags.put(Constants.JAEGER_CLIENT_VERSION_TAG_KEY, this.version);
```
yurishkuro: nit: instead of modifying `builder.tags` I would create a local variable as a copy
```java
Map<String, String> tags = new HashMap<String, String>(builder.tags);
```

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
```diff
@@ -429,15 +425,17 @@ public JaegerSpan start() {
...
-      JaegerSpan jaegerSpan = new JaegerSpan(
-              JaegerTracer.this,
-              operationName,
-              context,
-              startTimeMicroseconds,
-              startTimeNanoTicks,
-              computeDurationViaNanoTicks,
-              tags,
-              references);
+      JaegerSpan jaegerSpan =
```
yurishkuro: Just curious, why not `JaegerSpan jaegerSpan = getObjectFactory().createSpan(` on one line? 
Would be less code changes overall.

jaeger-core/src/main/java/io/jaegertracing/internal/PropagationRegistry.java
```diff
@@ -22,7 +22,7 @@
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 
-class PropagationRegistry {
+public class PropagationRegistry {
```
yurishkuro: looking at https://github.com/jaegertracing/legacy-client-java/pull/20/files, I don't see it being constructed there. So why public?

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
```diff
@@ -0,0 +1,196 @@
...
+      }
+
+      @Override
+      public CustomTracer build() {
+        return (CustomTracer) super.build();
+      }
+
+      @Override
+      protected JaegerTracer createTracer() {
+        return new CustomTracer(this);
```
yurishkuro: neat! Managed without making builder fields public/protected :-)

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
yurishkuro: _referring_ to the class, yes, but I doubt they use these methods. But ok if you want to keep them.

yurishkuro: I'd just be skeptical about adding them to the actual subclass (not this test)

yurishkuro:
overall lgtm. Let's make sure to minimize the unnecessary changes (like formatting only, or the removal of SpanContext.fromString()). Also, what's with the broken build?

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
isaachier: This was auto-formatted. I will try your style and see what the auto-formatter does with it.

jaeger-core/src/main/java/io/jaegertracing/internal/PropagationRegistry.java
isaachier: My mistake. Forgot to remove.

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
isaachier: Ya I forgot this feature to, but it worked.

isaachier:
I think Travis is having issues with the checkstyle job. I reverted a change I made in a JavaDoc that crashed the checkstyle plugin.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
isaachier: Didn't let me use this format. Changed it for me.

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerObjectFactoryTest.java
```diff
@@ -0,0 +1,175 @@
...
+import io.opentracing.Scope;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class JaegerObjectFactoryTest {
```
yurishkuro: technically, this is not testing the object factory, but the extension mechanism.

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerObjectFactoryTest.java
```diff
@@ -0,0 +1,175 @@
...
+    @Override
+    public CustomTracer.CustomSpanBuilder createSpanBuilder(
+        JaegerTracer tracer, String operationName) {
+      return ((CustomTracer) tracer).new CustomSpanBuilder(operationName);
+    }
+  }
+
+  @Test
+  public void testTracer() {
+    final CustomObjectFactory tracingObjectFactory = new CustomObjectFactory();
```
yurishkuro: is this used?

yurishkuro:
please update PR description to more clearly explain the purpose of this change.

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerObjectFactoryTest.java
isaachier: Will check.

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerObjectFactoryTest.java
isaachier: Fair enough. Will revert renaming.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
+#Mon Aug 13 12:13:38 EDT 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.7-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.7-all.zip
```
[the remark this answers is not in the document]
isaachier: This happened automatically somehow... will revert.

yurishkuro:
I would like @jpkrohling or @pavolloffay or @objectiser to approve before we merge, since they've done the most work on Java client recently.

isaachier:
I'm cool with that. I will admit this review is blocking some work internally. Using snapshot releases isn't easy for the projects I am dealing with, so I'd like to wrap this up ASAP.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerObjectFactory.java
```diff
@@ -0,0 +1,72 @@
+/*
+ * Copyright (c) 2018, Uber Technologies, Inc
```
jpkrohling: nit: The Jaeger Authors

jaeger-core/src/main/java/io/jaegertracing/internal/propagation/B3TextMapCodec.java
```diff
@@ -129,6 +140,7 @@ public JaegerSpanContext extract(TextMap carrier) {
 
   public static class Builder {
     private String baggagePrefix = BAGGAGE_PREFIX;
+    private JaegerObjectFactory objectFactory = new JaegerObjectFactory();
```
jpkrohling: You could use the service loader API here, so that having a JaegerObjectFactory in the classpath would be sufficient to use an alternate object factory. Look at how we did it for the MetricsFactory/Sender for a reference.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
```diff
@@ -107,7 +117,7 @@ public String toString() {
     return contextAsString();
   }
 
-  public static JaegerSpanContext contextFromString(String value)
+  public static JaegerSpanContext contextFromString(String value, JaegerObjectFactory objectFactory)
```
jpkrohling: To avoid breaking current code (there should be none, but still), keep the old signature, add `@Deprecated` and use the new one with `new JaegerObjectFactory()` (or whatever the default value should be):

```java
/**
* @deprecated see {@link #contextFromString(String, JaegerObjectFactory)}
*/
@Deprecated
public static JaegerSpanContext contextFromString(String value) {
  return contextFromString(value, new JaegerObjectFactory());
}
```

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
```diff
@@ -0,0 +1,174 @@
+/*
+ * Copyright (c) 2018, Uber Technologies, Inc
```
jpkrohling: :-)

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerObjectFactory.java
isaachier: Good point. This happens to be copied from other files, but we should correct them.

jaeger-core/src/main/java/io/jaegertracing/internal/propagation/B3TextMapCodec.java
isaachier: I had actually been considering something along those lines, but didn't realize there were other implementations. Thanks!

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
isaachier: @yurishkuro points out these internal classes aren't part of official API and users are very unlikely to be calling `contextFromString` directly.

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSubclassTest.java
isaachier: Got it thanks :).

isaachier:
I just squashed the commits and rebased them on top of latest master branch (includes TextMapCodec changes).

PR #517, opened 2018-08-17
Move contextFromString out of JaegerSpanContext into TextMapCodec

### LINKED ISSUE #518: Move internal method contextFromString out of JaegerSpanContext

Currently, the internal `contextFromString` method resides in the `JaegerSpanContext` class. It produces a `JaegerSpanContext` instance from a string that conforms with the `TextMapCodec`. Arguably, this is not an ideal location for it to reside.

In the future, we would like to support other codecs, such as W3C trace identifiers (https://github.com/jaegertracing/jaeger/issues/855). This method implies the only valid codec is the `TextMapCodec` seeing as it is the canonical method for using strings to produce `JaegerSpanContext` and implicitly relies on the `TextMapCodec` format.

Move the `contextFromString` method into the `TextMapCodec` class instead.

isaachier:
Fixes #518.

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TextMapCodec.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TextMapCodec.java
@@ -16,9 +16,12 @@ package io.jaegertracing.internal.propagation;
 
 import io.jaegertracing.internal.Constants;
 import io.jaegertracing.internal.JaegerSpanContext;
+import io.jaegertracing.internal.exceptions.EmptyTracerStateStringException;
+import io.jaegertracing.internal.exceptions.MalformedTracerStateStringException;
 import io.jaegertracing.spi.Codec;
 import io.opentracing.propagation.TextMap;
 import java.io.UnsupportedEncodingException;
+import java.math.BigInteger;
 import java.net.URLDecoder;
 import java.net.URLEncoder;
 import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


isaachier:
As usual, Travis fails due to crossdock timeout.

jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TextMapCodec.java
```diff
@@ -54,6 +57,25 @@ private TextMapCodec(Builder builder) {
     this.baggagePrefix = builder.baggagePrefix;
   }
 
+  public static JaegerSpanContext contextFromString(String value)
```
jpkrohling: If it's used only in tests, can't we make this private and test the inject operation instead?

yurishkuro: This has been blocking the other PR, so I would rather not cause major refactoring of the tests. But going for minimum visibility is always a good goal. In this case, wouldn't all the tests work just as well if this was package level viz?

isaachier: I tried it and sadly no. `propagation` is its own package.

yurishkuro: Ah, yes. Ok then.

isaachier: I can probably fix the other tests. Don't merge yet.

yurishkuro: Oops too late. Put another one if a simple fix.

PR #519, opened 2018-08-17
Reduce visibility of TextMapCodec.contextFromString to package scope

isaachier:
- Additional fix for #518 to reduce visibility.

- Change visibility.
- Move all tests that rely on this method's functionality into the TextMapCodecTest class.


jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSpanContextTest.java
```diff
@@ -57,10 +35,5 @@ public void testToStringFormatsPositiveFields() {
 
     assertEquals(
         "fffffffffffffff6:fffffffffffffff6:fffffffffffffff6:81", context.contextAsString());
-    JaegerSpanContext contextFromStr = TextMapCodec.contextFromString(context.contextAsString());
-    assertEquals(traceId, contextFromStr.getTraceId());
-    assertEquals(spanId, contextFromStr.getSpanId());
-    assertEquals(parentId, contextFromStr.getParentId());
-    assertEquals(flags, contextFromStr.getFlags());
```
yurishkuro: don't see how removing this makes the test equivalent. But this whole test can move to text codec test, since it's not really testing the context toString

isaachier: Right let me fix that.

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSpanTest.java
```diff
@@ -208,18 +208,6 @@ public void testWithoutTimestampsInaccurateClock() {
...
-  @Test
-  public void testSpanToString() {
-    JaegerSpan jaegerSpan = tracer.buildSpan("test-operation").start();
-    JaegerSpanContext expectedContext =  jaegerSpan.context();
-    JaegerSpanContext actualContext = TextMapCodec.contextFromString(expectedContext.contextAsString());
-
-    assertEquals(expectedContext.getTraceId(), actualContext.getTraceId());
-    assertEquals(expectedContext.getSpanId(), actualContext.getSpanId());
-    assertEquals(expectedContext.getParentId(), actualContext.getParentId());
-    assertEquals(expectedContext.getFlags(), actualContext.getFlags());
```
yurishkuro: removing it is not equivalent. But you can change L215 to use TextMapCodec's extract method.

isaachier: I am not sure if this still applies. See testSpanToString in `TextMapCodec`.

jaeger-core/src/test/java/io/jaegertracing/internal/propagation/TextMapCodecTest.java
```diff
@@ -44,4 +50,55 @@ public void testWithoutBuilder() {
...
+
+  @Test
+  public void testToStringFormatsPositiveFields() {
+    long traceId = -10L;
+    long spanId = -10L;
+    long parentId = -10L;
+    byte flags = (byte) 129;
+
+    // I use MIN_VALUE because the most significant bit, and thats when
+    // we want to make sure the hex number is positive.
```
yurishkuro: this comment looks like some old left-over. Doesn't seem to make any sense now.

isaachier: Good catch.

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSpanTest.java
yurishkuro: Here's what I am thinking. The method is called `testSpanToString`. Such method should belong in SpanTest class, not the codecs. But it's not actually testing `span.toString()`, only `spanContext.contextAsString()`. So let's separate these.
  * to test `span.toString()`, we can simply compare with an exact literal string (including the operation name).
  * what is the contract of `contextAsString()`? Do we, strictly speaking, care if the output matches the output of the codec? What if we use a different codec in the future?  I think it mattered more when span context had the opposite `fromString()` method, but now `contextAsString()` seems merely a dependency of `span.toString()`, nothing more. Or is it used elsewhere?  If not, we can not test it at all (since it's covered by `span.toString()` test), and we can consider making it package viz.



Counter-proposal: keep the test here (since it's testing Span's behavior, not the codec), but

isaachier: But what? :)

yurishkuro: sorry, a left over

jaeger-core/src/test/java/io/jaegertracing/internal/JaegerSpanTest.java
```diff
@@ -210,14 +209,11 @@ public void testWithoutTimestampsInaccurateClock() {
...
+    String operation = "test-operation";
+    JaegerSpan span = tracer.buildSpan(operation).start();
+    String expectedString = span.context().toString() + " - " + operation;
```
yurishkuro: prod == prod.  Please test for exact string.

isaachier: You mean I should use a string literal instead of `expectedString`? Not sure what the suggestion is.

isaachier: Right. Will do.

PR #525, opened 2018-08-20
Support `jaeger-baggage` header for ad-hoc baggage

yurishkuro:
Per https://github.com/jaegertracing/jaeger/issues/1010

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
@@ -53,7 +53,7 @@ public class JaegerSpanContext implements SpanContext {
       String debugId,
       JaegerObjectFactory objectFactory) {
     if (baggage == null) {
-      throw new NullPointerException();
+      baggage = Collections.<String, String>emptyMap();
     }
     this.traceId = traceId;
     this.spanId = spanId;
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TextMapCodec.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TextMapCodec.java
@@ -22,15 +22,18 @@ import io.jaegertracing.internal.exceptions.EmptyTracerStateStringException;
 import io.jaegertracing.internal.exceptions.MalformedTracerStateStringException;
 import io.jaegertracing.spi.Codec;
 import io.opentracing.propagation.TextMap;
+
 import java.io.UnsupportedEncodingException;
 import java.math.BigInteger;
 import java.net.URLDecoder;
 import java.net.URLEncoder;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
 
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
 public class TextMapCodec implements Codec<TextMap> {
   /**
    * Key used to store serialized span context representation
@@ -142,14 +145,15 @@ public class TextMapCodec implements Codec<TextMap> {
   }
 
   private Map<String, String> parseBaggageHeader(String header, Map<String, String> baggage) {
-    String[] parts = header.split("\\ *,\\ *");
-    for (int i = 0; i < parts.length; i++) {
-      String[] kv = parts[i].split("\\ *=\\ *");
+    for (String part : header.split("\\s*,\\s*")) {
+      String[] kv = part.split("\\s*=\\s*");
       if (kv.length == 2) {
         if (baggage == null) {
           baggage = new HashMap<String, String>();
         }
         baggage.put(kv[0], kv[1]);
+      } else {
+        log.debug("malformed token in {} header: {}", Constants.BAGGAGE_HEADER_KEY, part);
       }
     }
     return baggage;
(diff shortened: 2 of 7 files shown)


jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TextMapCodec.java
```diff
@@ -116,20 +116,39 @@ public JaegerSpanContext extract(TextMap carrier) {
...
 
+  private Map<String, String> parseBaggageHeader(String header, Map<String, String> baggage) {
+    String[] parts = header.split("\\ *,\\ *");
+    for (int i = 0; i < parts.length; i++) {
+      String[] kv = parts[i].split("\\ *=\\ *");
+      if (kv.length == 2) {
+        if (baggage == null) {
+          baggage = new HashMap<String, String>();
+        }
+        baggage.put(kv[0], kv[1]);
```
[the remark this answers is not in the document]
jpkrohling: How about making this configurable? `JAEGER_MAX_NUM_BAGGAGE_ITEMS`, `JAEGER_MAX_SIZE_BAGGAGE_ITEMS`, with a default of 10 baggage items and 256 bytes per key and value (512 bytes per pair)?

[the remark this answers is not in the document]
yurishkuro: Env vars feel out of place for this. The baggage size policy should be controlled centrally in organization, and we already have a mechanism for that via baggage restrictions manager.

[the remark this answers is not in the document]
jpkrohling: Do we have it documented somewhere? I never used it myself, wouldn't know how to even start using that.

jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TextMapCodec.java
```diff
@@ -116,20 +116,39 @@ public JaegerSpanContext extract(TextMap carrier) {
...
       return context;
     }
+    if (context == null) {
+      context = new JaegerSpanContext(0, 0, 0, (byte)0);
+    }
     return context.withBaggage(baggage);
   }
 
+  private Map<String, String> parseBaggageHeader(String header, Map<String, String> baggage) {
+    String[] parts = header.split("\\ *,\\ *");
```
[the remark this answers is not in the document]
yurishkuro: It's similar, but that one has additional variable resolution logic for values.

jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TextMapCodec.java
```diff
@@ -125,24 +125,34 @@ public JaegerSpanContext extract(TextMap carrier) {
...
+      context == null ? 0 : context.getSpanId(),
+      context == null ? 0 : context.getParentId(),
+      context == null ? (byte)0 : context.getFlags(),
+      baggage,
+      debugId);
+  }
+
+  private Map<String, String> parseBaggageHeader(String header, Map<String, String> baggage) {
+    String[] parts = header.split("\\ *,\\ *");
+    for (int i = 0; i < parts.length; i++) {
```
vprithvi: minor nit: can be simplified to a range for loop because we don't use the outer index

`for (String part : header.split(" *, *")) {`

Also, are the regex escapes required here?

yurishkuro: I means `\s`, will change.

jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TextMapCodec.java
```diff
@@ -125,24 +125,34 @@ public JaegerSpanContext extract(TextMap carrier) {
...
+      context == null ? 0 : context.getTraceId(),
+      context == null ? 0 : context.getSpanId(),
+      context == null ? 0 : context.getParentId(),
+      context == null ? (byte)0 : context.getFlags(),
+      baggage,
+      debugId);
+  }
+
+  private Map<String, String> parseBaggageHeader(String header, Map<String, String> baggage) {
+    String[] parts = header.split("\\ *,\\ *");
```
vprithvi: Programmatically set baggage can have `,` or `=` in the baggage value while it seems adhoc baggage cannot.  While it is obvious, we should call out this limitation in documentation.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
```diff
@@ -53,7 +53,7 @@ protected JaegerSpanContext(
       String debugId,
       JaegerObjectFactory objectFactory) {
     if (baggage == null) {
-      throw new NullPointerException();
+      baggage = Collections.<String, String>emptyMap();
```
isaachier: +1

jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TextMapCodec.java
yurishkuro: That seems kind of obvious if you're asked to provide baggage in the `k1=v1, k2=v2` format.

PR #527, opened 2018-08-21
Implement TracerFactory

jpkrohling:

- See https://github.com/opentracing-contrib/java-tracerresolver/issues/30

- Implements the new `TracerFactory` interface, changing the current tracer resolver to get a tracer from this new class (so that we have only one place knowing how to create a tracer).

### DIFF
--- a/jaeger-tracerresolver/src/main/java/io/jaegertracing/tracerresolver/internal/JaegerTracerFactory.java
+++ b/jaeger-tracerresolver/src/main/java/io/jaegertracing/tracerresolver/internal/JaegerTracerFactory.java
@@ -16,16 +16,11 @@ package io.jaegertracing.tracerresolver.internal;
 
 import io.jaegertracing.Configuration;
 import io.jaegertracing.internal.JaegerTracer;
-import io.opentracing.Tracer;
 import io.opentracing.contrib.tracerresolver.TracerFactory;
 
 public class JaegerTracerFactory implements TracerFactory {
   @Override
-  public Tracer getTracer() {
-    return getJaegerTracer();
-  }
-
-  public JaegerTracer getJaegerTracer() {
+  public JaegerTracer getTracer() {
     return Configuration.fromEnv().getTracer();
   }
 }
--- a/jaeger-tracerresolver/src/main/java/io/jaegertracing/tracerresolver/internal/JaegerTracerFactory.java
+++ b/jaeger-tracerresolver/src/main/java/io/jaegertracing/tracerresolver/internal/JaegerTracerFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, Uber Technologies, Inc
+ * Copyright (c) 2018, The Jaeger Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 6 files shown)


isaachier:
I'm glad I ended up using `JaegerObjectFactory` instead of `JaegerTracerFactory` in #509. :)

jpkrohling:
codecov is lying: it claims that `JaegerTracerResolver#resolve` isn't being called, which is false. Running the tests in IntelliJ with code coverage, shows that it gets hit once:

image

I'm adding an explicit call to that method anyway, but codecov is lying!

yurishkuro:
Afaik codecov just works off the coverage files we send to it. Worth checking if the execution is being captured there first.

vprithvi:
@yurishkuro is correct here. You can inspect these files by clicking download on the codecov build view: https://codecov.io/gh/jaegertracing/jaeger-client-java/commit/c858eba8871b19269718525ab28b083e48446772/build

jpkrohling:
It shows the line as covered now, as I'm making an explicit call to the method, but it wasn't before.

jpkrohling:
I just tried locally with `./gradlew codeCoverageReport` and there's something strange there...

image

If I have a code like this in the test:
```java
    JaegerTracer tracer = ((JaegerTracerResolver) tracerResolver).resolve();
    assertNotNull(tracer);
```

I get a different coverage result:
image

I can't explain this, so, I'll just leave the test with the explicit call to make codecov happy. But it's still lying.

isaachier:
I happen to agree that I've had my fair share of issues with codecov. Don't know if it stems from the language tooling or the codecov analysis.

yurishkuro:
Still way better than coveralls

jpkrohling:
Apart from the code coverage discussion, is this OK to get merged?

jaeger-tracerresolver/src/main/java/io/jaegertracing/tracerresolver/internal/JaegerTracerFactory.java
```diff
@@ -0,0 +1,31 @@
...
+
+public class JaegerTracerFactory implements TracerFactory {
+  @Override
+  public Tracer getTracer() {
+    return getJaegerTracer();
+  }
+
+  public JaegerTracer getJaegerTracer() {
+    return Configuration.fromEnv().getTracer();
+  }
```
yurishkuro: I suggest dropping this method

jpkrohling: And leave the old resolver as it is, right?

yurishkuro: Don't know about the resolver, no comment. But this method should just be folded into the one above, it doesn't add anything useful.

jpkrohling: I simplified this a bit. As the resolver was also calling the `JaegerTracerFactory#getJaegerTracer()` method, I just changed the return object from `getTracer()` to be `JaegerTracer`. This way, the old resolver can still call it, ensuring that both the resolver and the factory yields similar tracers.

jpkrohling: Cool, there's a new "Resolve conversation" button here :)

PR #546, opened 2018-09-07
Close tracer on JVM shutdown

### LINKED ISSUE #544: jaeger-client-java does not automatically flush when program exits
## Requirement

I was trying to get the opentracing tutorial to work with the latest version of jaeger-client-java and it wasn't flushing the span to the jaeger server

## Problem

The span is not getting flushed to the jaeger server

Please see line 36/37 of
https://github.com/ackerleytng/opentracing-tutorial-debug/blob/master/java/src/main/java/lesson01/exercise/Hello.java

## Proposal

@jpkrohling suggested registering the tracer with a shutdown hook, to flush it when the JVM goes down

jpkrohling:

- Resolves #544

- Adds a shutdown hook to close the tracer before the JVM dies. I tested this with the same code as the reporter. When running the code *without* an explicit `tracer.close()`, I see this in the logs for the client:

```
12:35:51.066 [main] INFO io.jaegertracing.internal.reporters.LoggingReporter - Span reported: f9e218d52dc14279:f9e218d52dc14279:0:1 - say-hello
```

But nothing on the server side. Starting an all-in-one container with `--log-level=debug` confirms that nothing is received on that end.

With this PR applied, I do see the trace:

```
12:40:39.544 [main] INFO io.jaegertracing.internal.reporters.LoggingReporter - Span reported: 3ce5effda0536731:3ce5effda0536731:0:1 - say-hello
```

And on the server side: 

```json
{"level":"debug","ts":1536316839.6288192,"caller":"app/span_processor.go:104","msg":"Span written to the storage by the collector","trace-id":"3ce5effda0536731","span-id":"3ce5effda0536731"}
```

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
@@ -120,6 +120,14 @@ public class JaegerTracer implements Tracer, Closeable {
     }
     this.ipv4 = ipv4;
     this.tags = Collections.unmodifiableMap(tags);
+
+    // register this tracer with a shutdown hook, to flush the spans before the VM shuts down
+    Runtime.getRuntime().addShutdownHook(new Thread() {
+      @Override
+      public void run() {
+        JaegerTracer.this.close();
+      }
+    });
   }
 
   public String getVersion() {


ackerleytng:
Thanks! Looking forward to the next release!

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
```diff
@@ -120,6 +120,14 @@ protected JaegerTracer(JaegerTracer.Builder builder) {
     }
     this.ipv4 = ipv4;
     this.tags = Collections.unmodifiableMap(tags);
+
+    // register this tracer with a shutdown hook, to flush the spans before the VM shuts down
+    Runtime.getRuntime().addShutdownHook(new Thread() {
```
yurishkuro: what will this do to unit tests?

PR #548, opened 2018-09-10
Report java version as process tag

quaff:
In a large organization, there is numbers of services deployed base on different version jdk, Java version should be collected as built-in tag. 
It's better jaeger spark job and ui can provide a list view for specified tag (here is java.version), something like:

| Service  | Tag (java.version) |
| ------------- | ------------- |
| A | 1.8.0_181  |
| B | 1.8.0_181  |
| C | 9  |
| D | 10.0.2  |

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/Constants.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/Constants.java
@@ -50,6 +50,11 @@ public class Constants {
   public static final String JAEGER_CLIENT_VERSION_TAG_KEY = "jaeger.version";
 
   /**
+   * The name of the tag used to report java version.
+   */
+  public static final String JAVA_VERSION_TAG_KEY = "java.version";
+
+  /**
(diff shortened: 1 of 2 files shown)


yurishkuro:
I have to disagree, for two reasons:

- it's going to add massive amount of redundant data in the tracing backend. Presumably, an organization can get a breakdown of app->Java version by other means.

- if your organization still wants to go in this direction, you can easily add the tags to the tracer through other means, eg via env variables. But if we add it to the core lib, everyone will get this whether they want it or not.

quaff:
It's OK, I thought it would be a general feature. The point is the second section, I reported https://github.com/jaegertracing/jaeger-ui/issues/246

jpkrohling:
I'm closing this, as I think @yurishkuro has a great point :)

In any case, thanks for the PR! Keep sending your ideas, perhaps opening an issue on the repository first to discuss whether it would be merged.

Issue #547, opened 2018-09-09
Fall back to JAEGER_AGENT_HOST for JAEGER_SAMPLER_MANAGER_HOST_PORT before localhost


lw346:
## Requirement - what kind of business use case are you trying to solve?

Making deployment in a Kubernetes environment simpler, with less configuration needing to be applied to all pods - following DRY principles.

## Problem - what in Jaeger blocks you from solving the requirement?

In the [Jaeger Kubernetes](https://github.com/jaegertracing/jaeger-kubernetes) project, the agents are deployed as a `DaemonSet`, such that they are present on each Kubernetes node.  Each running pod then needs to use the downward API to extract the IP address of the node it is running on as the agent IP.

This results in an environment variable configuration looking like this:

```yaml
env:
- name: JAEGER_AGENT_HOST
  valueFrom:
    fieldRef:
      fieldPath: status.hostIP
```

If we are using the `RemoteSampler` configuration though, we also expect the agent to be proxying the HTTP sampler API from the collector (which we can derive from the default being `localhost:5778`).  Unlike the agent configuration however, this can only be overridden in a single environment variable: host + port, rather than separate host and port variables.

This means we need in addition to the above, an extra environment variable in each pod:

```yaml
env:
- name: JAEGER_AGENT_HOST
  valueFrom:
    fieldRef:
      fieldPath: status.hostIP
- name:  JAEGER_SAMPLER_MANAGER_HOST_PORT
  value: "$(JAEGER_AGENT_HOST):5778"
```

(using runtime substitution to interpolate the first environment variable).

## Proposal - what do you suggest to solve the problem or improve the existing situation?

It would be nice if we could use another fallback mechanism, eg.

1. Look for JAEGER_SAMPLER_MANAGER_HOST_PORT
1. Look for JAEGER_AGENT_HOST, default to port 5778
1. Fall back to localhost:5778

## Any open questions to address

This would, however, be a breaking change in behaviour that people may not be expecting.  If done in conjunction with https://github.com/jaegertracing/jaeger-client-java/issues/521 it could be described as an upgrade in behaviour as well as a name change.

This behaviour should be consistent across the various different client libraries.

Issue #521, opened 2018-08-20
Rename `JAEGER_SAMPLER_MANAGER_HOST_PORT` to ` JAEGER_CONFIG_MGR_HOST_PORT`


jpkrohling:
Same as https://github.com/jaegertracing/jaeger-client-go/issues/282, but for the Java client.

yurishkuro:
Replacing this with https://github.com/jaegertracing/jaeger/issues/1849

PR #554, opened 2018-09-15
Introduce JAEGER_CONFIG_MANAGER_ENDPOINT property

lw346:
- Issue https://github.com/jaegertracing/jaeger-client-java/issues/521
  - Rename `JAEGER_SAMPLER_MANAGER_HOST_PORT` to ` JAEGER_CONFIG_MGR_HOST_PORT`
- Issue https://github.com/jaegertracing/jaeger-client-java/issues/547
  - Fall back to JAEGER_AGENT_HOST for JAEGER_SAMPLER_MANAGER_HOST_PORT before localhost

Introduce a new property, `JAEGER_CONFIG_MANAGER_ENDPOINT`, that should be set to the URL of the Jaeger Config Manager sampling endpoint. If not set, fall back to the now-deprecated `JAEGER_SAMPLER_MANAGER_HOST_PORT` as the host and port of the URL.

If `JAEGER_AGENT_HOST` has been set as a property, further fall back to that value as the host of the URL (using the default port of 5778).

This brings the `SamplerConfiguration` in-line with the Go client.

See also:
 * https://github.com/jaegertracing/jaeger-client-go/issues/326
 * https://github.com/jaegertracing/jaeger-client-go/issues/282

Note that this does remove `SampleConfiguration#getManagerHostPort()` - the alternative would be to try to parse the URI provided and return the derived host and port.  Thoughts?

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/Configuration.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/Configuration.java
@@ -124,6 +124,11 @@ public class Configuration {
   public static final String JAEGER_SAMPLER_MANAGER_HOST_PORT = JAEGER_PREFIX + "SAMPLER_MANAGER_HOST_PORT";
 
   /**
+   * The config manager endpoint URL.
+   */
+  public static final String JAEGER_CONFIG_MANAGER_ENDPOINT = JAEGER_PREFIX + "CONFIG_MANAGER_ENDPOINT";
+
+  /**
    * The service name.
    */
   public static final String JAEGER_SERVICE_NAME = JAEGER_PREFIX + "SERVICE_NAME";
@@ -332,26 +337,36 @@ public class Configuration {
     private Number param;
 
     /**
-     * HTTP host:port of the sampling manager that can provide sampling strategy to this service.
+     * HTTP URL of the sampling manager that can provide sampling strategy to this service.
      * Optional.
      */
-    private String managerHostPort;
+    private String serverUrl;
 
     public SamplerConfiguration() {
     }
 
     public static SamplerConfiguration fromEnv() {
+      String configManagerEndpoint = getProperty(JAEGER_CONFIG_MANAGER_ENDPOINT);
+      if (configManagerEndpoint == null) {
+        if (getProperty(JAEGER_SAMPLER_MANAGER_HOST_PORT) != null) {
+          configManagerEndpoint = "http://" + getProperty(JAEGER_SAMPLER_MANAGER_HOST_PORT) + "/sampling";
+        } else if (getProperty(JAEGER_AGENT_HOST) != null) {
+          configManagerEndpoint = "http://" + getProperty(JAEGER_AGENT_HOST) + ":"
+                  + HttpSamplingManager.DEFAULT_SERVER_PORT + "/sampling";
+        }
+      }
+
       return new SamplerConfiguration()
           .withType(getProperty(JAEGER_SAMPLER_TYPE))
           .withParam(getPropertyAsNum(JAEGER_SAMPLER_PARAM))
-          .withManagerHostPort(getProperty(JAEGER_SAMPLER_MANAGER_HOST_PORT));
+          .withServerUrl(configManagerEndpoint);
     }
 
     // for tests
     Sampler createSampler(String serviceName, Metrics metrics) {
       String samplerType = stringOrDefault(this.getType(), RemoteControlledSampler.TYPE);
       Number samplerParam = numberOrDefault(this.getParam(), ProbabilisticSampler.DEFAULT_SAMPLING_PROBABILITY);
-      String hostPort = stringOrDefault(this.getManagerHostPort(), HttpSamplingManager.DEFAULT_HOST_PORT);
+      String serverUrl = stringOrDefault(this.getServerUrl(), HttpSamplingManager.DEFAULT_SERVER_URL);
 
       if (samplerType.equals(ConstSampler.TYPE)) {
         return new ConstSampler(samplerParam.intValue() != 0);
@@ -367,7 +382,7 @@ public class Configuration {
 
       if (samplerType.equals(RemoteControlledSampler.TYPE)) {
         return new RemoteControlledSampler.Builder(serviceName)
-            .withSamplingManager(new HttpSamplingManager(hostPort))
+            .withSamplingManager(new HttpSamplingManager(serverUrl))
             .withInitialSampler(new ProbabilisticSampler(samplerParam.doubleValue()))
             .withMetrics(metrics)
             .build();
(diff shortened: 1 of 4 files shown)


lw346:
Is there any feedback on this?

jaeger-core/src/main/java/io/jaegertracing/Configuration.java
```diff
@@ -332,26 +337,36 @@ public MetricsFactory getMetricsFactory() {
...
     public SamplerConfiguration() {
     }
 
     public static SamplerConfiguration fromEnv() {
+      String configManagerEndpoint = getProperty(JAEGER_CONFIG_MANAGER_ENDPOINT);
+      if (configManagerEndpoint == null) {
+        if (getProperty(JAEGER_SAMPLER_MANAGER_HOST_PORT) != null) {
+          configManagerEndpoint = "http://" + getProperty(JAEGER_SAMPLER_MANAGER_HOST_PORT) + "/sampling";
+        } else if (getProperty(JAEGER_AGENT_HOST) != null) {
+          configManagerEndpoint = "http://" + getProperty(JAEGER_AGENT_HOST) + ":"
```
pavolloffay: nit: `String.format()` might be cleaner here.

jaeger-core/src/main/java/io/jaegertracing/Configuration.java
```diff
@@ -332,26 +337,36 @@ public MetricsFactory getMetricsFactory() {
...
+    private String serverUrl;
 
     public SamplerConfiguration() {
     }
 
     public static SamplerConfiguration fromEnv() {
+      String configManagerEndpoint = getProperty(JAEGER_CONFIG_MANAGER_ENDPOINT);
+      if (configManagerEndpoint == null) {
+        if (getProperty(JAEGER_SAMPLER_MANAGER_HOST_PORT) != null) {
+          configManagerEndpoint = "http://" + getProperty(JAEGER_SAMPLER_MANAGER_HOST_PORT) + "/sampling";
```
pavolloffay: nit: `String.format()` might be cleaner here.

jaeger-core/src/main/java/io/jaegertracing/Configuration.java
```diff
@@ -398,8 +413,17 @@ public SamplerConfiguration withParam(Number param) {
       return this;
     }
 
+    /**
+     * @deprecated Use {@link #withServerUrl(String)} instead
+     */
+    @Deprecated
     public SamplerConfiguration withManagerHostPort(String managerHostPort) {
-      this.managerHostPort = managerHostPort;
+      this.serverUrl = "http://" + managerHostPort + "/sampling";
```
pavolloffay: nit: `String.format()` might be cleaner here.

jaeger-core/src/main/java/io/jaegertracing/internal/samplers/HttpSamplingManager.java
```diff
@@ -28,20 +28,23 @@
 
 @ToString
 public class HttpSamplingManager implements SamplingManager {
-  public static final String DEFAULT_HOST_PORT = "localhost:5778";
-  private final String hostPort;
+  public static final String DEFAULT_SERVER_HOST = "localhost";
```
pavolloffay: It does not seem to be used. Only port is used in configuration

jaeger-core/src/main/java/io/jaegertracing/internal/samplers/HttpSamplingManager.java
```diff
@@ -28,20 +28,23 @@
...
+   * This constructor expects running sampling manager on {@link #DEFAULT_SERVER_URL}.
    */
   public HttpSamplingManager() {
-    this(DEFAULT_HOST_PORT);
+    this(DEFAULT_SERVER_URL);
   }
 
-  public HttpSamplingManager(String hostPort) {
-    this.hostPort = hostPort != null ? hostPort : DEFAULT_HOST_PORT;
+  public HttpSamplingManager(String serverUrl) {
```
pavolloffay: It seems like a breaking change. 

We can keep `hostport`. `HttpBaggageRestrictionManagerProxy` also accepts `hostport`. Only HTTP sender accepts full URL e.g. `http://jaeger-collector:111/api/traces`. 

Host port is easier to configure. Not sure if there are any advantages of using full URL.

jaeger-core/src/main/java/io/jaegertracing/Configuration.java
```diff
@@ -123,6 +123,11 @@
    */
   public static final String JAEGER_SAMPLER_MANAGER_HOST_PORT = JAEGER_PREFIX + "SAMPLER_MANAGER_HOST_PORT";
 
+  /**
+   * The config manager endpoint URL.
+   */
+  public static final String JAEGER_CONFIG_MANAGER_ENDPOINT = JAEGER_PREFIX + "CONFIG_MANAGER_ENDPOINT";
```
pavolloffay: We should use `CONFIG_MANAGER_HOST_PORT` as it is also used by baggage restriction. host port is also easier to configure.

jaeger-core/src/main/java/io/jaegertracing/internal/samplers/HttpSamplingManager.java
lw346: Agreed - the primary driver of using ENDPOINT was to keep this in-line with the Go library.  If that's not a concern, then this is something I can change.

jaeger-core/src/main/java/io/jaegertracing/Configuration.java
lw346: If we go with HOST_PORT, shouldn't need to use any sort of string manipulation; pending confirmation that we're happy to diverge from the Go lib.

jpkrohling:
@lw346 would you still be interested in this one?

lw346:
There are unanswered outstanding concerns about diverging from the Go lib in behaviour, but it's been long enough now that it's probably too late to be concerned about the divergence :man_shrugging: 

I don't think I have time to look at this myself.

jpkrohling:
@lw346 I apologize for letting this one reach this state without getting merged.

yurishkuro:
Let's resolve this first: https://github.com/jaegertracing/jaeger/issues/1849

chlunde:
Is the design clear now that #1849 is resolved? Solving #547 for daemonset-deployments with Java would be neat!

yurishkuro:
Per https://github.com/jaegertracing/jaeger/issues/1849, we decided to use `JAEGER_SAMPLING_ENDPOINT`.

houstonheat:
Any updates here? Jaeger-collector is now exposing sampling API endpoint too and we are really want to use it 🌚

PR #555, opened 2018-09-18
Fix repeated resolving of sender in configuration

pavolloffay:


jaeger-core/src/main/java/io/jaegertracing/Configuration.java
```diff
@@ -635,8 +635,7 @@ public Sender getSender() {
       if (null != sender) {
         return sender;
       }
-
-      return SenderResolver.resolve(this);
+      return (sender = SenderResolver.resolve(this));
```
yurishkuro: nit:
```
if sender == null {
    sender = resolve()
}
return sender
```

Issue #390, opened 2018-04-11
Release failures


pavolloffay:
Travis build: https://travis-ci.org/jaegertracing/jaeger-client-java/jobs/365086226#L1333 

Errors when repsitory is closed manually:
```
failureMessage | Missing SHA1: '/io/jaegertracing/jaeger-thrift/0.27.0-RC1/jaeger-thrift-0.27.0-RC1-sources.jar.sha1'
-- | --
failureMessage | Missing MD5: '/io/jaegertracing/jaeger-thrift/0.27.0-RC1/jaeger-thrift-0.27.0-RC1-sources.jar.md5'
failureMessage | Missing SHA1: '/io/jaegertracing/jaeger-thrift/0.27.0-RC1/jaeger-thrift-0.27.0-RC1.pom.sha1'
failureMessage | Missing MD5: '/io/jaegertracing/jaeger-thrift/0.27.0-RC1/jaeger-thrift-0.27.0-RC1.pom.md5'
failureMessage | Missing SHA1: '/io/jaegertracing/jaeger-core/0.27.0-RC1/jaeger-core-0.27.0-RC1.jar.sha1'
failureMessage | Missing MD5: '/io/jaegertracing/jaeger-core/0.27.0-RC1/jaeger-core-0.27.0-RC1.jar.md5'
failureMessage | Missing SHA1: '/io/jaegertracing/jaeger-crossdock/0.27.0-RC1/jaeger-crossdock-0.27.0-RC1.jar.sha1'
failureMessage | Missing MD5: '/io/jaegertracing/jaeger-crossdock/0.27.0-RC1/jaeger-crossdock-0.27.0-RC1.jar.md5'
```

pavolloffay:
The publish failed even after #391  https://travis-ci.org/jaegertracing/jaeger-client-java/jobs/365101107#L1331

pavolloffay:
The problem might be that publish creates 7 staging repositories. @jpkrohling  @yurishkuro have you seen this before?

jpkrohling:
I've seen it before when releasing via Travis. If you checkout the tag locally and run `./gradlew upload`, it usually works fine. Just make sure you clean up any garbage created by Travis (ie: close/drop those staging repositories).

pavolloffay:
I have tried the build 3 times, always it fails. I wondering if it even works sometimes automatically.

jpkrohling:
[...]

PR #558, opened 2018-10-15
Define nexus staging id explicitly

pavolloffay:
Related to https://github.com/jaegertracing/jaeger-client-java/issues/390#issuecomment-395428726

Attempt to fix release process on travis. Once this is merged I will try to cut patch release.

Signed-off-by: Pavol Loffay 🦄 <ploffay@redhat.com>

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -84,6 +84,7 @@ subprojects {
             // Dealing with error "Wrong number of received repositories in state 'open'" (http://bit.ly/2ybracm)
             numberOfRetries = 50
             delayBetweenRetriesInMillis = 3000
+            stagingProfileId = getVersionForBuild()
         }
 
         uploadArchives.finalizedBy closeAndReleaseRepository


build.gradle
```diff
@@ -84,6 +84,7 @@ subprojects {
             // Dealing with error "Wrong number of received repositories in state 'open'" (http://bit.ly/2ybracm)
             numberOfRetries = 50
             delayBetweenRetriesInMillis = 3000
+            stagingProfileId = getVersionForBuild()
```
jpkrohling: I don't think this is safe enough. If one build fails, the next build for the same version will use the old staging repository, which will probably contain some artifacts already, blocking new artifacts from being uploaded (the server returns a "Conflict" state, IIRC).

Ideally, this would be something like:

`"jaeger-java-client-%s-%s", version, timestamp`

This way, it's easy to find in the UI (the part up to the version) and it's almost guaranteed to be unique (the timestamp part).

pavolloffay: I think it's a good idea to add the timestamp. Also, I would encourage people dropping open staging repos after failed release. We should add that in release.md.

PR #591, opened 2019-02-22
Handle all exceptions during queue processing

### LINKED ISSUE #590: Warn/Crash when RemoteReporter QueueProcessor crashes
We are using the jaeger-client from Clojure and due to programmer error we got the following stack trace:

```
java.lang.ClassCastException: clojure.lang.Keyword cannot be cast to java.lang.String 
	at io.jaegertracing.thrift.internal.reporters.protocols.JaegerThriftSpanConverter.buildTags(JaegerThriftSpanConverter.java:98)
	at io.jaegertracing.thrift.internal.reporters.protocols.JaegerThriftSpanConverter.buildLogs(JaegerThriftSpanConverter.java:80)
	at io.jaegertracing.thrift.internal.reporters.protocols.JaegerThriftSpanConverter.convertSpan(JaegerThriftSpanConverter.java:58)
	at io.jaegertracing.thrift.internal.senders.ThriftSender.append(ThriftSender.java:55) 
	at io.jaegertracing.internal.reporters.RemoteReporter$AppendCommand.execute(RemoteReporter.java:140) 
	at io.jaegertracing.internal.reporters.RemoteReporter$QueueProcessor.run(RemoteReporter.java:173) 
	at java.lang.Thread.run(Thread.java:748)
```

It would be nice if we jaegertracing could either log warnings or crash when this (or something similar) happens, instead of silently stopping to work because there is no QueueProcessor. We only caught this because we set a default UncaughExceptionHandler for all threads.

dignati:
Fixes #590 

This reverts a change that made QueueProcessor only handle InterruptedException. Instead we want to ignore and log all exceptions because otherwise the QueueProcessor thread would fail silently.

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/reporters/RemoteReporter.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/reporters/RemoteReporter.java
@@ -177,7 +177,7 @@ public class RemoteReporter implements Reporter {
           } catch (SenderException e) {
             metrics.reporterFailure.inc(e.getDroppedSpanCount());
           }
-        } catch (InterruptedException e) {
+        } catch (Exception e) {
           log.error("QueueProcessor error:", e);
           // Do nothing, and try again on next span.
         }


dignati:
I'm not sure this needs tests. If it does I'd be grateful for some suggestions on how to add a meaningful test.

yurishkuro:
I'll merge, but if you'd like to add a test it would prevent the regression in the future. The easiest way to do that is by using a mock Sender that throws an exception.

PR #609, opened 2019-03-29
Concurrency improvements to RemoteControlledSampler

### LINKED ISSUE #608: RemoteControlledSampler.sample() is a bottleneck
## Requirement

Tracing should not be a bottleneck in multithreaded apps.

`RemoteControlledSampler.sample()` is [synchronized](https://github.com/jaegertracing/jaeger-client-java/blob/master/jaeger-core/src/main/java/io/jaegertracing/internal/samplers/RemoteControlledSampler.java#L140).

This is causing a major slowdown in multithreaded applications:

sample

1. Remove unneeded synchronization.
2. Synchronize on non-public objects (instead of `this` or `.class`) where synchronization is unavoidable.
3. Clean up `.equals()` - avoid synchronized blocks (and remove the read locks?)
4. Explain the need for every remaining synchronization (in the code comments).

What is the purpose of `RemoteControlledSampler.getLock()`? Some JavaDoc would be helpful there.

yborovikov:
This removes synchronization from RemoteControlledSampler and makes RateLimiter thread-safe.

As described in #608 Synchronization in `RemoteControlledSampler` was causing severe performance degradation in multi-threaded applications.

`RateLimiter.checkBalance()` implementation was not thread-safe and didn't correctly initialize `lastTick`, causing non-deterministic behavior for the first invocation of `.checkBalance()`.

Synchronization removed from `RemoteControlledSampler`.

Internal state of `RateLimiter` is now stored in `AtomicLong`s, with `.checkBalance()` modified to use an approach similar to `AtomicLong.getAndAccumulate()` - to prevent internal state corruption in concurrent flows.

Fixes #608

### DIFF
--- a/jaeger-core/src/test/java/io/jaegertracing/internal/utils/RateLimiterTest.java
+++ b/jaeger-core/src/test/java/io/jaegertracing/internal/utils/RateLimiterTest.java
@@ -14,15 +14,19 @@
 
 package io.jaegertracing.internal.utils;
 
-import static junit.framework.TestCase.assertFalse;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
 import io.jaegertracing.internal.clock.Clock;
+
+import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
 import org.junit.Test;
 
 public class RateLimiterTest {
-  RateLimiter limiter;
 
   private static class MockClock implements Clock {
 
@@ -127,4 +131,59 @@ public class RateLimiterTest {
     assertTrue(limiter.checkCredit(1.0));
     assertFalse(limiter.checkCredit(1.0));
   }
+
+  /**
+   * Validates rate limiter behavior with {@link System#nanoTime()}-like (non-zero) initial nano ticks.
+   */
+  @Test
+  public void testRateLimiterInitial() {
+    MockClock clock = new MockClock();
+    clock.timeNanos = TimeUnit.MILLISECONDS.toNanos(-1_000_000);
+    RateLimiter limiter = new RateLimiter(1000, 100, clock);
+
+    assertTrue(limiter.checkCredit(100)); // consume initial (max) balance
+    assertFalse(limiter.checkCredit(1));
+
+    clock.timeNanos += TimeUnit.MILLISECONDS.toNanos(49); // add 49 credits
+    assertFalse(limiter.checkCredit(50));
+
+    clock.timeNanos += TimeUnit.MILLISECONDS.toNanos(1); // add one credit
+    assertTrue(limiter.checkCredit(50)); // consume accrued balance
+    assertFalse(limiter.checkCredit(1));
+
+    clock.timeNanos += TimeUnit.MILLISECONDS.toNanos(1_000_000); // add a lot of credits (max out balance)
+    assertTrue(limiter.checkCredit(1)); // take one credit
+
+    clock.timeNanos += TimeUnit.MILLISECONDS.toNanos(1_000_000); // add a lot of credits (max out balance)
+    assertFalse(limiter.checkCredit(101)); // can't consume more than max balance
+    assertTrue(limiter.checkCredit(100)); // consume max balance
+    assertFalse(limiter.checkCredit(1));
+  }
+
+  /**
+   * Validates concurrent credit check correctness.
+   */
+  @Test
+  public void testRateLimiterConcurrency() {
+    int numWorkers = ForkJoinPool.getCommonPoolParallelism();
+    int creditsPerWorker = 1000;
+    MockClock clock = new MockClock();
+    RateLimiter limiter = new RateLimiter(1, numWorkers * creditsPerWorker, clock);
+
+    AtomicInteger count = new AtomicInteger();
+    for (int w = 0; w < numWorkers; ++w) {
+      ForkJoinPool.commonPool().execute(() -> {
+        for (int i = 0; i < creditsPerWorker * 2; ++i) {
+          if (limiter.checkCredit(1)) {
+            count.getAndIncrement(); // count allowed operations
+          }
+        }
+      });
+    }
+    ForkJoinPool.commonPool().awaitQuiescence(1000, TimeUnit.MILLISECONDS);
+
+    assertEquals("Exactly the allocated number of credits must be consumed", numWorkers * creditsPerWorker,count.get());
+    assertFalse(limiter.checkCredit(1));
+  }
+
 }
--- a/jaeger-core/src/test/java/io/jaegertracing/internal/utils/RateLimiterTest.java
+++ b/jaeger-core/src/test/java/io/jaegertracing/internal/utils/RateLimiterTest.java
@@ -180,7 +180,7 @@ public class RateLimiterTest {
         }
       });
     }
-    ForkJoinPool.commonPool().awaitQuiescence(1000, TimeUnit.MILLISECONDS);
+    ForkJoinPool.commonPool().awaitQuiescence(1, TimeUnit.SECONDS);
 
     assertEquals("Exactly the allocated number of credits must be consumed", numWorkers * creditsPerWorker,count.get());
     assertFalse(limiter.checkCredit(1));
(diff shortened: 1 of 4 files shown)


yurishkuro:
This is a tricky change, I haven't spent enough time reviewing it yet. High level impressions though:
  * The rate limiter is being made thread-safe by this PR, we should reflect it in the class javadoc.
  * The way it is being made thread-safe (via atomic longs) is tricky, I had a hard time following whether it introduces more race conditions. It would be useful to see a benchmark of whether atomic longs are actually saving time vs. just synchronizing the rate limiter update methods.
  * The change to the remote sampler makes a major assumption that the `update` methods are thread-safe on every possible sampler implementation. It's not obvious at all if that's the case. Also, even if update() methods are thread-safe, it seems like the equals() method now becomes racy.

I would like to see documentation added to the expectation of thread-safety on different samplers (this was not a requirement previously). Alternatively, we should consider if pushing thread-safety to individual samplers is the right approach. The alternative could be a "copy on write" implementation of the top-level adaptiveSampler.update() method where each individual sampler is asked to update() itself to new parameters, and it internally decides whether to return `this` (in case of no changes) or a new instance. However, this approach still has race conditions for rate limiting samplers which might be concurrently asked isSampled() and update their internal state.

yborovikov:
Thank you for reviewing the change!

> * The rate limiter is being made thread-safe by this PR, we should reflect it in the class javadoc.

Are there any classes in Jaeger that are _not_ expected to be thread-safe? If so, I'd suggest marking _those_ as such, not the other way around. Using `@javax.annotation.concurrent.[Not]ThreadSafe` might be a good way to express these contracts.

> * The way it is being made thread-safe (via atomic longs) is tricky, I had a hard time following whether it introduces more race conditions. It would be useful to see a benchmark of whether atomic longs are actually saving time vs. just synchronizing the rate limiter update methods.

Most of the trickery comes from the need to convert double arithmetic to integer-based. I'd suggest changing the API to use integer quotas. Given it's an internal implementation class and the only use case is `.checkCredit(1.0)` that should be easy. But not strictly necessary.

Another bit than might feel tricky is a standard optimistic locking over CAS pattern, very similar to Java 8+'s `AtomicLong.accumulateAndGet()` implementation (as mentioned in the PR).

The exising (and added concurrent) unit tests confirm the correcness of the behavior is preserved.

Concurrent performance wise, rule of thumb would be for `volatile` to be typically faster than `synchronized` - fairly understandable when one takes the respective granularity of happens-before guarantees. There's plenty of articles and benchmarks on the topic, here's one of the first ones - [from 2004 by Brian Goetz](https://www.ibm.com/developerworks/library/j-jtp11234/).

Of course, any specific use case can be unique (and the real comparisons should be derived from observing the real-life, production behavior). Anyway, here's some JMH results I got from benchmarking synchronized and atomic implementations of `RateLimiter` (8 threads, ~10% time spent in `.checkCredit()`):
```
Benchmark                             Mode  Cnt         Score        Error  Units
SynchronizedRateLimiter.baseline     thrpt  100  17299664.806 ± 250785.751  ops/s
SynchronizedRateLimiter.checkCredit  thrpt  100   4743227.226 ± 141156.036  ops/s
AtomicRateLimiter.checkCredit        thrpt  100  13088194.369 ± 154981.298  ops/s
```
`.baseline` is original (no locking, not thread-safe) implementation - for comparison.
This is, obviously, still a synthetic benchmark - and, personally, I expect way more of a difference in a synchronized `.checkCredit()` performance in an efficient concurrent application.

> * The change to the remote sampler makes a major assumption that the `update` methods are thread-safe on every possible sampler implementation. It's not obvious at all if that's the case. Also, even if update() methods are thread-safe, it seems like the equals() method now becomes racy.

I checked every implementation in the project (that's how I came to fixing `RateLimiter` in this PR). There are some improvements that can be made (e.g. switching `PerOperationSampler` from `syncronized` to using `ConcurrentHashMap`) but overall it looks clean. Adding concurrent behavior tests sounds like a reasonable next step.

Same applies to `.equals()` - granted, I do not fully understand the logic/purpose behind some of its implementations. Anyway, `.equals()` on a mutable object is not guaranteed to return the same result for the same argument over time.

> I would like to see documentation added to the expectation of thread-safety on different samplers (this was not a requirement previously).

I believe it was - given that many of the samplers (e.g. `RateLimitingSampler`) are instantiated by `Configuration` and are directly wired into `Tracer` - there seems to be no expectation that every sampler will be wrapped by a `RemoteControlledSampler`  (to guarantee thread-safety?).

> Alternatively, we should consider if pushing thread-safety to individual samplers is the right approach.

Yes, absolutely.

> The alternative could be a "copy on write" implementation of the top-level adaptiveSampler.update() method where each individual sampler is asked to update() itself to new parameters, and it internally decides whether to return `this` (in case of no changes) or a new instance. However, this approach still has race conditions for rate limiting samplers which might be concurrently asked isSampled() and update their internal state.

Thanks again for your review, comments, and questions!

jpkrohling:
LGTM, but I wonder if this is sufficient to fix the behavior you were seeing as part of #608. Could you share a memory profile similar to what you have in the original issue? This would help evaluate the gains from this PR.

Additionally, would you be able to publish the JMH tests in a public repository? This would be very useful for us in the future.

yborovikov:
I updated `RateLimiter` with a simpler implementation that uses a single `AtomicLong` to track spent (and wasted - due to exceeding max balance) credits (in nano ticks). Hopefully, it's easier to comprehend, too.

This approach yields performance on par with the baseline (or, possibly, better - due to less floating point operations and main memory writes):
```
Benchmark                             Mode  Cnt         Score        Error  Units
SynchronizedRateLimiter.baseline     thrpt  100  16462034.465 ± 402504.035  ops/s
SynchronizedRateLimiter.checkCredit  thrpt  100   4744500.578 ±  88917.828  ops/s
AtomicRateLimiter.checkCredit        thrpt  100  13292796.354 ± 313919.827  ops/s
AtomicRateLimiter1.checkCredit       thrpt  100  17130924.009 ± 238918.453  ops/s
```

yborovikov:
> LGTM, but I wonder if this is sufficient to fix the behavior you were seeing as part of #608. Could you share a memory profile similar to what you have in the original issue? This would help evaluate the gains from this PR.

I updated our internal code to work around #608 by moving all (or most of) the sampling to a single thread. This resulted in much healthier threads:
sample-fixed
Arguably, getting rid of the `synchronized` bottleneck should produce a similar effect (except that `PerOperationSampler` is still heavily synchronized...) - but I haven't tried it in a real (production) environment.

> Additionally, would you be able to publish the JMH tests in a public repository? This would be very useful for us in the future.

Let me see if I can remove internal build scaffolding from my JMH project - this should be fairly easy.

yborovikov:
I ran the `RateLimiter` benchmark (with 100% time spent in `.checkCredit()`) on a 32(logical)-core Linux server for different numbers of threads.
Here's the resulting graph (vertical axis: ops/s, log scale; horizontal axis: number of threads, log scale):
32-core-linux-benchmark
- `baseline` (green) is the original (not thread-safe) implementation.
- `_synchronized` (blue) is the original with synchronization added.
- `atomic` (orange) is using a single `AtomicLong` to store the state (this PR).

I'm fairly satisfied with the state of this PR and don't have any further changes planned.

Please take some time to review the code and let me know if something needs to be fixed.

jaeger-core/src/test/java/io/jaegertracing/internal/utils/RateLimiterTest.java
```diff
@@ -127,4 +131,59 @@ public void testRateLimiterMaxBalance() {
...
+    for (int w = 0; w < numWorkers; ++w) {
+      ForkJoinPool.commonPool().execute(() -> {
+        for (int i = 0; i < creditsPerWorker * 2; ++i) {
+          if (limiter.checkCredit(1)) {
+            count.getAndIncrement(); // count allowed operations
+          }
+        }
+      });
+    }
+    ForkJoinPool.commonPool().awaitQuiescence(1000, TimeUnit.MILLISECONDS);
```
jpkrohling: Or `(1, TimeUnit.SECONDS)` :)

jpkrohling:
@yurishkuro are there any remaining concerns about this one?

pavolloffay:
I have talked to @jpkrohling that we should merge this PR. It's been a long time since it was approved.

william-tran:
@yborovikov thanks for this. 

>There are some improvements that can be made (e.g. switching PerOperationSampler from syncronized to using ConcurrentHashMap) but overall it looks clean. Adding concurrent behavior tests sounds like a reasonable next step.

I was looking at the PerOperationSampler code and the `public synchronized SamplingStatus sample()`  popped out at me. This PR fixed application wide contention on calls to RemoteControlledSampler.sample, but given PerOperationSampler's synchronized sample method, there is contention per operation. Did you end up seeing this as a result?

yborovikov:
Due to the workaround on our side I lost the ability to "reproduce" the contention in production environment at that time and didn't (need to) run any further performance tests.
The fix to `PerOperationSampler` seemed to be obvious (yet exact gains would depend on concrete customer app patterns).

PR #613, opened 2019-04-18
Binary format implementation

ColinSullivan1:
Hello from the NATS team!

We're adding a reference architecture for using the OpenTracing API over Jaeger in [NATS](https://nats.io).  We've completed a NATS Open Tracing golang implementation, found [here](https://github.com/nats-io/not.go), and want to provide a java implementation as it's one of our more popular client languages and has been specifically requested.  I noticed there was no binary format support, so added some here.  It is wire compatible with the Jaeger golang client binary format.

I'd be happy to discuss on a call, zoom, or google hangouts if that would help.

Signed-off-by: Colin Sullivan <colin@synadia.com>

- There is no binary format support in the Java client

- Added binary format support, with tests and descriptive comments.  I've mirrored what the go client has done, and have verified compatibility.


jaeger-core/src/main/java/io/jaegertracing/internal/propagation/BinaryCodec.java
```diff
@@ -0,0 +1,270 @@
...
+
+    // Now we have a stream and a size, and we'll copy it into the byte
+    // buffer.
+    int size = stream.size();
+    ByteBuffer buf = carrier.injectionBuffer(size);
+
+    // Java defaults to big endian (network order), but enforce it just
+    // in case the carrier set the wrong byte order before passing it in.
+    if (buf.order() != ByteOrder.BIG_ENDIAN) {
+      throw new IllegalStateException("Carrier byte order must be big endian.");
```
jpkrohling: Out of curiosity: did you run into this in practice? This check seems strangely too specific :-)

jaeger-core/src/test/java/io/jaegertracing/ConfigurationTest.java
```diff
@@ -257,6 +261,25 @@ public void testPropagationJaegerAndB3() {
...
+  @Test
+  public void testPropagationBinary() {
+    System.setProperty(Configuration.JAEGER_PROPAGATION, "jaeger");
+    System.setProperty(Configuration.JAEGER_SERVICE_NAME, "Test");
+
+    long traceIdLow = 1234L;
+    long spanId = 5678L;
+
+    TestBinaryCarrier buffer = new TestBinaryCarrier();
+    JaegerSpanContext spanContext = new JaegerSpanContext(0, traceIdLow, spanId, 0, (byte)0);
```
jpkrohling: Do we have boundary tests here? If so, it would be nice to have them for the binary codec as well.

jaeger-core/src/test/java/io/jaegertracing/internal/AdhocHeadersTest.java
```diff
@@ -27,6 +27,7 @@
 import io.opentracing.propagation.TextMap;
 import io.opentracing.propagation.TextMapAdapter;
 
+import java.nio.ByteBuffer;
```
jpkrohling: Perhaps a left-over from a local test?

jaeger-core/src/main/java/io/jaegertracing/internal/propagation/BinaryCodec.java
ColinSullivan1: I didn't in this case, although did notice that ByteBuffer supports specifying byte order.  I figured it'd be an extremely difficult issue to debug if someone someday implemented a carrier that set it to little endian, so figured this might save someone some time in the future.  ;)

jaeger-core/src/test/java/io/jaegertracing/internal/AdhocHeadersTest.java
ColinSullivan1: Ah, it is, I'll remove it.  Thanks!

jaeger-core/src/test/java/io/jaegertracing/ConfigurationTest.java
ColinSullivan1: Do you mean boundaries as in max values for trace ids, spanIds, etc?  I didn't see any elsewhere (may have missed them), but can add them to the binary codec if you'd like.

ColinSullivan1:
@jpkrohling Would you have guidance on the test, or is this good to go?  Thanks!

jaeger-core/src/test/java/io/jaegertracing/ConfigurationTest.java
jpkrohling: Don't worry about this test for now: if we get an ID which is bigger than the max value, it would also fail for other codecs. I was also a bit concerned about parsing failures (baggage keys without values, for instance) , but I believe this first implementation is already a great improvement.

ColinSullivan1:
Sounds good - thanks @jpkrohling!

PR #618, opened 2019-04-29
Upgrade gradle to 5.0 and use new maven-publish

pavolloffay:


pavolloffay:
I have released 0.35.0 locally with this config, I will merge and cut 0.35.1 to test it out via travis.

yurishkuro:
does it make the release process more reliable?

pavolloffay:
It should do, the problem was that travis is changing public IPs during the build - now there is a plugin which creates staging repo with an ID to overcome this problem.

the build failed though I will have a look

Issue #614, opened 2019-04-25
Release 0.35.0 based on OpenTracing 0.32.0


pavolloffay:
Master now contains an upgrade to OpenTracing 0.32.0. Is there anything blocking the release or can we release?

yurishkuro:
+1  to release

pavolloffay:
@yurishkuro would you have time to look at https://github.com/jaegertracing/jaeger-client-java/pull/609? I will cut the release early next week.

pavolloffay:
I am going to cut it now

pavolloffay:
The artifact has been published https://repo.maven.apache.org/maven2/io/jaegertracing/jaeger-core/0.35.0/

Fox32:
@pavolloffay  I just noticed that it is also required to upgrade [`java-tracerresolver`](https://github.com/jaegertracing/jaeger-client-java/blob/master/build.gradle) to version [0.1.6](https://github.com/opentracing-contrib/java-tracerresolver/releases), otherwise there is still a dependency to 0.31.0 of OpenTracing.

PR #620, opened 2019-04-29
Update java-tracerresolver to version 0.1.6

Fox32:
Version 0.1.5 of java-tracerresolver still dependes on OpenTracing 0.31.0.

Signed-off-by: Oliver Sand <oliver.sand@sda-se.com>

-  java-tracerresolver 0.1.5 still depends on OpenTracing 0.31.0, but we would like to use 0.32.0, see my comment in #614

- Change java-tracerresolver to version 0.1.6


Fox32:
The first Travis build failed, but now it works.

PR #621, opened 2019-05-06
Fix nexus-publish config

marcphilipp:

Prior to this commit the configuration of the `nexus-publish` plugin was ineffective because it was only applied to the root project instead of all subprojects.

Apply the nexus-publish plugin to all subprojects instead of the root
project, remove the explicitly configured repository and the default
server URLs.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,7 +14,7 @@ plugins {
     id 'maven-publish'
     id 'signing'
     id 'io.codearte.nexus-staging' version '0.20.0'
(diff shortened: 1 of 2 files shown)


pavolloffay:
Sweet this worked well locally on 0.35.2. I will try from travis.

pavolloffay:
Awesome release from travis worked - 0.35.3

marcphilipp:
Glad it worked! 🙂

pavolloffay:
thanks @marcphilipp !

PR #622, opened 2019-05-06
Simplify release process

### LINKED ISSUE #58: Automate release process
Current release process is documented in the readme and is fairly painful. The most ideal process would be to push a tag and have the release happen automatically (but I am not clear what happens to the changelog in this case). Using a make command would also be acceptable if it can do a bunch of changes in a row like bump the version, publish, bump version to snapshot again, and then automatically close & release the repo from staging to maven via sonatype.

pavolloffay:
Resolves #58 

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -6,6 +6,7 @@ The release process consists of these steps:
   1. Create a pull request with:
      * Add an entry to [`CHANGELOG`](CHANGELOG.md) with changes since the last release
      * add a new section to the CHANGELOG with that version as `(unreleased)` and a bullet point `- Nothing yet`
+     * bump the version in `build.gradle` to `major.minor.(patch+1)-SNAPSHOT` eg. `0.20.1-SNAPSHOT` - notice the `SNAPSHOT`!
   1. Commit your `CHANGELOG` changes as `git commit -m "Preparing release 0.20.0" -s`
   1. Create and push tag with the new version `git tag v0.20.0 && git push origin v0.20.0`
   1. Once the *tag* build finishes in Travis, the artifacts should have been uploaded to Sonatype staging,
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -6,7 +6,7 @@ The release process consists of these steps:
   1. Create a pull request with:
      * Add an entry to [`CHANGELOG`](CHANGELOG.md) with changes since the last release
      * add a new section to the CHANGELOG with that version as `(unreleased)` and a bullet point `- Nothing yet`
-  1. Commit your `CHANGELOG` changes
+  1. Commit your `CHANGELOG` changes as `git commit -m "Preparing release 0.20.0" -s`
   1. Create and push tag with the new version `git tag v0.20.0 && git push origin v0.20.0`
   1. Once the *tag* build finishes in Travis, the artifacts should have been uploaded to Sonatype staging,
      the staging repository closed, and the artifacts on the way to Maven Central (it takes 20min+ to get there).
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -4,19 +4,15 @@
 
 The release process consists of these steps:
   1. Create a pull request with:
-     * Change the version in [`build.gradle`](build.gradle) to match the release version e.g. `0.20.0`
      * Add an entry to [`CHANGELOG`](CHANGELOG.md) with changes since the last release
+     * add a new section to the CHANGELOG with that version as `(unreleased)` and a bullet point `- Nothing yet`
   1. Commit your `CHANGELOG` changes
   1. Create and push tag with the new version `git tag v0.20.0 && git push origin v0.20.0`
   1. Once the *tag* build finishes in Travis, the artifacts should have been uploaded to Sonatype staging,
      the staging repository closed, and the artifacts on the way to Maven Central (it takes 20min+ to get there).
      In case of failures it is safe to retry by restarting the maven-deploy job. If it keeps
-     failing, sometimes it may be necessary to close the staging repository manually.
+     failing, sometimes it may be necessary to close/drop the staging repository manually.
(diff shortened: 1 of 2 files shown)


RELEASE.md
```diff
@@ -4,19 +4,15 @@
...
+     * add a new section to the CHANGELOG with that version as `(unreleased)` and a bullet point `- Nothing yet`
+  1. Commit your `CHANGELOG` changes as `git commit -m "Preparing release 0.20.0" -s`
   1. Create and push tag with the new version `git tag v0.20.0 && git push origin v0.20.0`
   1. Once the *tag* build finishes in Travis, the artifacts should have been uploaded to Sonatype staging,
      the staging repository closed, and the artifacts on the way to Maven Central (it takes 20min+ to get there).
      In case of failures it is safe to retry by restarting the maven-deploy job. If it keeps
-     failing, sometimes it may be necessary to close the staging repository manually.
+     failing, sometimes it may be necessary to close/drop the staging repository manually.
   1. Create a release on GitHub for the new tag. Use the changes from the `CHANGELOG` as the description.
-  1. Once the artifacts are available on Maven
```
yurishkuro: It seems this would cause snapshots to report the official version as Jaeger.version span tag.

pavolloffay: The version value in snapshots would be `999-SNAPSHOT` and for released artifacts the version is derived from git tag.

How the version is reported is agnostic to this PR.

yurishkuro: The version reported to Jaeger backend as tag is derived from the artifact version, is it not?

pavolloffay: Yes, I do not understand what are you trying to point out. 

The released artifacts will report the correct version as they do now. The snapshots will report only `999-SNAPSHOT` is that a problem?

yurishkuro: internally we monitor which versions of Jaeger clients are being used in production by what `jaeger.version` tab they report to Jaeger backend, and have thresholds that warn teams if they are below minimal recommended version. Sometimes people do end up using snapshots, which is still ok if we can attribute it to a concrete min major.minor version. But with this change it looks like they all  be reporting the same literal string `999-SNAPSHOT` instead of `#.#.#-SNAPSHOT`.

So we don't care how the artifacts themselves are called, we care that accurate major.minor.patch version is reported. 

But even with snapshots, don't we want to distinguish by version? Presumably people use snapshots when there are some new features that are not released yet, it's better to have some reasonably stable version number in this case.

pavolloffay: If there are folks using snapshots and relying on a version we should keep the current behavior.

My main motivation here is to remove post-release steps with updating the changelog and gradle script. We could require bumping version in gradle in pre-release steps when updating changelog. 

E.g. if we want to release `0.36.0` somebody would bump int to `0.36.0-SNAPSHOT`.

yurishkuro: TBH the post release steps don't bother me as it's a normal practice in other languages (eg go, python, and node Jaeger clients are released the same way). But yes, ideally the bump could be done automatically. Do we already perform commit from CI when doing a release and converting the tag to version?

pavolloffay: No we do not perform commit from this repository. From what I have seen we usually forget to remove the snapshot version after the release. The bad thing is also that the build is failing then bc it expects encrypted variables.


Are you fine with removing post version bump and changing it only before the release? It's almost the same thing except that tagged version in git repository will have a snapshot suffix.

yurishkuro: But then we still need to change the version manually before the release?

pavolloffay: Yes like it is done right now. It's not ideal just an incremental improvement.

yurishkuro: Improvement in some areas, but creates inconsistency in others, eg instead of M.m.p+1-SNAPSHOT we'll have released version+SNAPSHOT.

Are we against doing auto-commits from CI on successful release? That would improve the process even more.

pavolloffay: I don't have anything against committing from travis, but I don't want to invest time to do it given OT and OC merger. To do it properly we will need 2 commits and additional tag. One tag triggers build which creates commits and tags released version.

I will close this PR if there is no interest on removing post-relese steps.

pavolloffay: > Improvement in some areas, but creates inconsistency in others, eg instead of M.m.p+1-SNAPSHOT we'll have released version+SNAPSHOT.

pavolloffay: > Improvement in some areas, but creates inconsistency in others, eg instead of M.m.p+1-SNAPSHOT we'll have released version+SNAPSHOT.

To be precise during the pre-release steps a committer can set the `M.m.p+1` in the gradle.

```
* bump the version in `build.gradle` to `major.minor.(patch+1)-SNAPSHOT` eg. `0.20.1-SNAPSHOT` - notice the `SNAPSHOT`!
```

RELEASE.md
```diff
@@ -4,19 +4,16 @@
 
 The release process consists of these steps:
   1. Create a pull request with:
-     * Change the version in [`build.gradle`](build.gradle) to match the release version e.g. `0.20.0`
      * Add an entry to [`CHANGELOG`](CHANGELOG.md) with changes since the last release
-  1. Commit your `CHANGELOG` changes
+     * add a new section to the CHANGELOG with that version as `(unreleased)` and a bullet point `- Nothing yet`
+     * bump the version in `build.gradle` to `major.minor.(patch+1)-SNAPSHOT` eg. `0.20.1-SNAPSHOT` - notice the `SNAPSHOT`!
```
pavolloffay: @yurishkuro would this satisfy your snapshot requirements? 

This way we have a snapshot version like before we just set it in advance. The only difference is if somebody checkouts a tag the and does `gradlew build` it builds snapshot instead of final.

pavolloffay:
Done in #627

PR #634, opened 2019-05-30
Add public method isFinished() on JaegerSpan

dougEfresh:
In complicated future/multi-threading environments, such as RX java, you need to pass trace context to the executing threads.  In some cases we noticed that spans where been activated with the ScopeManager long after the span has finished (like 30 minutes). 
I've noticed hourly repeating tasks used stale trace context. By adding isFinished on the JaegerSpan, we can keep track (via metrics) of spans that are being activated but are themselves finished. 

Adding public method isFinished to JaegerSpan.

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpan.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpan.java
@@ -75,6 +75,10 @@ public class JaegerSpan implements Span {
     return startTimeMicroseconds;
   }
 
+  public boolean isFinished() {
+    return finished;
+  }
+
   public long getDuration() {
     synchronized (this) {
       return durationMicroseconds;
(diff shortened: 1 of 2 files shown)


jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpan.java
```diff
@@ -75,6 +75,10 @@ public long getStart() {
     return startTimeMicroseconds;
   }
 
+  public boolean isFinished() {
+    return finished;
```
yurishkuro: needs `synchronized (this)`, see finish() method

yurishkuro:
@dougEfresh are you planning to finish this PR?

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpan.java
dougEfresh: updated

PR #649, opened 2019-08-22
Support for basic propagation of trace information using W3 TraceContext

### LINKED ISSUE #648: Support W3C Trace Context header
see https://github.com/jaegertracing/jaeger/issues/855

backjo:
Closes #648 

- Basic codec for `traceparent` header. Only supports trace/span propagation (no baggage support at the moment).

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TraceContextCodec.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TraceContextCodec.java
@@ -42,7 +42,7 @@ import java.util.Map;
  * <p>
  */
 public class TraceContextCodec implements Codec<TextMap> {
-  protected static final String TRACE_CONTEXT_NAME = "tracecontext";
+  protected static final String TRACE_CONTEXT_NAME = "traceparent";
 
   private static final int VERSION_SIZE = 2;
   private static final int TRACEPARENT_DELIMITER_SIZE = 1;
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TraceContextCodec.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TraceContextCodec.java
@@ -55,19 +55,11 @@ public class TraceContextCodec implements Codec<TextMap> {
       SPAN_ID_OFFSET + SPAN_ID_HEX_SIZE + TRACEPARENT_DELIMITER_SIZE;
   private static final int SAMPLED_FLAG = 1;
 
-  private static final int JAEGER_SAMPLED_FLAG = 1;
+  private static final byte JAEGER_SAMPLED_FLAG = 1;
 
 
   private final JaegerObjectFactory objectFactory;
   private final Long version;
-
-  /**
-   * @deprecated use {@link Builder} instead
-   */
-  @Deprecated
-  public TraceContextCodec() {
-    this(new Builder());
-  }
 
   private TraceContextCodec(Builder builder) {
     this.objectFactory = builder.objectFactory;
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TraceContextCodec.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TraceContextCodec.java
@@ -68,21 +68,9 @@ public class TraceContextCodec implements Codec<TextMap> {
 
   @Override
   public void inject(JaegerSpanContext spanContext, TextMap carrier) {
-    long traceIdHigh = spanContext.getTraceIdHigh();
-
-    //From the specification:
-    //When a system operates with a trace-id that is shorter than 16 bytes,
-    // it SHOULD fill-in the extra bytes with random values rather than zeroes.
-    // Let's say the system works with an 8-byte trace-id like 3ce929d0e0e4736.
-    // Instead of setting trace-id value to 0000000000000003ce929d0e0e4736,
-    // it SHOULD generate a value like 4bf92f3577b34da6a3ce929d0e0e4736.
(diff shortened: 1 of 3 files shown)


backjo:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-client-java/649)

jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TraceContextCodec.java
```diff
@@ -0,0 +1,157 @@
...
+    this.objectFactory = builder.objectFactory;
+    this.version = builder.version;
+  }
+
+  @Override
+  public void inject(JaegerSpanContext spanContext, TextMap carrier) {
+    long traceIdHigh = spanContext.getTraceIdHigh();
+
+    //From the specification:
+    //When a system operates with a trace-id that is shorter than 16 bytes,
```
adriancole: this is a terrible idea. sometimes IDs are truncated due to old instrumentation and this will ensure you can't match them later. is this really in the spec?

adriancole: I mean imagine if upstream is 128-bit a middle hop accidentally truncates making several outbound client calls. now you have effectively a new trace for all those client calls as there's no way to detect the truncation anymore, unless you process the entire tree and make some assumptions. All the last years heuristics to match no-longer work.

jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TraceContextCodec.java
```diff
@@ -0,0 +1,157 @@
...
+    // it SHOULD fill-in the extra bytes with random values rather than zeroes.
+    // Let's say the system works with an 8-byte trace-id like 3ce929d0e0e4736.
+    // Instead of setting trace-id value to 0000000000000003ce929d0e0e4736,
```
bogdandrutu: I think you should stick with 0 and this should be documented to make sure if you receive in the same backend spans with 8bytes you can reconstruct the same trace id.

Imagine in the backend you receive spans with 8 and 16 bytes then you need a way to transform both in 16 bytes and therefore you need a deterministic algorithm to do that.

jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TraceContextCodec.java
adriancole: also if you choose against changing the trace ID you were given, you have the second benefit of a consistent subtree. as this ends up in logs it is helpful. If it turns out the intent was not a 64-bit, rather a truncation, you would see that also in the logs.. and see the downstream affected. if everyone changes the trace id upon seeing a short one, not only can you not easily tell, but also nothing matches. I hope this makes sense.

jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TraceContextCodec.java
bogdandrutu: Please ignore this comment because I think I missunderstood the problem.

jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TraceContextCodec.java
```diff
@@ -0,0 +1,157 @@
...
+    // it SHOULD generate a value like 4bf92f3577b34da6a3ce929d0e0e4736.
+    // where 4bf92f3577b34da6a is a random value or a function of time and host value.
+    if (traceIdHigh == 0L) {
```
bogdandrutu: I think now I understand. The spec does not say that you should guarantee that the traceIdHigh != 0. It is valid to have a traceIdHigh as long it was randomly generated and not set to 0.

jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TraceContextCodec.java
backjo: I don't disagree that adding this additional information in makes it harder to use in older systems, but the spec does seem pretty rigid about this. 

> Because tracing systems may make sampling decisions based on the value of trace-id, for increased interoperability vendors MUST keep the random part of trace-id ID on the left side.
> 
> When a system operates with a trace-id that is shorter than 16 bytes, it SHOULD fill-in the extra bytes with random values rather than zeroes. Let's say the system works with an 8-byte trace-id like 3ce929d0e0e4736. Instead of setting trace-id value to 0000000000000003ce929d0e0e4736 it SHOULD generate a value like 4bf92f3577b34da6a3ce929d0e0e4736 where 4bf92f3577b34da6a is a random value or a function of time and host value.

bogdandrutu: The spec does not say that you need to correct this. The spec says that if you generate it you should follow that rule. As I mentioned in a different comment the left part can be 0 if it is randomly generated

yurishkuro: agree with @bogdandrutu's interpretation. 

Having said that, that's what happens when specs are written without reference implementation. What if we all join forces and create a single reusable library for Zipkin, Jaeger, OpenCensus, and OpenTelemetry, for parsing W3C trace context? All 4 projects have identical format of trace IDs. The `tracestate` might need some customization.

backjo: Going to remove this - thanks for the help clarifying :) 

Curious how we should handle the case where users have configured the tracer to not be 128 bit but register the trace context codec

pavolloffay:
Done in #694

PR #654, opened 2019-09-30
Run crossdock with all-in-one

bhavin192:
Reference: jaegertracing/jaeger#1762

### DIFF
--- a/jaeger-crossdock/docker-compose.yml
+++ b/jaeger-crossdock/docker-compose.yml
@@ -35,23 +35,32 @@ services:
 
     go:
         image: jaegertracing/xdock-go
+        links:
+            - "jaeger:jaeger-agent"
         ports:
             - "8080-8082"
 
     python:
         image: jaegertracing/xdock-py
+        links:
+            - "jaeger:jaeger-agent"
         ports:
             - "8080-8082"
 
     nodejs:
         image: jaegertracing/xdock-node
+        links:
+            - "jaeger:jaeger-agent"
         ports:
             - "8080-8082"
 
     java-udp:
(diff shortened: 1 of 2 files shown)


bhavin192:
One test is still failing, I will try to investigate this later today.

jaeger-crossdock/docker-compose.yml
```diff
@@ -35,40 +35,73 @@ services:
...
     nodejs:
         image: jaegertracing/xdock-node
+        links:
+            - "jaeger:jaeger-agent"
         ports:
             - "8080-8082"
 
     java-udp:
         depends_on:
             - test_driver
```
yurishkuro: I think this needs to depend on `jaeger`, if you look at PRs for the other clients

bhavin192: Yes, you are right. I think earlier it was like `jave-udp -> test_driver -> (jaeger components)`.

jaeger-crossdock/docker-compose.yml
```diff
@@ -35,40 +35,74 @@ services:
...
+            - jaeger
         ports:
             - "8080-8082"
         environment:
             - SENDER=udp
 
     java-http:
         build: .
+        links:
+            - "jaeger:jaeger-agent"
```
yurishkuro: Did you get the test pass locally? Failure looks specific to the change.

Java http actor would be trying to submit spans to collector, so the renaming probably needs to be to jaeger-collector.

bhavin192: Yep, you are right. With suggested change I got it working locally

PR #655, opened 2019-10-02
Java: Cleanup operation samplers if absent from strategy response

guo0693:
Signed-off-by: jung <jung@uber.com>

- Fixes [#1728](https://github.com/jaegertracing/jaeger/issues/1728)

- If the server did not return a strategy for operation X, the operation sampler will be removed and the default sampler will be used for operation X

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/samplers/PerOperationSampler.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/samplers/PerOperationSampler.java
@@ -18,12 +18,14 @@ import io.jaegertracing.internal.samplers.http.OperationSamplingParameters;
 import io.jaegertracing.internal.samplers.http.PerOperationSamplingParameters;
 import io.jaegertracing.spi.Sampler;
 import java.util.HashMap;
+import java.util.Map;
 import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.ToString;
 import lombok.extern.slf4j.Slf4j;
+
 
 /**
  * Computes {@link #sample(String, long)} using the name of the operation, and maintains a specific
@@ -36,7 +38,7 @@ import lombok.extern.slf4j.Slf4j;
 @Getter(AccessLevel.PACKAGE) //Visible for testing
 public class PerOperationSampler implements Sampler {
   private final int maxOperations;
-  private final HashMap<String, GuaranteedThroughputSampler> operationNameToSampler;
+  private Map<String, GuaranteedThroughputSampler> operationNameToSampler;
   private ProbabilisticSampler defaultSampler;
   private double lowerBound;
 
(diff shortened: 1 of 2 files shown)


jaeger-core/src/main/java/io/jaegertracing/internal/samplers/PerOperationSampler.java
```diff
@@ -17,14 +17,13 @@
 import io.jaegertracing.internal.samplers.http.OperationSamplingParameters;
 import io.jaegertracing.internal.samplers.http.PerOperationSamplingParameters;
 import io.jaegertracing.spi.Sampler;
-import java.util.HashMap;
-import lombok.AccessLevel;
-import lombok.AllArgsConstructor;
-import lombok.EqualsAndHashCode;
-import lombok.Getter;
-import lombok.ToString;
+import lombok.*;
```
vprithvi: Seems like an unnecessary change

guo0693: IDE automatically optimized the import.

jaeger-core/src/main/java/io/jaegertracing/internal/samplers/PerOperationSampler.java
```diff
@@ -64,6 +63,20 @@ public synchronized boolean update(OperationSamplingParameters strategies) {
...
+    for (PerOperationSamplingParameters strategy : strategies.getPerOperationStrategies()) {
+      newOps.add(strategy.getOperation());
+    }
+
+    for (String op : operationNameToSampler.keySet()) {
+      if (!newOps.contains(op)) {
+        operationNameToSampler.remove(op);
+        isUpdated = true;
+      }
+    }
```
vprithvi: Could it be simpler to create a new `HashMap<String, GuaranteedThroughputSampler>`, and either insert a sampler update, or create a new sampler; and replace `operationNameToSampler` with this new Map?

jaeger-core/src/main/java/io/jaegertracing/internal/samplers/PerOperationSampler.java
vprithvi: It's still an unnecessary change

jaeger-core/src/main/java/io/jaegertracing/internal/samplers/PerOperationSampler.java
guo0693: You are right, I thought about recreate the remote sampler using the response and then thought we need to preserve the sampler's internal state, haven't thought about re-insert the sampler to new map. Updating.

jaeger-core/src/test/java/io/jaegertracing/internal/samplers/PerOperationSamplerTest.java
```diff
@@ -169,10 +169,29 @@ public void testUpdateAddOperation() {
...
-                 operationToSamplers.get(OPERATION));
+            undertest.getOperationNameToSampler().get(OPERATION));
   }
 
+  @Test
+  public void testAbsentOperationIsRemoved() {
+    String absentOp = "ShouldBeRemoved";
+    operationToSamplers.put(absentOp, mock(GuaranteedThroughputSampler.class));
+    PerOperationSamplingParameters perOperationSamplingParameters1 =
+            new PerOperationSamplingParameters(OPERATION, new ProbabilisticSamplingStrategy(SAMPLING_RATE));
```
vprithvi: Shouldn't `operationToSamplers` be passed in here to control the starting state of the test?

guo0693: undertest is inited with the operationToSamplers during setup, adding a sampler to the map should do the job regarding the starting state.

PR #662, opened 2019-10-17
Add logging for SenderException in RemoteReporter

### LINKED ISSUE #653: Log SenderException errors in RemoteReporter
Troubleshooting errors in the trace pipeline can be difficult. The metrics emitted by Jaeger tell me there is an error, but I'm unable to see any information in the logs about that error. 

SenderException is swallowed in RemoteReporter

Log the exception at error level

esukram:
Closes #653

In order to get more context if a `SenderException` has been thrown, the complete exception is logged with `error` level.


jpkrohling:
Thanks for your contribution!

PR #664, opened 2019-10-18
Production and test code cleanup

esukram:
None

During the work of #653 I saw some code that was generating warnings in the IDE. This PR does solve these.

Please let me know, what you think.


esukram:
One thing that made me wonder in particular:
`JaegerSpanTest.java` has three test methods that have *testConcurrentModification* in its name. But as far I can tell, the concurrence is not tested at all.
I did not change that in this PR to not modify the test behaviour, but I would be interested in your opinion.

yurishkuro:
`testConcurrentModification` happens because the tests are changing the same collection (conceptully) that they are iterating over.

PR #670, opened 2019-11-30
Fix ThriftSender max span size check

### LINKED ISSUE #660: client stops working after failing to send messages that are too large

Sending Jaeger spans.

After a few unsuccessful sends the client stops sending *any* spans. I've got these error messages by logging the exception as explained in #653 .

There's a couple of these

```
io.jaegertracing.internal.exceptions.SenderException: ThriftSender received a span that was too large, size = 109837, max = 64967
        at io.jaegertracing.thrift.internal.senders.ThriftSender.append(ThriftSender.java:58) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        at com.gratex.bl.upm2.config.JaegerRemoteReporter$AppendCommand.execute(JaegerRemoteReporter.java:139) ~[classes/:1.0.1]
        at com.gratex.bl.upm2.config.JaegerRemoteReporter$QueueProcessor.run(JaegerRemoteReporter.java:174) [classes/:1.0.1]
        at java.lang.Thread.run(Thread.java:748) [?:1.8.0_191]
```

*NOTE*: JaegerRemoteReporter is RemoteReporter + the 1 line for the log.error

After a while I get errors like this:

```
io.jaegertracing.internal.exceptions.SenderException: Failed to flush spans.
        at io.jaegertracing.thrift.internal.senders.ThriftSender.append(ThriftSender.java:76) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        at com.gratex.bl.upm2.config.JaegerRemoteReporter$AppendCommand.execute(JaegerRemoteReporter.java:139) ~[classes/:1.0.1]
        at com.gratex.bl.upm2.config.JaegerRemoteReporter$QueueProcessor.run(JaegerRemoteReporter.java:174) [classes/:1.0.1]
        at java.lang.Thread.run(Thread.java:748) [?:1.8.0_191]
Caused by: io.jaegertracing.internal.exceptions.SenderException: Could not send 1 spans
        at io.jaegertracing.thrift.internal.senders.UdpSender.send(UdpSender.java:66) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        at io.jaegertracing.thrift.internal.senders.ThriftSender.flush(ThriftSender.java:110) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        at io.jaegertracing.thrift.internal.senders.ThriftSender.append(ThriftSender.java:73) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        ... 3 more
Caused by: org.apache.thrift.transport.TTransportException: Message size too large: 1 > 65000
        at io.jaegertracing.thrift.internal.reporters.protocols.ThriftUdpTransport.write(ThriftUdpTransport.java:132) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        at org.apache.thrift.protocol.TCompactProtocol.writeByteDirect(TCompactProtocol.java:486) ~[libthrift-0.12.0.jar:0.12.0]
        at org.apache.thrift.protocol.TCompactProtocol.writeByteDirect(TCompactProtocol.java:493) ~[libthrift-0.12.0.jar:0.12.0]
        at org.apache.thrift.protocol.TCompactProtocol.writeFieldBeginInternal(TCompactProtocol.java:260) ~[libthrift-0.12.0.jar:0.12.0]
        at org.apache.thrift.protocol.TCompactProtocol.writeFieldBegin(TCompactProtocol.java:242) ~[libthrift-0.12.0.jar:0.12.0]
        at io.jaegertracing.thriftjava.Tag$TagStandardScheme.write(Tag.java:920) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        at io.jaegertracing.thriftjava.Tag$TagStandardScheme.write(Tag.java:809) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        at io.jaegertracing.thriftjava.Tag.write(Tag.java:709) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        at io.jaegertracing.thriftjava.Span$SpanStandardScheme.write(Span.java:1360) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        at io.jaegertracing.thriftjava.Span$SpanStandardScheme.write(Span.java:1144) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        at io.jaegertracing.thriftjava.Span.write(Span.java:1028) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        at io.jaegertracing.thriftjava.Batch$BatchStandardScheme.write(Batch.java:473) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        at io.jaegertracing.thriftjava.Batch$BatchStandardScheme.write(Batch.java:407) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        at io.jaegertracing.thriftjava.Batch.write(Batch.java:344) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        at io.jaegertracing.agent.thrift.Agent$emitBatch_args$emitBatch_argsStandardScheme.write(Agent.java:1056) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        at io.jaegertracing.agent.thrift.Agent$emitBatch_args$emitBatch_argsStandardScheme.write(Agent.java:1018) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        at io.jaegertracing.agent.thrift.Agent$emitBatch_args.write(Agent.java:969) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        at org.apache.thrift.TServiceClient.sendBase(TServiceClient.java:71) ~[libthrift-0.12.0.jar:0.12.0]
        at org.apache.thrift.TServiceClient.sendBaseOneway(TServiceClient.java:66) ~[libthrift-0.12.0.jar:0.12.0]
        at io.jaegertracing.agent.thrift.Agent$Client.send_emitBatch(Agent.java:70) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        at io.jaegertracing.agent.thrift.Agent$Client.emitBatch(Agent.java:63) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        at io.jaegertracing.thrift.internal.senders.UdpSender.send(UdpSender.java:64) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        at io.jaegertracing.thrift.internal.senders.ThriftSender.flush(ThriftSender.java:110) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        at io.jaegertracing.thrift.internal.senders.ThriftSender.append(ThriftSender.java:73) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        ... 3 more
```

```
io.jaegertracing.internal.exceptions.SenderException: Failed to flush spans.
        at io.jaegertracing.thrift.internal.senders.ThriftSender.append(ThriftSender.java:76) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        at com.gratex.bl.upm2.config.JaegerRemoteReporter$AppendCommand.execute(JaegerRemoteReporter.java:139) ~[classes/:1.0.1]
        at com.gratex.bl.upm2.config.JaegerRemoteReporter$QueueProcessor.run(JaegerRemoteReporter.java:174) [classes/:1.0.1]
        at java.lang.Thread.run(Thread.java:748) [?:1.8.0_191]
Caused by: io.jaegertracing.internal.exceptions.SenderException: Could not send 8 spans
        at io.jaegertracing.thrift.internal.senders.UdpSender.send(UdpSender.java:66) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        at io.jaegertracing.thrift.internal.senders.ThriftSender.flush(ThriftSender.java:110) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        at io.jaegertracing.thrift.internal.senders.ThriftSender.append(ThriftSender.java:73) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        ... 3 more
Caused by: org.apache.thrift.transport.TTransportException: Message size too large: 1 > 65000
        at io.jaegertracing.thrift.internal.reporters.protocols.ThriftUdpTransport.write(ThriftUdpTransport.java:132) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        at org.apache.thrift.protocol.TCompactProtocol.writeByteDirect(TCompactProtocol.java:486) ~[libthrift-0.12.0.jar:0.12.0]
        at org.apache.thrift.protocol.TCompactProtocol.writeMessageBegin(TCompactProtocol.java:206) ~[libthrift-0.12.0.jar:0.12.0]
        at org.apache.thrift.TServiceClient.sendBase(TServiceClient.java:70) ~[libthrift-0.12.0.jar:0.12.0]
        at org.apache.thrift.TServiceClient.sendBaseOneway(TServiceClient.java:66) ~[libthrift-0.12.0.jar:0.12.0]
        at io.jaegertracing.agent.thrift.Agent$Client.send_emitBatch(Agent.java:70) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        at io.jaegertracing.agent.thrift.Agent$Client.emitBatch(Agent.java:63) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        at io.jaegertracing.thrift.internal.senders.UdpSender.send(UdpSender.java:64) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        at io.jaegertracing.thrift.internal.senders.ThriftSender.flush(ThriftSender.java:110) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        at io.jaegertracing.thrift.internal.senders.ThriftSender.append(ThriftSender.java:73) ~[jaeger-thrift-1.0.0.jar:1.0.0]
        ... 3 more
```

A few failing messages shouldn't cause the whole client to fall down. I suspect some internal byte buffer doesn't get reinitialized correctly.

Also, I'd like to have the message size configurable via an environment variable.

None. Thank you!

yhpark:
Signed-off-by: Yeonghoon Park <me@yhpark.io>

Fixes #660. Currently the span size check only calculates the byte size of span, where it should also take the process size into account.

Consider a span with size less than `getMaxSpanBytes()` and greater than `getMaxSpanBytes() - processSize`. We cannot create a batch that fits UDP limit because we always have to send the `Process`. Still it passes the size check and corrupts write buffer in ThriftUdpTransport when flushed (which is another problem not dealt in this PR). Then the write buffer is always full with 65000 bytes in the buffer, never reinitialized, preventing all further spans from being sent.

I think this is a very critical issue because it causes the client to stop sending any spans altogether, and we actually encountered this problem frequently in production. For now, a workaround is to set the `ThriftSenderBase.maxPacketSize` value to something like 64000 by injecting a custom `ThriftSenderFactory` via ServiceLoader (see `SenderResolver.resolve()`). `Process` without custom tags takes around 100 bytes, so using value less than 64800 would be okay.

- Before: `span size + Batch thrift overhead < 65000`
  (`spanSize > getMaxSpanBytes()`)
- After: `process size + span size + Batch thrift overhead < 65000`
  (`(processSize + spanSize) > getMaxSpanBytes()`)

### DIFF
--- a/jaeger-thrift/src/main/java/io/jaegertracing/thrift/internal/senders/ThriftSender.java
+++ b/jaeger-thrift/src/main/java/io/jaegertracing/thrift/internal/senders/ThriftSender.java
@@ -54,15 +54,15 @@ public abstract class ThriftSender extends ThriftSenderBase implements Sender {
 
     io.jaegertracing.thriftjava.Span thriftSpan = JaegerThriftSpanConverter.convertSpan(span);
     int spanSize = calculateSpanSize(thriftSpan);
-    if ((processBytesSize + spanSize) > getMaxSpanBytes()) {
+    if (spanSize > getMaxSpanBytes()) {
       throw new SenderException(String.format("ThriftSender received a span that was too large, size = %d, max = %d",
-          (processBytesSize + spanSize), getMaxSpanBytes()), null, 1);
+          spanSize, getMaxSpanBytes()), null, 1);
     }
 
     byteBufferSize += spanSize;
-    if (byteBufferSize <= getMaxSpanBytes()) {
+    if (byteBufferSize <= getMaxBatchBytes()) {
       spanBuffer.add(thriftSpan);
-      if (byteBufferSize < getMaxSpanBytes()) {
+      if (byteBufferSize < getMaxBatchBytes()) {
         return 0;
       }
       return flush();
@@ -95,6 +95,10 @@ public abstract class ThriftSender extends ThriftSenderBase implements Sender {
     } catch (Exception e) {
       throw new SenderException("ThriftSender failed writing Span to memory buffer.", e, 1);
     }
+  }
+
+  protected int getMaxSpanBytes() {
+    return getMaxBatchBytes() - processBytesSize;
   }
 
   public abstract void send(Process process, List<io.jaegertracing.thriftjava.Span> spans) throws SenderException;
--- a/jaeger-thrift/src/main/java/io/jaegertracing/thrift/internal/senders/ThriftSender.java
+++ b/jaeger-thrift/src/main/java/io/jaegertracing/thrift/internal/senders/ThriftSender.java
@@ -115,7 +115,7 @@ public abstract class ThriftSender extends ThriftSenderBase implements Sender {
       throw new SenderException("Failed to flush spans.", e, n);
     } finally {
       spanBuffer.clear();
-      spanBytesSize = processBytesSize;
+      spanBytesSize = 0;
     }
     return n;
   }
--- a/jaeger-thrift/src/main/java/io/jaegertracing/thrift/internal/senders/ThriftSender.java
+++ b/jaeger-thrift/src/main/java/io/jaegertracing/thrift/internal/senders/ThriftSender.java
@@ -29,7 +29,7 @@ public abstract class ThriftSender extends ThriftSenderBase implements Sender {
 
   private Process process;
   private int processBytesSize;
-  private int byteBufferSize;
+  private int spanBytesSize;
 
   @ToString.Exclude private List<io.jaegertracing.thriftjava.Span> spanBuffer;
 
@@ -49,7 +49,6 @@ public abstract class ThriftSender extends ThriftSenderBase implements Sender {
       process = new Process(span.getTracer().getServiceName());
       process.setTags(JaegerThriftSpanConverter.buildTags(span.getTracer().tags()));
       processBytesSize = calculateProcessSize(process);
-      byteBufferSize += processBytesSize;
     }
 
     io.jaegertracing.thriftjava.Span thriftSpan = JaegerThriftSpanConverter.convertSpan(span);
@@ -59,10 +58,10 @@ public abstract class ThriftSender extends ThriftSenderBase implements Sender {
           spanSize, getMaxSpanBytes()), null, 1);
     }
 
-    byteBufferSize += spanSize;
-    if (byteBufferSize <= getMaxBatchBytes()) {
+    spanBytesSize += spanSize;
+    if (spanBytesSize <= getMaxSpanBytes()) {
       spanBuffer.add(thriftSpan);
-      if (byteBufferSize < getMaxBatchBytes()) {
+      if (spanBytesSize < getMaxSpanBytes()) {
         return 0;
       }
       return flush();
@@ -77,7 +76,7 @@ public abstract class ThriftSender extends ThriftSenderBase implements Sender {
     }
 
     spanBuffer.add(thriftSpan);
-    byteBufferSize = processBytesSize + spanSize;
+    spanBytesSize = spanSize;
     return n;
   }
 
@@ -116,7 +115,7 @@ public abstract class ThriftSender extends ThriftSenderBase implements Sender {
       throw new SenderException("Failed to flush spans.", e, n);
     } finally {
       spanBuffer.clear();
-      byteBufferSize = processBytesSize;
+      spanBytesSize = processBytesSize;
     }
     return n;
   }
--- a/jaeger-thrift/src/main/java/io/jaegertracing/thrift/internal/senders/ThriftSender.java
+++ b/jaeger-thrift/src/main/java/io/jaegertracing/thrift/internal/senders/ThriftSender.java
@@ -54,9 +54,9 @@ public abstract class ThriftSender extends ThriftSenderBase implements Sender {
 
     io.jaegertracing.thriftjava.Span thriftSpan = JaegerThriftSpanConverter.convertSpan(span);
     int spanSize = calculateSpanSize(thriftSpan);
-    if (spanSize > getMaxSpanBytes()) {
+    if ((processBytesSize + spanSize) > getMaxSpanBytes()) {
       throw new SenderException(String.format("ThriftSender received a span that was too large, size = %d, max = %d",
-          spanSize, getMaxSpanBytes()), null, 1);
+          (processBytesSize + spanSize), getMaxSpanBytes()), null, 1);
     }
 
     byteBufferSize += spanSize;
--- a/jaeger-thrift/src/test/java/io/jaegertracing/thrift/internal/senders/UdpSenderTest.java
+++ b/jaeger-thrift/src/test/java/io/jaegertracing/thrift/internal/senders/UdpSenderTest.java
@@ -84,21 +84,30 @@ public class UdpSenderTest {
     reporter.close();
   }
 
-  @Test(expected = SenderException.class)
-  public void testAppendSpanTooLarge() throws Exception {
+  /**
+   * create a mock span with target size
+   */
+  private JaegerSpan buildSpanWithSize(int targetSpanSize) throws Exception {
     JaegerSpan jaegerSpan = tracer.buildSpan("raza").start();
-    String msg = "";
-    for (int i = 0; i < 10001; i++) {
-      msg += ".";
-      jaegerSpan.log(msg);
+    jaegerSpan.setTag("mock", "");
+
+    // contains mock span tag with empty string value, which is later used to match target span size
+    io.jaegertracing.thriftjava.Span emptySpan = JaegerThriftSpanConverter.convertSpan(jaegerSpan);
+
+    int emptySpanSize = sender.getSize(emptySpan);
+    int freePacketSizeLeft = targetSpanSize - emptySpanSize;
+
+    // each "0" takes 1 byte in UTF-8
+    String mockStr = String.join("", Collections.nCopies(freePacketSizeLeft, "0"));
+    jaegerSpan.setTag("mock", mockStr);
+
+    // subtract some chars since the actual encoding takes a few extra bytes
+    while (targetSpanSize != sender.getSize(JaegerThriftSpanConverter.convertSpan(jaegerSpan))) {
+      mockStr = mockStr.substring(1);
+      jaegerSpan.setTag("mock", mockStr);
     }
 
-    try {
-      sender.append(jaegerSpan);
-    } catch (SenderException e) {
-      assertEquals(e.getDroppedSpanCount(), 1);
-      throw e;
-    }
+    return jaegerSpan;
   }
 
   @Test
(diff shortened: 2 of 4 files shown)


jaeger-thrift/src/test/java/io/jaegertracing/thrift/internal/senders/UdpSenderTest.java
```diff
@@ -84,21 +84,29 @@ public void tearDown() throws Exception {
...
-    } catch (SenderException e) {
-      assertEquals(e.getDroppedSpanCount(), 1);
-      throw e;
-    }
+    int emptySpanSize = sender.getSize(emptySpan);
+    int freePacketSizeLeft = targetSpanSize - emptySpanSize;
+
+    // each "0" takes 1 byte in UTF-8
+    // size header seems to take 1 byte for every 2^7 string length
+    int lengthBytes = (31 - Integer.numberOfLeadingZeros(freePacketSizeLeft)) / 7;
```
pavolloffay: Could you please explain better in comments why 31 and 7?

jaeger-thrift/src/test/java/io/jaegertracing/thrift/internal/senders/UdpSenderTest.java
```diff
@@ -84,21 +84,29 @@ public void tearDown() throws Exception {
...
+   */
+  private JaegerSpan buildSpanWithSize(int targetSpanSize) throws Exception {
     JaegerSpan jaegerSpan = tracer.buildSpan("raza").start();
-    String msg = "";
-    for (int i = 0; i < 10001; i++) {
-      msg += ".";
-      jaegerSpan.log(msg);
-    }
+    jaegerSpan.setTag("mock", "");
+    io.jaegertracing.thriftjava.Span emptySpan = JaegerThriftSpanConverter.convertSpan(jaegerSpan);
```
pavolloffay: Add a comment that size contains mock span tag which is used later.

jaeger-thrift/src/test/java/io/jaegertracing/thrift/internal/senders/UdpSenderTest.java
yhpark: `(31 - Integer.numberOfLeadingZeros(x))` is a general formula for calculating stable log base 2 (w/o using floating point)
`7` is just a number I found but I don't understand how the protocol takes additional one byte every 2^7 character. I even read the protobuf spec, but don't know (why 7 instead of 8).
Overall I think the code is quite cryptic, do you think I should try a different approach? (maybe brute force by subtracting '0's one by one until `targetSpanSize == spanSize`)

jaeger-thrift/src/test/java/io/jaegertracing/thrift/internal/senders/UdpSenderTest.java
yhpark: done

jaeger-thrift/src/main/java/io/jaegertracing/thrift/internal/senders/ThriftSender.java
```diff
@@ -54,9 +54,9 @@ public int append(JaegerSpan span) throws SenderException {
 
     io.jaegertracing.thriftjava.Span thriftSpan = JaegerThriftSpanConverter.convertSpan(span);
     int spanSize = calculateSpanSize(thriftSpan);
-    if (spanSize > getMaxSpanBytes()) {
+    if ((processBytesSize + spanSize) > getMaxSpanBytes()) {
       throw new SenderException(String.format("ThriftSender received a span that was too large, size = %d, max = %d",
-          spanSize, getMaxSpanBytes()), null, 1);
+          (processBytesSize + spanSize), getMaxSpanBytes()), null, 1);
```
yurishkuro: I feel like this fix is not completely accurate. Look at the statements below, which again compare accumulated buffer size and getMaxSpanBytes(). I may be wrong, but it seems that the issue is with `getMaxSpanBytes()` - from its name it seems to mean "total space allocated for _span bytes only_, excluding process bytes", but the way it is calculated in the constructor is different, it does not include process bytes.

yhpark: @yurishkuro If you're referring to the if statements at line 63-69, when `process size + span size` exceeds the maximum byte size the if statements are bypassed (since `byteBufferSize` would be greater than `getMaxSpanBytes()`) then an exception is thrown at line 73 (`flush()`). When this happens `ThriftUdpTransport` is left in an unrecoverable state, as I mentioned in the PR description. You can reproduce this by running the test cases.

yhpark: Or do you mean the fix should be to modify getMaxSpanBytes() to return the space excluding the process bytes?

yurishkuro: I am not sure which one is best, it's just that the logic is getting very confusing. `getMaxSpanBytes()`, despite its name, returns the capacity of _both_ the spans and the process. Same for `byteBufferSize` variable - we have to keep remembering to add `processBytesSize`. For example, in L73 we call `flush()`, which assigns `byteBufferSize = byteBufferSize = processBytesSize;`, but then in L80 we add `processBytesSize` again.

Perhaps we should:
  * rename getMaxSpanBytes to getMaxBatchBytes in the base class and add comment that the size includes `process + spans`
  * implement getMaxSpanBytes just in this class as `getMaxBatchBytes() - processBytesSize´

yhpark: That makes total sense. Will update soon.

jaeger-thrift/src/main/java/io/jaegertracing/thrift/internal/senders/ThriftSender.java
```diff
@@ -60,9 +60,9 @@ public int append(JaegerSpan span) throws SenderException {
     }
 
     byteBufferSize += spanSize;
-    if (byteBufferSize <= getMaxSpanBytes()) {
+    if (byteBufferSize <= getMaxBatchBytes()) {
       spanBuffer.add(thriftSpan);
-      if (byteBufferSize < getMaxSpanBytes()) {
+      if (byteBufferSize < getMaxBatchBytes()) {
```
yhpark: Renamed but the comparisons are the same as before. Since `byteBufferSize` contains process bytes, here we check whether it exceeds max batch size.

jaeger-thrift/src/main/java/io/jaegertracing/thrift/internal/senders/ThriftSender.java
```diff
@@ -54,15 +54,15 @@ public int append(JaegerSpan span) throws SenderException {
 
     io.jaegertracing.thriftjava.Span thriftSpan = JaegerThriftSpanConverter.convertSpan(span);
     int spanSize = calculateSpanSize(thriftSpan);
-    if ((processBytesSize + spanSize) > getMaxSpanBytes()) {
+    if (spanSize > getMaxSpanBytes()) {
```
yhpark: span > max span

jaeger-thrift/src/main/java/io/jaegertracing/thrift/internal/senders/ThriftSender.java
yurishkuro: I think this should continue comparing with getMaxSpanBytes, but we should remove the code that adds process size to byteBufferSize (in a couple of places)

yhpark: Yeah that would simplify the code overall. Will update soon

jaeger-thrift/src/main/java/io/jaegertracing/thrift/internal/senders/ThriftSender.java
```diff
@@ -116,7 +115,7 @@ public int flush() throws SenderException {
       throw new SenderException("Failed to flush spans.", e, n);
     } finally {
       spanBuffer.clear();
-      byteBufferSize = processBytesSize;
+      spanBytesSize = processBytesSize;
```
yurishkuro: shouldn't this be `spanBytesSize = 0`?

yhpark: yeah... it seems so. I'll also add some test for this

jaeger-thrift/src/test/java/io/jaegertracing/thrift/internal/senders/UdpSenderTest.java
```diff
@@ -131,6 +140,41 @@ public void testAppend() throws Exception {
...
+
+    int result = sender.append(jaegerSpan);
+
+    assertEquals(1, result);
+
+    // test if the buffer is reinitialized correctly
+    result = sender.append(jaegerSpan);
+
+    assertEquals(1, result);
+  }
```
yurishkuro: nice!

PR #679, opened 2020-01-08
Unregister shutdown hook when tracer is closed

### LINKED ISSUE #677: Memory leak in JaegerTracer

Basic usage, just registering a JaegerTracer in an application.

Each JaegerTracer which is created registers a shutdown hook, which purpose is to close the tracer on application exit. This shutdown hook is however not unregistered when the tracer is closed programmatically. This hook holds references to the tracer, so it can't be garbage collected.

Unregister the shutdown hook when the tracer is closed.

Not aware.

TomasHofman:
Resolves #677

- Avoid memory leak in JaegerTracer.

- Unregister the shutdown hook when tracer is closed. (The hook is registered during tracer instance creation.)


pavolloffay:
@TomasHofman do you also need this fix in version 0.34.0?

TomasHofman:
> @TomasHofman do you also need this fix in version 0.34.0?

That would be ideal, that's the version we are using in Wildfly.

PR #682, opened 2020-01-23
Feature request: sampling based on tags.

hairyhum:
Sometimes it's important to be able to sample based on data rather than
opertaion names. Since tags are already available during the sampling
they can be used as an additional value to make a sampling decision.

For example in messages processing it would be helpful to trace some
topics more than other.

This is an example implementation of how could this work on the tracer
level.
The ProbabilisticTagSampler can be created to set different sampling rates for different tag values:
```
Map<Object, Double> rates = new HashMap<>();
rates.put("interesting_topic", 1.0);
rates.put("less_interesting_topic", 0.5);
Tracer tracer = Configuration.fromEnv().withSampler(new ProbabilisticTagSampler("topic", rates, 0.1).getTracer();
Span alwaysSampled = tracer.buildSpan("operation").withTag("topic", "interesting_topic").start();
Span oftenSampled = tracer.buildSpan("operation").withTag("topic", "less_interesting_topic").start();
Span rareSampled = tracer.buildSpan("operation").withTag("topic", "some_other_topic").start();

Tracer tracer = io.jaegertracing.Configuration.fromEnv().withSampler().getTracer();
final Span span = tracer.buildSpan("foo").withTag("interesting_topic").start();
```

Ideally it would be nice to have this feature on the agent/collector level as
well.

This is more of a feature request than a PR, the code is mostly to demonstrate possible API.

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
@@ -29,6 +29,7 @@ import io.jaegertracing.internal.propagation.TextMapCodec;
 import io.jaegertracing.internal.reporters.RemoteReporter;
 import io.jaegertracing.internal.samplers.RemoteControlledSampler;
 import io.jaegertracing.internal.samplers.SamplingStatus;
+import io.jaegertracing.internal.samplers.TagSamplerWrapper;
 import io.jaegertracing.internal.utils.Utils;
 import io.jaegertracing.spi.BaggageRestrictionManager;
 import io.jaegertracing.spi.Extractor;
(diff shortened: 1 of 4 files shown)


yurishkuro:
Note that tag sampling is already implemented in Node and Go clients, with certain upgrades to the overall sampling framework used by the clients.  https://github.com/jaegertracing/jaeger/issues/1861

hairyhum:
This changes propose a different approach to delayed sampling, because sampling is still done in the beginning of the span. This is just to minimize the code changes.
If this is interesting, I can try to move sampling to the later stages, but that may require more changes.

yurishkuro:
We want to keep consistency in the clients. Delayed sampling is a more general solution that covers tag sampling.

william-tran:
@yurishkuro are you looking for contributions to support delayed sampling in the java client?

yurishkuro:
@william-tran certainly!

PR #685, opened 2020-02-12
Configuration do not cache sender instance

### LINKED ISSUE #684: Configuration shouldn't keep a reference to a closed sender
When reusing a Configuration to create a tracer, if a previous tracer was closed then the sender should be created again as its connections to the agent are closed.

When redeploying an application the produced tracer's sender is failing because the connection should be reopened.

Adding a closed attribute on a sender to resolve it again when creating a new tracer.

ehsavoie:
Resolves #684 

 * Adding a closed attribute on sender
 * Sender configuration is resolving again is sender is closed.

Issue: #684

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/Configuration.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/Configuration.java
@@ -616,11 +616,6 @@ public class Configuration {
   @Getter
   public static class SenderConfiguration {
     /**
-     * A custom sender set by our consumers. If set, nothing else has effect. Optional.
-     */
-    private Sender sender;
-
-    /**
      * The Agent Host. Has no effect if the sender is set. Optional.
      */
     private String agentHost;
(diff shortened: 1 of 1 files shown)


pavolloffay:
There are couple of inconsistencies in the `Configuration` class. For instance some objects are cached (tracer, sender) and some not.

Also weird is that `Configuration.getTracerBuilder()` returns a new builder instance even though a tracer has been initialized and it is cached in the class.

Instead of adding a `close` to `Sender` I am wondering whether a better solution would be to do not cache configuration objects in  `getTracerBuilder` and always resolve new instances.

ehsavoie:
@pavolloffay I agree, a simpler solution is to remove those cached instances. Just did that  :)

jaeger-core/src/main/java/io/jaegertracing/Configuration.java
```diff
@@ -181,10 +181,6 @@
   private Map<String, String> tracerTags;
   private boolean useTraceId128Bit;
 
-  /**
-   * lazy singleton JaegerTracer initialized in getTracer() method.
-   */
-  private JaegerTracer tracer;
```
pavolloffay: This will have to stay here as people are caching `Configuration.getTracer()`.

 We just don't want to cache other parts related to tracer builder.

ehsavoie: agreed

PR #686, opened 2020-02-12
[Backport to 0.34] Configuration do not cache sender instance

### LINKED ISSUE #684: Configuration shouldn't keep a reference to a closed sender

ehsavoie:
Backport #685

* Adding a closed attribute on sender
* Sender configuration is resolving again is sender is closed.

Issue: #684
Signed-off-by: Emmanuel Hugonnet <ehugonne@redhat.com>

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/Configuration.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/Configuration.java
@@ -580,11 +580,6 @@ public class Configuration {
   @Getter
   public static class SenderConfiguration {
     /**
-     * A custom sender set by our consumers. If set, nothing else has effect. Optional.
-     */
-    private Sender sender;
-
-    /**
      * The Agent Host. Has no effect if the sender is set. Optional.
      */
     private String agentHost;
@@ -648,15 +643,11 @@ public class Configuration {
     }
 
     /**
-     * Returns a sender if one was given when creating the configuration, or attempts to create a sender based on the
-     * configuration's state.
+     * Returns a sender based on the configuration's state.
      * @return the sender passed via the constructor or a properly configured sender
      */
     public Sender getSender() {
-      if (sender == null) {
-        sender = SenderResolver.resolve(this);
-      }
-      return sender;
+      return SenderResolver.resolve(this);
     }
 
     /**
--- a/jaeger-core/src/test/java/io/jaegertracing/ConfigurationTest.java
+++ b/jaeger-core/src/test/java/io/jaegertracing/ConfigurationTest.java
@@ -313,8 +313,11 @@ public class ConfigurationTest {
   @Test
   public void testDefaultTracer() {
     Configuration configuration = new Configuration("name");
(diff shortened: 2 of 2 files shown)


pavolloffay:
@yurishkuro see the comment in https://github.com/jaegertracing/jaeger-client-java/pull/685#issuecomment-585194473

My idea is to cache the tracer but the `Configuration.getTracerBuilder()` would return a new instance of the builder with all parts. I cannot think of a reason why we should cache inner configuration objects.

ehsavoie:
@yurishkuro so changing the configuration will return the cached tracer and not a tracer in sync with the configuration you are calling the getTracer from ? I find this counter intuitive. Also that means that you can get a closed tracer or as i found out a tracer with an underlying closed sender.

yurishkuro:
@pavolloffay the `getTracer()` method no longer returns cached tracer, that is a breaking change. To my knowledge there was no other "caching" behavior, e.g. a sender is not cached, but can be provided externally as already built.

@ehsavoie this configuration mechanism was not designed to be "changed" after the tracer is created. Generally, a tracer is created only once per application. I don't understand what use case you have in mind in the ticket. "Reloading" the application would create a new Config object anyway.

ehsavoie:
@yurishkuro the sender is being cached once it is resolved
I'm using the client in WildFly where the tracer is injected into a application but the configuration can be 'shared': for example to have several wars tracing to a single jaeger instance.

ehsavoie:
@yurishkuro tbh I can leave the cached tracer in place if that's your concern, as long as the sender is not cached I can live with it since I'm using a tracerBuilder

yurishkuro:
> where the tracer is injected into a application but the configuration can be 'shared': for example to have several wars tracing to a single jaeger instance.

Shouldn't each application have a distinct service name, which necessitates a separate config?

And couldn't you simply clone the config object from a shared one before calling getTracer()?

yurishkuro:
I'm ok with removing caching of the sender, it looks like the comment 
```
// A custom sender set by our consumers. If set, nothing else has effect. Optional.
```
is misleading, since there is no setter for that private field.

pavolloffay:
> @pavolloffay the getTracer() method no longer returns cached tracer, that is a breaking change. To my knowledge there was no other "caching" behavior, e.g. a sender is not cached, but can be provided externally as already built.

I am saying that tracer instance returned by `Configuration.getTracer()` _should be cached_. But It does not make sense to cache `Configuration.getTracerBuilder()` - it already returns a new tracer builder but the class caches `SenderConfiguration` and other sub-configuration classes.

pavolloffay:
@yurishkuro could you please have a look? It looks good to go.

PR #687, opened 2020-02-12
Revert "Unregister shutdown hook when tracer is closed (#678)"

TomasHofman:
This reverts commit ed5a8c58b91ac62cfccc789dd096a87c4e81c0b8.

Upstream commit: https://github.com/jaegertracing/jaeger-client-java/pull/688

This reverts fix for issue https://github.com/jaegertracing/jaeger-client-java/issues/677

The fix is causing issues when JaegerTracer#close is called when shutdown is already in progress - leading to IllegalStateException:

```
 java.lang.IllegalStateException: Shutdown in progress
	at java.lang.ApplicationShutdownHooks.remove(ApplicationShutdownHooks.java:82)
	at java.lang.Runtime.removeShutdownHook(Runtime.java:240)
	at io.jaegertracing.internal.JaegerTracer.close(JaegerTracer.java:229)
```

The memory leak itself, which the reverted commit intended to avoid, should not be a problem for regular applications as the JaegerTracer is designed to work as a singleton and the applications are not supposed to create more instances. Application servers can avoid it by other means.

Just revert previous commit.

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
@@ -77,7 +77,6 @@ public class JaegerTracer implements Tracer, Closeable {
   @ToString.Exclude private final BaggageSetter baggageSetter;
   @ToString.Exclude private final JaegerObjectFactory objectFactory;
   @ToString.Exclude private final int ipv4; // human readable representation is present within the tag map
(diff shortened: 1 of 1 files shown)


TomasHofman:
@yurishkuro I'm sorry, I discovered a problem with my fix. It should better be rolled back.

TomasHofman:
Superseded by https://github.com/jaegertracing/jaeger-client-java/pull/690

PR #688, opened 2020-02-12
Revert "Unregister shutdown hook when tracer is closed (#679)"

TomasHofman:
This reverts commit 4bde959f18fe0302968cc391075cf97dd5aecdcc.

This reverts fix for issue https://github.com/jaegertracing/jaeger-client-java/issues/677

The fix is causing issues when JaegerTracer#close is called when shutdown is already in progress - leading to IllegalStateException:

```
 java.lang.IllegalStateException: Shutdown in progress
	at java.lang.ApplicationShutdownHooks.remove(ApplicationShutdownHooks.java:82)
	at java.lang.Runtime.removeShutdownHook(Runtime.java:240)
	at io.jaegertracing.internal.JaegerTracer.close(JaegerTracer.java:229)
        ...
```

The memory leak itself, which the reverted commit intended to avoid, should not be a problem for regular applications as the JaegerTracer is designed to work as a singleton and the applications are not supposed to create more instances. Application servers can avoid it by other means.

Just revert previous commit.

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
@@ -79,7 +79,6 @@ public class JaegerTracer implements Tracer, Closeable {
   @ToString.Exclude private final BaggageSetter baggageSetter;
   @ToString.Exclude private final JaegerObjectFactory objectFactory;
   @ToString.Exclude private final int ipv4; // human readable representation is present within the tag map
-  @ToString.Exclude private Thread shutdownHook;
 
   protected JaegerTracer(JaegerTracer.Builder builder) {
     this.serviceName = builder.serviceName;
@@ -130,13 +129,12 @@ public class JaegerTracer implements Tracer, Closeable {
       log.info("No shutdown hook registered: Please call close() manually on application shutdown.");
     } else {
       // register this tracer with a shutdown hook, to flush the spans before the VM shuts down
-      shutdownHook = new Thread() {
+      Runtime.getRuntime().addShutdownHook(new Thread() {
         @Override
         public void run() {
           JaegerTracer.this.close();
         }
-      };
-      Runtime.getRuntime().addShutdownHook(shutdownHook);
+      });
     }
   }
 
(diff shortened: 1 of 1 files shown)


jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
```diff
@@ -130,13 +129,12 @@ protected JaegerTracer(JaegerTracer.Builder builder) {
       log.info("No shutdown hook registered: Please call close() manually on application shutdown.");
     } else {
       // register this tracer with a shutdown hook, to flush the spans before the VM shuts down
-      shutdownHook = new Thread() {
+      Runtime.getRuntime().addShutdownHook(new Thread() {
         @Override
         public void run() {
           JaegerTracer.this.close();
```
yurishkuro: I think the proper fix is to set `shutdownHook = null` before calling close().

TomasHofman: If I understand your suggestion that wouldn't help.

The problem is that when `JaegerTracer#close()` is triggered by shutdown hook (i.e. shutdown is already in progress), then calling `Runtime.removeShutdownHook(shutdownHook)` will fail with IllegalStateException. I.e. we cannot remove hooks when shutdown is already in progress.

Hm, or I could just catch the IllegalStateException, I don't know why I didn't thought of that before :/.

TomasHofman: I will prepare another PR where the exception is caught.

TomasHofman:
Superseded by https://github.com/jaegertracing/jaeger-client-java/pull/689

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
yurishkuro: I think it _will_ help. The sequence would be:
  * System.exit()
  * shutdown hook invoked
  * before closing the tracer, we reset `shutdownHook = null`
  * tracer.Close() won't attempt to remove the hook

TomasHofman: OK, you are right. I'm gonna update currently opened PRs.

PR #690, opened 2020-02-13
Prevent a shutdown hook from being removed in JaegerTracer when a shutdown is in progress

TomasHofman:
Prevent a shutdown hook from being removed in JaegerTracer when a shutdown is in progress

This is follow up PR for https://github.com/jaegertracing/jaeger-client-java/pull/679

The problem this PR is solving that when a JaegerTracer is closed as a result of shutdown, the attempt to remove its' close shutdown hook fails with IllegalStateException. I.e. the shutdown hook cannot be removed when shutdown is already in progress.

Do not remove a shutdown hook when a shutdown is already in progress.

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
@@ -131,6 +131,7 @@ public class JaegerTracer implements Tracer, Closeable {
       shutdownHook = new Thread() {
         @Override
         public void run() {
+          shutdownHook = null;
           JaegerTracer.this.close();
         }
       };


jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
```diff
@@ -226,7 +226,11 @@ public void close() {
     reporter.close();
     sampler.close();
     if (shutdownHook != null) {
-      Runtime.getRuntime().removeShutdownHook(shutdownHook);
+      try {
+        Runtime.getRuntime().removeShutdownHook(shutdownHook);
```
yurishkuro: please see https://github.com/jaegertracing/jaeger-client-java/pull/688#discussion_r378923234

yurishkuro:
looks like a duplicate of #689

TomasHofman:
@yurishkuro this is for 0.34 stream. The original commit is in 0.34 too.

Issue #380, opened 2018-04-08
Flaky test affecting code coverage


yurishkuro:
PR #379 was a change in the README but showed a drop in [code coverage][1] due to exception clauses not being hit:

![image](https://user-images.githubusercontent.com/3523016/38470532-2551475e-3b32-11e8-8d4a-930161414710.png)

[1]: https://codecov.io/gh/jaegertracing/jaeger-client-java/compare/6d822c3471896c7c3c4ced14c969b51bacaf3994...c6a597fb5bd1da9afbb2f8f85b3dd389d21d8bea/changes

piyushmor:
Hi, just raised a PR for the first coverage issue!

adnxn:
#658 and #692 addresses the two coverage gaps. i think we're good to close this issue.

yurishkuro:
Thanks!

PR #692, opened 2020-02-26
Improve RemoteReporter code coverage (#380)

adnxn:
- #380, specifically the changes in `RemoteReporter.java`

- added a new test

### DIFF
--- a/jaeger-core/src/test/java/io/jaegertracing/internal/reporters/RemoteReporterTest.java
+++ b/jaeger-core/src/test/java/io/jaegertracing/internal/reporters/RemoteReporterTest.java
@@ -368,6 +368,31 @@ public class RemoteReporterTest {
     assertEquals(0, metricsFactory.getCounter("jaeger_tracer_reporter_spans", "result=dropped"));
   }
 
+  @Test
+  public void testUpdateErrorMetricWhenCommandExecuteFails() throws Exception {
+
+    int reporterFailures = 5;
+
+    sender = new InMemorySender() {
+      @Override
+      public int append(JaegerSpan span) throws SenderException {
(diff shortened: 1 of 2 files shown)


jaeger-core/src/test/java/io/jaegertracing/internal/reporters/RemoteReporterTest.java
```diff
@@ -368,6 +368,31 @@ public int append(JaegerSpan span) {
...
   }
 
+  @Test
+  public void testUpdateErrorMetricWhenCommandExecuteFails() throws Exception {
+
+    int reporterFailures = 5;
+
+    sender = new InMemorySender() {
+      @Override
+      public int append(JaegerSpan span) throws SenderException {
```
pavolloffay: Can be a runtime exception used to avoid the signature change?

adnxn: so the [codepath](https://github.com/jaegertracing/jaeger-client-java/blob/5e81d89303d3a670e09664413a52fdba23783cbf/jaeger-core/src/main/java/io/jaegertracing/internal/reporters/RemoteReporter.java#L105) that we want to cover and called out in #380 has a catch on `SenderException`

pavolloffay: +1 just posting sender interface

```
public interface Sender {
  int append(JaegerSpan span) throws SenderException;

  int flush() throws SenderException;

  int close() throws SenderException;
}

```

pavolloffay:
@adnxn thanks for the PR!

PR #698, opened 2020-03-23
Avoid warning if traceparent not present

quaff:
TextMapCodec.extract() return null with no warning if SPAN_CONTEXT_KEY not present, TraceContextCodec should align with it.
After switch `TextMapCodec` to `TraceContextCodec`, my application show lots of  [warning](https://github.com/jaegertracing/jaeger-client-java/blob/568ab68ec2ac5e9e811b1d0d76f752f3242d45d3/jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TraceContextCodec.java#L73) .
I have to change my code to suppress it.
```
SpanContext parent = tracer.extract(Format.Builtin.HTTP_HEADERS, new HttpServletRequestTextMap(request));
```
add ugly null check
```
SpanContext parent = request.getHeader("traceparent") != null ? tracer.extract(Format.Builtin.HTTP_HEADERS, new HttpServletRequestTextMap(request)) : null;
```

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TraceContextCodec.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TraceContextCodec.java
@@ -111,6 +111,9 @@ public class TraceContextCodec implements Codec<TextMap> {
         traceState = entry.getValue();
       }
     }
+    if (traceParent == null) {
+      return null;
+    }
     return extractContextFromTraceParent(traceParent, traceState);
   }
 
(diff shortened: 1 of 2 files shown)


quaff:
> please add a unit test

It's already covered by unit test.
https://github.com/jaegertracing/jaeger-client-java/blob/master/jaeger-core/src/test/java/io/jaegertracing/internal/propagation/TraceContextCodecTest.java#L96

yurishkuro:
Yes, but you are adding new behavior and a new expectation that no logs are produced on missing traceparent. That's not covered by the existing unit test.

pavolloffay:
@quaff could you please fix the CI?

pavolloffay:
thanks @quaff

quaff:
I admit the unit test is a bit fragile, It depends gradle, since gradle reassigned `System.out`, maybe there is a better way to detect log message, please improve it. @pavolloffay

Issue #699, opened 2020-03-24
Supports "jaeger-debug-id" in W3C TraceContextCodec


quaff:
## Requirement - what kind of business use case are you trying to solve?
Replace `TextMapCodec` with `TraceContextCodec`

## Problem - what in Jaeger blocks you from solving the requirement?
`jaeger-debug-id` is not extracted in `TraceContextCodec`

## Proposal - what do you suggest to solve the problem or improve the existing situation?
Align `TraceContextCodec` with `TextMapCodec`

pavolloffay:
@quaff would you like to submit a PR?

PR #701, opened 2020-03-24
Supports jaeger-debug-id in TraceContextCodec

quaff:
Fix https://github.com/jaegertracing/jaeger-client-java/issues/699

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TraceContextCodec.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TraceContextCodec.java
@@ -16,6 +16,7 @@
 
 package io.jaegertracing.internal.propagation;
 
+import io.jaegertracing.internal.Constants;
 import io.jaegertracing.internal.JaegerObjectFactory;
 import io.jaegertracing.internal.JaegerSpanContext;
 import io.jaegertracing.spi.Codec;
@@ -58,7 +59,7 @@ public class TraceContextCodec implements Codec<TextMap> {
     this.objectFactory = builder.objectFactory;
   }
 
-  private JaegerSpanContext extractContextFromTraceParent(String traceparent, String tracestate) {
+  private JaegerSpanContext extractContextFromTraceParent(String traceparent, String tracestate, String debugId) {
     // TODO(bdrutu): Do we need to verify that version is hex and that
     // for the version the length is the expected one?
     boolean isValid =
@@ -95,7 +96,7 @@ public class TraceContextCodec implements Codec<TextMap> {
         spanId,
         0,
         sampled ? (byte) 1 : (byte) 0,
-        Collections.<String, String>emptyMap(), null);
+        Collections.<String, String>emptyMap(), debugId);
     return spanContext.withTraceState(tracestate);
   }
 
@@ -103,6 +104,7 @@ public class TraceContextCodec implements Codec<TextMap> {
   public JaegerSpanContext extract(TextMap carrier) {
     String traceParent = null;
     String traceState = null;
+    String debugId = null;
     for (Map.Entry<String, String> entry: carrier) {
       if (TRACE_PARENT.equals(entry.getKey())) {
         traceParent = entry.getValue();
@@ -110,11 +112,17 @@ public class TraceContextCodec implements Codec<TextMap> {
       if (TRACE_STATE.equals(entry.getKey())) {
         traceState = entry.getValue();
       }
+      if (Constants.DEBUG_ID_HEADER_KEY.equals(entry.getKey())) {
+        debugId = entry.getValue();
+      }
     }
     if (traceParent == null) {
+      if (debugId != null) {
+        return objectFactory.createSpanContext(0L, 0L, 0L, 0L, (byte) 0, null, debugId);
+      }
       return null;
     }
-    return extractContextFromTraceParent(traceParent, traceState);
+    return extractContextFromTraceParent(traceParent, traceState, debugId);
   }
 
   @Override
--- a/jaeger-core/src/test/java/io/jaegertracing/internal/propagation/TraceContextCodecTest.java
+++ b/jaeger-core/src/test/java/io/jaegertracing/internal/propagation/TraceContextCodecTest.java
@@ -17,19 +17,27 @@ package io.jaegertracing.internal.propagation;
 import static io.jaegertracing.internal.propagation.TraceContextCodec.TRACE_PARENT;
 import static io.jaegertracing.internal.propagation.TraceContextCodec.TRACE_STATE;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.anyInt;
+import static org.mockito.Mockito.atLeast;
 import static org.mockito.Mockito.reset;
 import static org.mockito.Mockito.times;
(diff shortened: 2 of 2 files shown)


jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
```diff
@@ -135,7 +135,7 @@ public boolean isSampled() {
   }
 
   public boolean isDebug() {
-    return (flags & flagDebug) == flagDebug;
+    return (flags & flagDebug) == flagDebug || debugId != null;
```
pavolloffay: Why this check? Can we keep this class without changes?

yurishkuro: isDebug is decided purely on the flag, whether there is a correlation identifier is irrelevant.

jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TraceContextCodec.java
```diff
@@ -95,26 +96,34 @@ private JaegerSpanContext extractContextFromTraceParent(String traceparent, Stri
...
+        debugId = entry.getValue();
+      }
     }
     if (traceParent == null) {
-      return null;
+      if (debugId != null) {
+        return objectFactory.createSpanContext(0L, 0L, 0L, 0L, (byte) 0, null, debugId);
+      } else {
+        return null;
+      }
```
yurishkuro:
```suggestion
      }
      return null;
```

`else` is redundant

jaeger-core/src/test/java/io/jaegertracing/internal/propagation/TraceContextCodecTest.java
```diff
@@ -160,10 +164,31 @@ public void testEmptyTraceStateNotPropagated() {
     assertEquals(EXAMPLE_TRACE_PARENT, injectCarrier.get(TRACE_PARENT));
   }
 
+  @Test
+  public void testDebugIdWithoutTraceHeader() {
+    Map<String, String> extractCarrier = new HashMap<>();
+    TextMapAdapter textMap = new TextMapAdapter(extractCarrier);
+    textMap.put(Constants.DEBUG_ID_HEADER_KEY, EXAMPLE_DEBUG_ID);
+    JaegerSpanContext spanContext = traceContextCodec.extract(textMap);
+    assertTrue(spanContext.isDebug());
```
yurishkuro: the context itself is not "debug". Instead, you can check that it contains the correlation ID.

jaeger-core/src/test/java/io/jaegertracing/internal/propagation/TraceContextCodecTest.java
```diff
@@ -160,10 +164,31 @@ public void testEmptyTraceStateNotPropagated() {
...
+  }
+
   private void verifyWarningPresent() {
     ArgumentCaptor<byte[]> captor = ArgumentCaptor.forClass(byte[].class);
     try {
-      verify(System.out).write(captor.capture());
+      verify(System.out, atLeast(0)).write(captor.capture(), anyInt(), anyInt());
+      // gradle will call write(byte[]) instead of write(byte[],off,len)
```
yurishkuro: What does "gradle" have to do with unit test?

jaeger-core/src/test/java/io/jaegertracing/internal/propagation/TraceContextCodecTest.java
```diff
@@ -160,10 +164,31 @@ public void testEmptyTraceStateNotPropagated() {
...
+  }
+
+  @Test
+  public void testDebugIdWithTraceHeader() {
+    Map<String, String> extractCarrier = new HashMap<>();
+    TextMapAdapter textMap = new TextMapAdapter(extractCarrier);
+    textMap.put(TRACE_PARENT, EXAMPLE_TRACE_PARENT);
+    textMap.put(Constants.DEBUG_ID_HEADER_KEY, EXAMPLE_DEBUG_ID);
+    JaegerSpanContext spanContext = traceContextCodec.extract(textMap);
+    assertTrue(spanContext.isDebug());
```
yurishkuro: Same as above.

What we really want to test is whether the tracer will be starting spans with debug flag when given this span context as a parent. But such test should be in the TracerTests.

quaff:
ping

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpanContext.java
```diff
@@ -194,7 +194,7 @@ boolean hasTrace() {
    *
    * @see Constants#DEBUG_ID_HEADER_KEY
    */
-  String getDebugId() {
+  public String getDebugId() {
```
yurishkuro, line 197: Please do not expand public API of this class.

jaeger-core/src/test/java/io/jaegertracing/internal/propagation/TraceContextCodecTest.java
```diff
@@ -160,10 +164,31 @@ public void testEmptyTraceStateNotPropagated() {
     assertEquals(EXAMPLE_TRACE_PARENT, injectCarrier.get(TRACE_PARENT));
   }
 
+  @Test
+  public void testDebugIdWithoutTraceHeader() {
+    Map<String, String> extractCarrier = new HashMap<>();
+    TextMapAdapter textMap = new TextMapAdapter(extractCarrier);
+    textMap.put(Constants.DEBUG_ID_HEADER_KEY, EXAMPLE_DEBUG_ID);
+    JaegerSpanContext spanContext = traceContextCodec.extract(textMap);
+    assertEquals(EXAMPLE_DEBUG_ID, spanContext.getDebugId());
```
yurishkuro, line 173: you can mock `objectFactory` and verify that the debug ID is passed this way, instead of exposing a new method on the context. Or you can create a tracer with this codec, start a span, and verify that span has the tag recorded.

jaeger-core/src/test/java/io/jaegertracing/internal/propagation/TraceContextCodecTest.java
```diff
@@ -160,10 +164,31 @@ public void testEmptyTraceStateNotPropagated() {
...
+  }
+
+  @Test
+  public void testDebugIdWithTraceHeader() {
+    Map<String, String> extractCarrier = new HashMap<>();
+    TextMapAdapter textMap = new TextMapAdapter(extractCarrier);
+    textMap.put(TRACE_PARENT, EXAMPLE_TRACE_PARENT);
+    textMap.put(Constants.DEBUG_ID_HEADER_KEY, EXAMPLE_DEBUG_ID);
+    JaegerSpanContext spanContext = traceContextCodec.extract(textMap);
+    assertEquals(EXAMPLE_DEBUG_ID, spanContext.getDebugId());
```
yurishkuro, line 183: Technically, this test is identical to the previous one in terms of asserts. It's worth verifying that the TRACE_PARENT was also parsed correctly, e.g. by asserting the trace ID.

jaeger-core/src/test/java/io/jaegertracing/internal/propagation/TraceContextCodecTest.java
yurishkuro: ?

quaff: gradle output log via `write(byte[])` and eclipse output log via `write(byte[],int,int)` , before this change, tests failed with eclipse junit runner. I know this is fragile, but I cannot find a better way to assert log message.

yurishkuro:
Thanks!

Issue #624, opened 2019-05-08
Support of 'Defer' sampling state


krabaey:
## Requirement - what kind of business use case are you trying to solve?

We would like to set http tracing headers (in our case B3 compatible headers) on a proxy, but defer the sampling decision to the actual servers it is proxying. The proxy however is not capable of doing reporting. 
In this way the b3 headers could be centrally generated & injected. 

Currently this does not seem not possible.

## Problem - what in Jaeger blocks you from solving the requirement?

According to [zipkin-b3-propagation](https://github.com/apache/incubator-zipkin-b3-propagation#sampling-state) sampling is a state with 4 possible values: Defer,Deny,Accept,Debug.

In JaegerSpanContext this is modeled into two booleans: sampled and debug.
The consequence is that the B3TextMapCodec cannot report the 'defer' sampling state.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Adapt the JaegerSpanContext & B3TextMapCodec to take into account the possible values of the sampling state so defering the sampling desision is supported.

## Any open questions to address

None

yurishkuro:
> set http tracing headers (in our case B3 compatible headers) on a proxy

can you elaborate for what reason?

krabaey:
> > set http tracing headers (in our case B3 compatible headers) on a proxy
> 
> can you elaborate for what reason?

All access to our services is via a proxy (ha-proxy in this case). As part of improving observability we use the trace id to correlate all logging of all services for a particular trace. As the proxy is part of the infrastructure, we want to include the trace id in the proxy access log. Since the proxy is our first component in the chain, this one will create the trace id (if none is present).  

An unintended side effect of injecting the X-B3-TraceId and X-B3-SpanId header is that no reporting of traces is done anymore by the other components, as Jaeger in this case interprets the absence of the X-B3-Sampled header as an instruction to not sample the trace.

The reason of not taking the decision to sample on the proxy is twofold:

- integration of the 'sampling decision' module would be challenging as I don't know of any existing integration. 
- we want the decision of sampling to be taken by the specific service as for some services we may take a different decision to sample or not per service (e.g. high load services, critical services, ...)

yurishkuro:
I'm not opposed to it if you're interested to contribute the enhancement.

PR #707, opened 2020-04-25
Support sampling decision in B3-protocol

lujiajing1126:
Signed-off-by: megrez <lujiajing1126@gmail.com>

Focus on #708 

Also related to #624 

According to the [B3 propagation protocol](https://github.com/openzipkin/b3-propagation),
client can send header "X-B3-Sampled: 0" without any TraceId and ParentId fields.

The sampling decision will be respected if no other trace exists.

- Change `B3TextMapCodec` to support passing `X-B3-Sampled` only.
- Change `JaegerTracer` to support this feature.
- Necessary test cases are added to cover all code paths.
- In `jaeger` tracer, there is no `SamplingSet` mark in the flag, so it is impossible to determine the sampling status. In this PR, I also add a flag `1 << 2` to mark whether the sampling decision has been made.

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpan.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpan.java
@@ -223,9 +223,12 @@ public class JaegerSpan implements Span {
       int priority = ((Number) value).intValue();
       byte newFlags;
       if (priority > 0) {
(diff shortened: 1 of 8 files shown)


lujiajing1126:
> Adding a new bit to the context flags is a major decision that affects all Jaeger clients and needs to be thoroughly documented for its semantic meaning and expected behavior. Please open a ticket in the main Jaeger repo with the detailed proposal.

I have opened an issue in the main repository, we may discuss the details there.

Issue #710, opened 2020-05-04
Weird reported span duration


sheinbergon:
Hello. 

We are trying to evaluate jaeger-tracing for tracing our various applications.
We are managing spans our selves (via a self-assigned ID), and both starting it and logging events using a self-deduced timestamps in micros-seconds, which seem to be valid.

However, reported span duration is literally off the charts.

Here's a sample span json:

```{
  "data": [
    {
      "traceID": "3f86c53578807ac2",
      "spans": [
        {
          "traceID": "3f86c53578807ac2",
          "spanID": "3f86c53578807ac2",
          "flags": 1,
          "operationName": "market-message-consumption",
          "references": [
            
          ],
          "startTime": 1588586480766066,
          "duration": 18446744073709551550,
          "tags": [
            {
              "key": "span-id",
              "type": "string",
              "value": "369746"
            },
            {
              "key": "sampler.type",
              "type": "string",
              "value": "ratelimiting"
            },
            {
              "key": "sampler.param",
              "type": "float64",
              "value": 1000
            },
            {
              "key": "internal.span.format",
              "type": "string",
              "value": "jaeger"
            }
          ],
          "logs": [
            {
              "timestamp": 1588586480766080,
              "fields": [
                {
                  "key": "event-id",
                  "type": "string",
                  "value": "message-encoded"
                }
              ]
            },
            {
              "timestamp": 1588586480766111,
              "fields": [
                {
                  "key": "event-id",
                  "type": "string",
                  "value": "message-dequeued"
                }
              ]
            },
            {
              "timestamp": 1588586480766191,
              "fields": [
                {
                  "key": "event-id",
                  "type": "string",
                  "value": "message-produced"
                }
              ]
            }
          ],
          "processID": "p1",
          "warnings": null
        }
      ],
      "processes": {
        "p1": {
          "serviceName": "sivron-feed-test",
          "tags": [
            {
              "key": "hostname",
              "type": "string",
              "value": "crl8-devel-desktop"
            },
            {
              "key": "ip",
              "type": "string",
              "value": "127.0.1.1"
            },
            {
              "key": "jaeger.version",
              "type": "string",
              "value": "Java-1.2.0"
            }
          ]
        }
      },
      "warnings": null
    }
  ],
  "total": 0,
  "limit": 0,
  "offset": 0,
  "errors": null
}```

Any idea what might be causing the span duration to look like this?

10x in advance

yurishkuro:
[...]

Issue #192, opened 2017-05-31
Side effects of imprecise clock


mabn:
I found an interesting problem that took some time to track down.

Basically spans in the jaeger UI were "centered" in a weird way. Initially I thought that it's some UI issue, but after comparing the json returned by jaeger-query with cassandra it turned out that jaeger-query completely messes span start times. And by completely I mean that span taking 1ms was shifted by 3ms. Here's how it looks like:

![Alt text](https://monosnap.com/file/YZPE6yqX0wzLVQ0lDTcY6ake5DgmZa.png)

In this example "child 2" should happen as the last span inside the "parent" - should be stuck to the right edge of the parent. There's no delay because all those spans are reported by the same process.

After digging through the jaeger code I found `ClockSkew` and there are two problems here:
1. `ClockSkew` adjust span start times if they overlap with their parent. It does it only if parent and child are reported by different hosts, but the host is taken from "ip" tag of the process. If it's missing the default is to assume that these are different hosts. 
2. jaeger-client-java by default uses `SystemClock` which has millisecond precision, but timestamps reported to jaeger are in microseconds. To convert the time to micros it multiplies it by 1000:
    ```java
    public long currentTimeMicros() {  return System.currentTimeMillis() * 1000; }
    ```
    One can think of it as rounding down exact start time to milliseconds.

    Span duration on the other hand is calculated with more precise `System.nanoTime`. 
  As a result of this rounding it's possible that:
    ```
    round(parent.start) + parent.duration < round(child.start) + child.duration
    ```
    for children which end at the same time as the parent it happens very often.

So - child span end time exceeds parent's finish time by between 0 and 1000 micros which in combination with missing process "ip" tag triggers clock skew adjuster. And for very long parents and very short children the adjustment is drastic (around 50% of parent duration).

There are two options:
1. change the default in jaeger-query
2. better clock implementation in jaeger-client-java
3. add "ip" process tag by default

I guess that the default make sense so better clock implementation would be a good improvement.

yurishkuro:
Nice catch.

Are you using java client that reports jaeger-thrift model (recently released)? It should not have the problem with "ip" tag.

The behavior of clock skew adjustment aside, the real issue seems to be the difference in precision between startTime and duration. Unfortunately, I don't know of a good way to get accurate startTime precision. The System.nanoSeconds() we're using for duration is not an offset from epoch, but from some random point in time, so not suitable for startTime.

mabn:
[...]

PR #712, opened 2020-05-20
Real microseconds timestamp accuracy for JDK 9 and above

sheinbergon:
This PR addresses microseconds timestamp issues described in #710


jaeger-core/src/main/java/io/jaegertracing/internal/clock/SystemClock.java
```diff
@@ -14,19 +14,45 @@
...
+
+  private static CurrentTimeSupplier getCurrentTimeSupplier() {
+    return MICROS_ACCURATE
+        ? CurrentTimeSupplier.MicrosAccuracy.INSTANCE
+        : CurrentTimeSupplier.MillisAccuracy.INSTANCE;
+  }
+
+  static {
+    val version = getJavaVersion();
+    MICROS_ACCURATE = version >= 9;
```
quaff, line 49: What about 
```java
try {
  Class.forName("java.lang.StackWalker");
  MICROS_ACCURATE = true;
} catch( ClassNotFoundException e ) {
  MICROS_ACCURATE = false;
}
```

jaeger-core/src/main/java/io/jaegertracing/internal/clock/SystemClock.java
```diff
@@ -14,19 +14,45 @@
...
+    CURRENT_TIME_SUPPLIER = getCurrentTimeSupplier();
+  }
+
   @Override
   public long currentTimeMicros() {
-    return System.currentTimeMillis() * 1000;
+    return CURRENT_TIME_SUPPLIER.micros();
```
quaff, line 55:
```java
Instant now = Instant.now();
return now.atZone(ZoneId.systemDefault()).toEpochSecond() * 1000000 + now.getNano() / 1000;
```
Both JDK8 and JDK9+ are supported.

jaeger-core/src/main/java/io/jaegertracing/internal/clock/SystemClock.java
sheinbergon: Implict assumption of the JDK version by the existence of a class (which is more prone to changes) as opposed to literally and explicitly checking the condition we wish to validate - the actual JDK version, is less than optimal, and at least in my opinion, a poor practice.

As this is not my code-base, I'd do whatever the maintainer wishes, but I vote neigh.

jaeger-core/src/main/java/io/jaegertracing/internal/clock/SystemClock.java
sheinbergon: The code you presented is wrong.
`getNano()` returns a monotonous clock value, not an RTC clock.
microseconds (maximaul) resolution is supported only from JDK 9 and above.

You can see additional information in the original issue https://github.com/jaegertracing/jaeger-client-java/issues/710 and also search on stack overflow.

Closing this discussion

jaeger-core/src/main/java/io/jaegertracing/internal/clock/CurrentTimeSupplier.java
```diff
@@ -0,0 +1,103 @@
...
+    static final CurrentTimeSupplier INSTANCE = new MicrosAccuracy();
+
+    private static final Method NOW;
+    private static final Object EPOCH;
+    private static final Object CHRONO_UNIT_MICROS;
+    private static final Method CHRONO_UNIT_BETWEEN;
+
+    static {
+      try {
+        val classLoader = ClassLoader.getSystemClassLoader();
```
pavolloffay, line 78: Can we use FQCN instead of reflection? Maybe it's better to move this all together to a separate file and load only if the instant exists or the java 9 is present. 

It seems redundant to be using reflection for the instant class and also `java.version` can we unify it and use only one?

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
```diff
@@ -249,6 +251,20 @@ protected SpanBuilder(String operationName) {
...
+        digits++;
+        subject *= 10;
+      }
+      if (digits >= DIGITS_IN_MICROSECOND_TIMESTAMP) {
+        throw new IllegalArgumentException(String
+            .format("'startTimeMicroseconds' value %d contains only %d digits, while a minimum of %d is expected",
+                startTimeMicroseconds, digits, DIGITS_IN_MICROSECOND_TIMESTAMP));
+      }
+    }
+
```
ilylia, line 267: is this necessary?
if yes, shell we just check`startTimeMicroseconds ` with `1e16`, instead of a loop check?

jaeger-core/src/main/java/io/jaegertracing/internal/clock/CurrentTimeSupplier.java
sheinbergon: @pavolloffay I agree, it would  have been much better to go multi release.
However, it'll require additional changes to code layout and gradle definitions. and might create development inconvenience for IDE users (depends on the IDE and it's understanding of  gradle `sourceSets`/`configurations`.

Again this is not my codebase - I'll need a strong clear decision on what's acceptable as part of the scope of this PR.

let me know.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
sheinbergon: The test is necessary because otherwise you get weird unsupported behavior and I think it's a good idea to resolve it.
As for your suggestion - you're right - I'll fix it.

jaeger-core/src/main/java/io/jaegertracing/internal/clock/CurrentTimeSupplier.java
pavolloffay: We don't want to do multi-release nor mess up with grade config.

My suggestion is to just move the jdk9 code to a separate class. The client code would load the class only when running on JDK9 - checked by `java.version` or the presence of JDK9 classes. This should work well as the class would be loaded only when referenced.

sheinbergon: So just to clarify - keep the code as it is (reflection based access to newer classes ) , but separate different use cases to different classes?

pavolloffay: yes

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
sheinbergon: Fixed

jaeger-core/src/main/java/io/jaegertracing/internal/clock/CurrentTimeSupplier.java
sheinbergon: Fixed

sheinbergon:
@pavolloffay  @ilylia I fixed both of you rejects, how do we proceed from here?

sheinbergon:
@ilylia  @pavolloffay  any news about finalizing this contribution?

ilylia:
LGTM.
but you still need to fix the ci errors     ╮(╯_╰)╭
and wait for @pavolloffay 's approval.

sheinbergon:
> LGTM.
> but you still need to fix the ci errors ╮(╯_╰)╭
> and wait for @pavolloffay 's approval.

I fixed the travis issues, turns out some tests were no failing as they were getting bogus start timestamps instead of valid ones.

Added another test to cover the invalid start timestamp case

However codecov now complains on coverage on the JDK 9+ specific code.
There's no way to run that code in the current CI build. And I don't think trying to mock the very thing  we're testing would make any sense.

How'd you like to proceed about this?

ilylia:
sorry, I have no idea about this.
Maybe @pavolloffay can give you some suggestions.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
```diff
@@ -462,6 +473,8 @@ public JaegerSpan start() {
           startTimeNanoTicks = clock.currentNanoTicks();
           computeDurationViaNanoTicks = true;
         }
+      } else {
+        verifyStartTimeInMicroseconds();
```
yurishkuro, line 477: Why do we need this? Starting a new span is never expected to throw exceptions.

sheinbergon:
> I have one high level concern, which has to do with the overall design of the Clock interface (see issue #192). There were a few proposals in that issue on how to fix the problem but nobody did a PR. I'm concerned that doubling down on the broken Clock interface to get higher precision is only making things worse overall. If people are not ok with the current precision but are ok with timestamps potentially going backwards, then they can always provide a custom Clock implementation with micro-precision (and not deal with Java version compatibility if their apps are already on Java9).

Well, that's kind of frustrating. I specifically asked @ilylia in https://github.com/jaegertracing/jaeger-client-java/issues/710 before approaching to writing this code for contribution if you'd be interested in receiving this specific contribution (microseconds support for the system clock on supported JVMs), and was responded with a definite **yes**. 

The state of things is that, you're java client library has timestamp precision bugs, up to a point where my first usage attempt triggered and sent me digging through your source code as to why I was getting negative durations. If that's a user-experience you're content with, well, then we can just close this PR. 

Yes, it's mostly the broken interface and the fact this library (and I'm guess also opentracing interfaces) still target 1.6 (disallowing the usage of containers such as `Instant`) while 1.8 had already seen a public EOL are the main cause, but getting microseconds support  `free-of-charge` (no structural/breaking changes to the library) seemed like a win to me.

At this point, I'm opting out. You have this contribution to do with as you see fit, if you want me to fix additional things in it or drop the verification for the valid values, I'll gladly do so but aside from that, It's up to you to accept or reject this PR.

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
sheinbergon: well, your API specifies passing a start timestamp in microseconds. and your internal API  logs events in microseconds (whether _fake_ or not) and does timestamp arithmetic based on this value . Thus, it makes no sense to allow a non micros value when passing supplied timestamps. This fix helps minimize the buggy behavior of the interface somehow by alerting the user he's doing something completely wrong. It'll either fail all of the times or none of times, once the user fixes his code.

yurishkuro: we can log an error, I am still against throwing exception

jaeger-core/src/main/java/io/jaegertracing/internal/clock/MillisAccurrateClock.java
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright (c) 2016, Uber Technologies, Inc
```
yurishkuro, line 2:
```suggestion
 * Copyright (c) 2020, The Jaeger Authors
```

jaeger-core/src/main/java/io/jaegertracing/internal/clock/MicrosAccurateClock.java
```diff
@@ -0,0 +1,87 @@
+/*
+ * Copyright (c) 2016, Uber Technologies, Inc
```
yurishkuro, line 2:
```suggestion
 * Copyright (c) 2020, The Jaeger Authors
```

jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java
sheinbergon: Ok, I'll change this to only log a warning.

sheinbergon: Fixed

sheinbergon:
> I am ok to merge this if we replace the exception with a log

@yurishkuro fixed

PR #721, opened 2020-06-26
Do not encode span context in HTTP_HEADERS format

### LINKED ISSUE #720: Do not URL-encode uber-trace-id headerr
The colon-separated format of uber-trace-id is safe for HTTP headers, we should not be forcing URL encoding on it.

https://github.com/jaegertracing/jaeger-client-java/blob/016e0c33be6b5327220b4048461527500842a077/jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TextMapCodec.java#L129

cykl:

String representation of a span context is safe to use in HTTP header. Thus, there is no need to encode it. JS & Go clients
already avoid that unnecessary encoding. 

Resolves #720 

This change is backward compatible because all clients unencode span context even if they do not encode it. More pragmatically, a client impacted by this change would already be incompatible with JS & Go clients.

### DIFF
--- a/jaeger-core/src/test/java/io/jaegertracing/internal/propagation/TextMapCodecTest.java
+++ b/jaeger-core/src/test/java/io/jaegertracing/internal/propagation/TextMapCodecTest.java
@@ -158,4 +158,45 @@ public class TextMapCodecTest {
     assertEquals("v2", context.getBaggageItem("k2"));
     assertEquals(null, context.getBaggageItem("k3"));
   }
+
+  @Test
+  public void testInjectDoNotEncodeSpanContext() {
+    TextMapCodec codec = new TextMapCodec(true);
+    Map<String, String> headers = new HashMap<String, String>();
+    long traceIdLow = 42;
+    long spanId = 1;
+    long parentId = 0;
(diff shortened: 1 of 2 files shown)


jaeger-core/src/test/java/io/jaegertracing/internal/propagation/TextMapCodecTest.java
```diff
@@ -158,4 +158,45 @@ public void testAdhocBaggageWithoutTraceId() {
...
+    long spanId = 1;
+    long parentId = 0;
+    codec.inject(new JaegerSpanContext(0L, traceIdLow, spanId, parentId, (byte)1), new TextMapAdapter(headers));
+
+    String traceId = headers.get("uber-trace-id");
+    assertEquals("2a:1:0:1", traceId);
+  }
+
+  @Test
+  public void testExtractSupportEncodedSpanContext() {
```
yurishkuro, line 176: I think this and next method names should be swapped

cykl: Oops, fixed.

Issue #722, opened 2020-06-27
Publishing to Nexus is failing


yurishkuro:
Release 1.3.0 failed to publish with the following error:

> Invalid POM: /io/jaegertracing/jaeger-thrift/1.3.0/jaeger-thrift-1.3.0.pom: Project name missing, Project description missing, Project URL missing, License information missing, SCM URL missing, Developer information missing

cc @pavolloffay @jpkrohling

yurishkuro:
After running `./gradlew publishToMavenLocal -PskipSigning=true` and comparing these files

* ~/.m2/repository/io/jaegertracing/jaeger-core/1.3.1-SNAPSHOT/jaeger-core-1.3.1-SNAPSHOT.pom
* ~/.m2/repository/io/jaegertracing/jaeger-client/1.3.1-SNAPSHOT/jaeger-client-1.3.1-SNAPSHOT.pom
* ~/.m2/repository/io/jaegertracing/jaeger-thrift/1.3.1-SNAPSHOT/jaeger-thrift-1.3.1-SNAPSHOT.pom

only jaeger-thrift has very differently looking (almost empty) pom file

cc @zhalas

pavolloffay:
Could you please past the content of the jaeger-thrift? Maybe it does not inherit settings from the parent.

The issue is most likely in this PR https://github.com/jaegertracing/jaeger-client-java/pull/704 by @zhalas

yurishkuro:
Do you mean content of the generated `jaeger-thrift.pom`?

```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <modelVersion>4.0.0</modelVersion>
  <groupId>io.jaegertracing</groupId>
  <artifactId>jaeger-thrift</artifactId>
  <version>1.3.1-SNAPSHOT</version>
  <packaging>pom</packaging>
  <dependencies/>
</project>
```

zhalas:
The problem is that while publishing the shadow aritifact we also publish POM for it, and it's the shadow POM that is causing problems. It is indeed caused by #704. I'm trying to find some solution to this problem, but if it's blocking release revert may not be a bad idea.

yurishkuro:
[...]

PR #704, opened 2020-03-30
Fix publishing jaeger-thrift shadow artifact

zhalas:
## Which problem is this PR solving?
- Missing shadow artifact (#645) 

## Short description of the changes
- add publication and signing configuration for jaeger-thrift shadow artifact (this is currently not published as the current configuration only publishes `components.java` + sources + javadocs)

## Testing performed
- `./gradlew publishToMavenLocal -PskipSigning=true` and verified that the shadow artifact is deployed to local maven repository
- `./gradlew publishToMavenLocal` and observed that `:jaeger-thrift:signShadowPublication` fails (it's not a full test, but should be good enough)

PR #725, opened 2020-06-29
Fix publishing to Nexus

zhalas:
- Publishing to Nexus is failing (#722)

- consolidate publishing shadow artifact and the rest into a single task (shadow publishing task was trying to publish its own POM and that POM was messing up with the main POM)

## Testing performed
- Run `./gradlew jaeger-thrift:publishToMavenLocal` on master with #704 reverted and this branch and observed that:
  - `jaeger-thrift-1.3.1-SNAPSHOT-shadow.jar` is present only when publishing from this branch
  - `jaeger-thrift-1.3.1-SNAPSHOT.pom` looks the same when publishing form both branches and is sane in general
- Since this change is affecting all the things that use `com.github.johnrengelman.shadow`, I also verified that what is published by `jaeger-crossdock` looks reasonably sane, so I hope this won't break
- Run `publishToMavenLocal` on all subprojects and verified that there are no errors

### DIFF
--- a/gradle/publish.gradle
+++ b/gradle/publish.gradle
@@ -19,6 +19,9 @@ publishing {
             from components.java
             artifact sourcesJar
             artifact javadocJar
+            project.plugins.withId('com.github.johnrengelman.shadow') {
+                artifact project.tasks.shadowJar
+            }
 
             pom {
                 name = 'jaeger-client'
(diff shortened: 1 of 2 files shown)


yurishkuro:
I think this fixed the POM issue, but release is still failing, I commented on the issue.

PR #727, opened 2020-07-14
avoid holding too many span instances in memory

dmsolr:
-

I found a leaking memory issue when I do stress testing within LoggingReporter. In case, many Span instances are held on memory. 

In fact, I am not very sure about the key causes yet.  But this change seems available in my case. 

-


jpkrohling:
Looks like you might want to open an issue instead of a pull request ;-)

yurishkuro:
LoggingReporter is a debugging tool not meant to be used in production settings.

dmsolr:
Got it. Thank you both. :)

PR #735, opened 2020-08-13
Allow SystemClock to recognize non-GA OpenJDK versions

reta:

While testing Apache CXF with OpenJDK 16 Early Builds, we run into the issues with Java version parsing which `SystemClock` does under the hood:

```
Caused by: java.lang.NumberFormatException: For input string: "16-ea"
        at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:68)
        at java.base/java.lang.Integer.parseInt(Integer.java:652)
        at java.base/java.lang.Integer.parseInt(Integer.java:770)
        at io.jaegertracing.internal.clock.SystemClock.getJavaVersion(SystemClock.java:34)
        at io.jaegertracing.internal.clock.SystemClock.<clinit>(SystemClock.java:39)
        ... 27 more
```

It turned out that `SystemClock` is able to deal with GA releases only, this PR adds the support for non-GA builds with respect to `java.version` property as described in https://openjdk.java.net/jeps/223 (section `System properties`).

Thank you.

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/clock/SystemClock.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/clock/SystemClock.java
@@ -31,7 +31,9 @@ public class SystemClock implements Clock {
 
   private static int getJavaVersion() {
     val sections = System.getProperty("java.version").split("\\.");
-    val major = Integer.parseInt(sections[0]);
+    // Checking if major is in fact GA release or not (see please https://openjdk.java.net/jeps/223)
+    val index = sections[0].indexOf("-");
+    val major = Integer.parseInt((index == -1) ? sections[0] : sections[0].substring(0, index));
     return major == 1 ? Integer.parseInt(sections[1]) : major;
   }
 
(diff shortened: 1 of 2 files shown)


jaeger-core/src/main/java/io/jaegertracing/internal/clock/SystemClock.java
```diff
@@ -31,7 +31,9 @@
 
   private static int getJavaVersion() {
     val sections = System.getProperty("java.version").split("\\.");
```
yurishkuro, line 33: can we move everything after `System.getProperty("java.version")` into parseJavaVersion() and add a unit test for it running against a few allowed strings?

reta: :+1: absolutely

reta: @yurishkuro done, thanks for the review

jaeger-core/src/test/java/io/jaegertracing/internal/clock/SystemClockTest.java
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright (c) 2018, The Jaeger Authors
```
yurishkuro, line 2:
```suggestion
 * Copyright (c) 2020, The Jaeger Authors
```

reta: :+1: copy/paste ... thanks!

PR #739, opened 2020-08-28
Process sampling tag before handling other tags (fixes #738)

### LINKED ISSUE #738: Ignores some tags when you create a span with forced sampling

We have an application which we generally do not want tracing enabled, but we have certain user triggered workflows
where we'd like to enable tracing for.

The way we achieve this is for user initiated actions we build a new span (hence context) with 
`Tags.SAMPLING_PRIORITY` set to `1`, together with other tags specific to the users request.

There is no specific order in which the tags get processed and added to the context/span, as it's an iteration over a map:

https://github.com/jaegertracing/jaeger-client-java/blob/f1a7ca91af05206b3c47573689331a1fafec16de/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpan.java#L69-L71

This causes a problem, because tags are only set if the context is sampled

https://github.com/jaegertracing/jaeger-client-java/blob/f1a7ca91af05206b3c47573689331a1fafec16de/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpan.java#L234-L236

But the decision whether it's sampled or not comes from one of the tags, which can arrive later (due to iteration over a map)
Hence all tags prior to the sampling tag are dropped on the floor and ignored.

I think it should set the tags on the context unconditionally, or explicitly check for the sampling tag and do context adjustments 
before setting the rest of the tags.

## Sad workaround

Create a noop root span with sampling enabled, create a child span with actual user specific tags

AudriusButkevicius:
Fixes #738 
New test fails before this change.

Signed-off-by: Audrius Butkevicius <audrius.butkevicius@gmail.com>

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpan.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpan.java
@@ -66,8 +66,18 @@ public class JaegerSpan implements Span {
     this.tags = new HashMap<String, Object>();
     this.references = references != null ? new ArrayList<Reference>(references) : null;
 
+    // Handle SAMPLING_PRIORITY tag first, as this influences whether setTagAsObject actually
+    // sets other tags on the context or not, and we're not guaranteed to hit SAMPLING_PRIORITY first
+    // because we're iterating over a map.
+    Object samplingPriorityValue = tags.get(Tags.SAMPLING_PRIORITY.getKey());
+    if (samplingPriorityValue != null) {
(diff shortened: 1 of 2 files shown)


AudriusButkevicius:
Any ideas of when this will be released?

AudriusButkevicius:
Bump regarding my question.
There doesn't seem to be any information on the release cycles or when things are released. 

I want to setup a reminder to remove the current workaround we have.

yurishkuro:
https://github.com/jaegertracing/jaeger-client-java/releases/tag/v1.4.0

AudriusButkevicius:
Thanks!

yurishkuro:
np

PR #766, opened 2021-01-03
Add Intial GH Actions

### LINKED ISSUE #765: Migrate CI to GitHub Actions
Similar to https://github.com/jaegertracing/jaeger/issues/2645

* [x] migrate tests steps from .travis.yml to GHA workflows #775
* [ ] migrate release steps from .travis.yml to GHA workflows -- pending #766
* [ ] make sure artifact publishing & release process are working
* [x] update README build badge 6003c01
* [x] change required checks to new GH jobs
* [x] remove .travis.yml #803

@jaegertracing/jaeger-maintainers fyi if someone has the cycles

tylerauerbeck:
- Resolves #765 

- This is an initial attempt at moving CI to GH Actions. This includes:
  - **unit tests** - Runs unit tests, code coverage and runs publish on push actions
  - **release** - When a release-* tag is pushed, verifies a release, publishes the release and then creates a GH release (with changelog)
  - **crossdock** - Runs crossdock and publishes to dockerhub

I believe that I've captured the appropriate flows, but would like some extra eyes on the deploy logic from `.travis.yml` (i.e. the `ci-release.yml` flow and the logic for the `publish` step in `ci-unit-tests.yml`.


.github/workflows/ci-crossdock.yml
```diff
@@ -0,0 +1,70 @@
...
+
+    - name: Publish crossdock
+      run: |
+        REPO=jaegertracing/xdock-java
+        TAG=$(if [ "$BRANCH" == "master" ]; then echo "latest"; else echo "${BRANCH///}"; fi)
+        COMMIT=${GITHUB_SHA::8}
+        echo "BRANCH=$BRANCH, REPO=$REPO, TAG=$TAG, COMMIT=$COMMIT"
+        ./gradlew shadowJar
+        pushd 'jaeger-crossdock'
+        docker build -f Dockerfile -t $REPO:$COMMIT
```
yurishkuro, line 63: please check this PR https://github.com/jaegertracing/jaeger-client-node/pull/454 on how to avoid dual-build of the Docker image and reuse the one from `make crossdock-fresh`

.github/workflows/ci-crossdock.yml
```diff
@@ -0,0 +1,70 @@
...
+        popd
+        docker tag $REPO:$COMMIT $REPO:$TAG
+        docker push $REPO
+      if: |
+        github.event_name != 'pull_request' &&
+        steps.crossdock.outcome == 'success' &&
+        (steps.branch-pr.outputs.branch == 'master' || steps.branch-push.outputs.branch == 'master')
```
yurishkuro, line 70: I think if you check for `github.event_name == 'push'` then you only need to check the branch name from `steps.branch-push.outputs`

.github/workflows/ci-crossdock.yml
```diff
@@ -0,0 +1,70 @@
...
+      if: steps.dockerhub-login.outcome == 'success'
+
+    - name: Export Branch variable for pull_request event
+      id: branch-pr
+      run: |
+        export BRANCH=${GITHUB_HEAD_REF}
+        echo "Running on branch=$BRANCH"
+        echo "BRANCH=${BRANCH}" >> $GITHUB_ENV
+        echo ::set-output name=branch::$BRANCH
+      if: github.event_name == 'pull_request'
```
yurishkuro, line 40: Do we even need `$BRANCH` for pull requests?  It doesn't look like we're doing any push for PRs.

.github/workflows/ci-crossdock.yml
```diff
@@ -0,0 +1,70 @@
...
+      run: echo "DOCKERHUB_LOGIN=true" >> $GITHUB_ENV
+      if: steps.dockerhub-login.outcome == 'success'
+
+    - name: Export Branch variable for pull_request event
+      id: branch-pr
+      run: |
+        export BRANCH=${GITHUB_HEAD_REF}
+        echo "Running on branch=$BRANCH"
+        echo "BRANCH=${BRANCH}" >> $GITHUB_ENV
+        echo ::set-output name=branch::$BRANCH
```
yurishkuro, line 39: what does `::set-output` do?

.github/workflows/ci-crossdock.yml
```diff
@@ -0,0 +1,70 @@
...
+      with:
+        username: jaegertracingbot
+        password: ${{ secrets.DOCKERHUB_TOKEN }}
+      env:
+        DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
+      if: env.DOCKERHUB_TOKEN != null
+
+    - name: Export DOCKERHUB_LOGIN variable
+      run: echo "DOCKERHUB_LOGIN=true" >> $GITHUB_ENV
+      if: steps.dockerhub-login.outcome == 'success'
```
yurishkuro, line 31: We don't test for DOCKERHUB_LOGIN anywhere in the script, do we? If not, I would skip this step and test for `dockerhub-login.outcome` in the last step.

.github/workflows/ci-crossdock.yml
```diff
@@ -0,0 +1,70 @@
...
+      if: github.event_name == 'pull_request'
+    
+    - name: Export Branch variable for push event
+      id: branch-push
+      run: |
+        export BRANCH=${GITHUB_REF##*/}
+        echo "Running on branch=$BRANCH"
+        echo $BRANCH >> $GITHUB_ENV
+        echo ::set-output name=branch::$BRANCH
+      if: github.event_name == 'push'
```
yurishkuro, line 49: it seems that we could replace all of this with a single line `export BRANCH=${GITHUB_REF##*/}` in the last step.

.github/workflows/ci-crossdock.yml
yurishkuro: nvm https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-output-parameter

.github/workflows/ci-release.yml
```diff
@@ -0,0 +1,77 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - 'release-*'
```
yurishkuro, line 6: Is there a way you can test this workflow?

.github/workflows/ci-unit-tests.yml
```diff
@@ -0,0 +1,73 @@
...
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  unit-tests:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
```
yurishkuro, line 13: This feels rather dangerous given that the last step is deploy to Maven - what happens if we were to have more than one JDK versions in the matrix? Perhaps there should be another flag in the matrix controlling the deploy.

tylerauerbeck: Yeah, right now I'm only testing with 1.8, but it wasn't clear to me if multiple versions are being tested in Travis. Is there a set of versions we'd like to test against? If not, then we can probably just remove the matrix and things are fine. Otherwise, I think it's probably best to control the deploy via the matrix like you mention.

.github/workflows/ci-release.yml
tylerauerbeck: I was thinking about this as well. As it stands, I don't think there is. But I believe a good way (without triggering any actual releases) may be to trigger it manually via the `workflow_dispatch` event, but then just "skip" any of the real deploys that we don't want to happen.

.github/workflows/ci-crossdock.yml
tylerauerbeck: Good point. I'll clean this up.

.github/workflows/ci-crossdock.yml
tylerauerbeck: Resolved

.github/workflows/ci-crossdock.yml
tylerauerbeck: @yurishkuro If I'm reading that change right, I just need to add `image:` in the docker-compose file right? The difference between the node and the java example here is that I see two java's in the docker-compose file in this repo (`java-http` and `java-udp`) should it be one of those? Or should there be a seperate `java` declaration to do the build? Looks like the only difference between the two existing is the environment variable, so not sure how much that matters when it would be published.

yurishkuro: > If I'm reading that change right, I just need to add image: in the docker-compose file right?

I think so, and remove the 2nd build step.

> (java-http and java-udp) should it be one of those?

Yes, I think the UDP one is the "official" image that will be used by crossdock tests in other Jaeger SDKs. You can confirm it by comparing with what is being built in the 2nd build step.

tylerauerbeck: @yurishkuro Sorry it's taken a bit of time to get back to this. The only difference I see between the UDP and Dockerfile is that there is an environment variable being set (`sender`). Looks like that var is different for the http and udp variants, so I'm assuming that it has some effect in the code. Would it be better to  just add a "plain" java to the docker-compose file and then use that to push? Or if the var has no real impact, then we can just go ahead and use the UDP one that is already there.

yurishkuro: UDP is the default mode, if the `SENDER` env var is not specified it will default to UDP. 

https://github.com/jaegertracing/jaeger-client-java/blob/2be525bdcf7021b4d250f95ca4b6e118a684da4c/jaeger-crossdock/src/main/java/io/jaegertracing/crossdock/JerseyServer.java#L149-L152

So `java-udp` from docker-compose is the image we want to publish.

tylerauerbeck: Got it. This should be resolved now.

.github/workflows/ci-release.yml
tylerauerbeck: I think this is the only outstanding issue on this PR. Curious on what your thoughts are on the above or potentially taking a different route.

yurishkuro: It would be great if you could make it work in your fork and publish to your personal docker hub acct. May need to add some small overrides for names just for the test, and remove before merging.

albertteoh:
Hi @tylerauerbeck, are you planning on continuing with this PR? I think it would be fantastic to get it over the line given we're sort of in a limbo state with no Travis and outdated Travis release documentation.

PR #776, opened 2021-03-27
Add github action for publish to sonatype

Ashmita152:
Signed-off-by: Ashmita Bohara <ashmita.bohara152@gmail.com>

Add GitHub Action for publishing to Sonatype.

I have to still test this PR from my fork. Still trying to figure out steps to test it. Meanwhile opening the PR to get feedbacks.

It expects following secrets to be available to the job:
- ENCRYPTION_IV
- ENCRYPTION_KEY
- SIGNING_KEY_ID
- SIGNING_KEY_PASSPHRASE
- SONATYPE_USERNAME
- SONATYPE_PASSWORD

### DIFF
--- a/.github/workflows/ci-publish.yml
+++ b/.github/workflows/ci-publish.yml
@@ -1,0 +1,41 @@
+name: Publish
+
+on:
+  push:
+    branches: [ master ]
+    tags: [ 'release-*' ]
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: true
+
+    - uses: actions/setup-java@v1
+      with:
+        java-version: 8
+
+    - name: Export BRANCH variable
+      uses: ./.github/actions/setup-branch
+
+    - name: Configure git credentials
+      run: |
+        git config user.name "jaegertracingci"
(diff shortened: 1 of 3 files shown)


yurishkuro:
One thing worth remembering is that the release process does two commits, if I am not mistaken, one to change the version to release number, another to change it back to "snapshot". So the gh action may need to run several times, not just on release-* tag.

Btw, I tried to make gh action work with committing changes to the repo in the docs, and gave up after couple hours. Hopefully you can find a solution. May be worth experimenting on a test repo.

Ashmita152:
Hi @yurishkuro 

Thank you for the point about release process of this repo. I will validate it.

> Btw, I tried to make gh action work with committing changes to the repo in the docs, and gave up after couple hours. Hopefully you can find a solution. May be worth experimenting on a test repo.

Sorry Yuri, I didn't get this part. Are we talking about some issue with GH Actions of this repo that was added in the last PR ?

yurishkuro:
I'm referring to documentation repo, which also does commits during the release process. It used to work in Travis, but since switching to GHA the release commits cannot be pushed.

Ashmita152:
Sure Yuri, I will see if I can fix for it with a test repo.

PR #785, opened 2021-05-03
Fix repeatedly error logging for HttpSender (#732)

### LINKED ISSUE #732: RemoteReporter repeatedly logging error and success during unavailable agent
Followup to #728 

This is very likely a bug I introduced myself in #729 which got released with 1.3.2.
It seems to toggle between the states even though the agent has not been reachable. I'll investigate this.

lenoch7:
- repeatedly error logging when `HttpSender` is used
- unfortunately does not solve the problem when `UdpSender` is used (see description)

Resolves #732 - repeatedly error logging when `HttpSender` is used. The problem was that removing "failed" command from `commandFailedBefore` _HashSet_ was always done when command execution ends successfully (i.e. without throwing `SenderException`). But there are some situations, when "span is only saved in a buffer" on sender level (or there is no any span to be sent) and remote communication is not started at all.

Unfortunately the solution does not solve a problem if `UdpSender` is used. The reason is that `PortUnreachableException` is not thrown immediately, when `DatagramSocket.send(DatagramPacket)` method is invoked - this exception is thrown when subsequent call to send or receive is done. See Javadoc:
 - https://docs.oracle.com/javase/8/docs/api/java/net/DatagramSocket.html#send-java.net.DatagramPacket-
 - https://docs.oracle.com/javase/8/docs/api/java/net/DatagramSocket.html#connect-java.net.InetAddress-int-

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/reporters/RemoteReporter.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/reporters/RemoteReporter.java
@@ -127,8 +127,8 @@ public class RemoteReporter implements Reporter {
    * but are fine with executing sequentially.  The advantage is simplified code where
    * tasks are put onto a blocking queue and processed sequentially by another thread.
    */
-  public interface Command {
-    void execute() throws SenderException;
+  interface Command {
+    int execute() throws SenderException;
   }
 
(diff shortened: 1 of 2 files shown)


yurishkuro (approved):
lgtm, thanks

PR #794, opened 2021-09-06
Reduce locking in JaegerSpan

### LINKED ISSUE #792: Reducing locking in JaegerSpan
Looking at profiling data for some of our services, I noticed contention in the `synchronized` blocks of [JaegerSpan](https://github.com/jaegertracing/jaeger-client-java/blob/master/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpan.java).
I suggest replacing some of the `synchronized` blocks by other means to achieve thread safety.

amirhadadi:
Looking at profiling data from some of our services, we see contended locking involving JaegerSpan methods.
This change aims to reduce the amount of locking in JaegerSpan.
It also removes Collections.unmodifiableXXX wrappers around copied collections.
Resolves #792

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpan.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpan.java
@@ -26,6 +26,9 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+
 import lombok.extern.slf4j.Slf4j;
 
 /**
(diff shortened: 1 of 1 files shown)


jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpan.java
```diff
@@ -36,17 +39,19 @@
  */
 @Slf4j
 public class JaegerSpan implements Span {
+  private static final AtomicReferenceFieldUpdater<JaegerSpan, Long> durationMicrosUpdater =
```
yurishkuro, line 42: What is the reflection overhead, since this is on critical path?

amirhadadi: @yurishkuro this is a static field, so the reflection cost is paid only when the class is loaded.
As for the usage of this updater for `compareAndSet`, I saw an interesting writeup [here](https://shipilev.net/blog/2015/faster-atomic-fu/).

yurishkuro: That's a great article!

PR #799, opened 2021-09-14
Optimize TextMapCodec::contextFromString

### LINKED ISSUE #798: Optimize span context extraction in TextMapCodec
`TextMapCodec::contextFromString` is quite expensive in terms of CPU and memory allocation.
This is mostly due to the use of `BigInteger` to convert hex string to long.
We can optimize it by using a dedicated hex -> long conversion method based on `HexCodec::hexToUnsignedLong`.

amirhadadi:
Replace BigInteger for converting hex strings to longs
with a dedicated method based on HexCodec::hexToUnsignedLong.
Also change NumberFormatException thrown by BigInteger to MalformedTracerStateStringException on
empty hex strings and illegal digits.

Signed-off-by: amirhadadi <amirhadadi@hotmail.com>

Resolves #798


yurishkuro:
Please add tests to cover all newly introduced edge cases (namely missing various parts between `:`)

jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TextMapCodec.java
```diff
@@ -117,13 +141,11 @@ private static long high(String hexString) {
...
-        .append(Utils.to16HexString(context.getSpanId())).append(":")
-        // parent=0 is special, no need to encode as full 16 characters, and more readable this way
-        .append(context.getParentId() == 0 ? "0" : Utils.to16HexString(context.getParentId())).append(":")
-        .append(Integer.toHexString(intFlag))
-        .toString();
+    return context.getTraceId() + ":"
+            + context.toSpanId() + ":"
+            // parent=0 is special, no need to encode as full 16 characters, and more readable this way
+            + (context.getParentId() == 0 ? "0" : Utils.to16HexString(context.getParentId())) + ":"
+            + Integer.toHexString(intFlag);
```
yurishkuro, line 148: What's the motivation for removing StringBuilder?

amirhadadi: Cleaner code.
I guess the reason the code is using StringBuilder is due to the assumption that when using plain concatenation the compiler / runtime will generate less optimal code. [Here's](https://stackoverflow.com/a/1532499) some relevant discussion (pertaining to Oracle JDK 6).
From Java 9 onwards the byte code generated from string concatenation is [implemented using invokedynamic](https://openjdk.java.net/jeps/280) to allow future optimisations, giving further motivation to not hardcode concatenation using StringBuilder.

yurishkuro: but we're not using Java 9

jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TextMapCodec.java
```diff
@@ -73,41 +72,66 @@ private TextMapCodec(Builder builder) {
...
 
-    String traceId = parts[0];
-    if (traceId.length() > 32 || traceId.length() < 1) {
-      throw new TraceIdOutOfBoundException("Trace id [" + traceId + "] length is not withing 1 and 32");
+    if (traceIdEnd > 32 || traceIdEnd == 0) {
+      throw new TraceIdOutOfBoundException(
+              "Trace id [" + value.substring(0, traceIdEnd) + "] length is not within 1 and 32");
     }
 
-    // TODO(isaachier): When we drop Java 1.6 support, use Long.parseUnsignedLong instead of using BigInteger.
```
yurishkuro, line 90: can we use Long.parseUnsignedLong now that the build is on 1.8?

jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TextMapCodec.java
amirhadadi: I referred to the changes introduced in Java 9 just to show that plain concatenation is something that VM implementors are optimising for.
For Java 8 this code will get replaced by StringBuilder, so it's equivalent to the current implementation.

jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TextMapCodec.java
amirhadadi: Yes, but in Java 8 it would require creating temporary String objects.
In Java 9 there's a variant of [parseUnsignedLong](https://github.com/openjdk/jdk/blob/b29fbad940808c59f30e60222a9ca7a23c8e54b9/src/java.base/share/classes/java/lang/Long.java#L996) with the following signature:
`parseUnsignedLong(CharSequence s, int beginIndex, int endIndex, int radix)`
which allows to avoid creating temporary strings.

yurishkuro: please restore the TODO commend with your explanation

amirhadadi: @yurishkuro added a comment.

amirhadadi (commented):
@yurishkuro this change avoids lower casing all carrier keys.
There is a change in logic here in case `contextKey` or `baggagePrefix` are not lower case. The previous logic never matched them, but now they can be matched.

yurishkuro (approved):

PR #801, opened 2021-09-19
jaeger-thrift: Shade kotlin runtime + other libraries

### LINKED ISSUE #797: Shade kotlin runtime in shadow jar
Currently, the "shadow" jar (of jaeger-thrift) contains some unshaded libraries - mainly kotlin runtime (coming from OkHttp), which causes runtime errors when using a different version of kotlin.
There are other libraries that probably should be excluded from the upber-jar (or shaded if needed), e.g javax, opentracing, intellij annotations.

ohadgur:
Currently, the shadow jar contains unshadowed depdendecies - kotlin, javax, intellij/jetbrains annotations.
The main issue is the inclusion of kotlin runtime, which causes version conflicts when used with a different kotlin version.
This change adds relocation of the dependencies above. 
Resolves #797 

Signed-off-by: Ohad Gur <ohadgur@gmail.com>


yurishkuro (commented):
@pavolloffay PTAL

ohadgur:
@yurishkuro @pavolloffay can you please take a look?

yurishkuro:
what is the difference in the shadow jar size from this change?

ohadgur:
@yurishkuro there's not a real difference and there shouldn't be since the current shadow jar contains the same classes with their fqn (e.g. java.annotation..., kotlin.collections etc.). my change only relocated them to a different package.
before:
image

after:
image

yurishkuro (approved):

PR #805, opened 2021-10-02
Prepare 1.7.0 release changelog

yurishkuro:


CHANGELOG.md
```diff
@@ -1,10 +1,14 @@
 Changes by Version
 ==================
 
-1.7.0 (unreleased)
```
albertteoh, line 4: I think we should add a `1.8.0 (unreleased)` header.

PR #816, opened 2021-12-03
Add ability to filter and defer spans

### LINKED ISSUE #815: Deferred span reporting

We want to be able to generate full detailed spans only for operations that are "slow" while suppressing those where the entire top-level operation is "fast". Additionally, it would be useful for us to be able to filter out spans that are below some threshold.

Currently, there is no ability to conditionally generate or report spans using the Jaeger client. This means we currently have to choose between always creating many fine-grained spans that generate too much data, or always creating a few coarse-grained spans that lack the requisite detail to be useful in diagnosing performance issues.

We have implemented a prototype that allows for configuring a `FilteringReporter` that:

1. drops all spans below a filter threshold, and
2. defers spans below a defer threshold.

Deferred spans are either sent once a parent span exceeds the defer threshold, or are dropped if we get to the root span and it does not exceed that threshold.

I believe that since this filtering is based solely on the span duration and child span duration is always <= parent span duration, we avoid the span orphaning problem mentioned in #471.

Anyone using this configuration would have to be aware of the additional memory consumed by the pending spans as well as the potential for "burstier" reporting of the spans when they are sent to the delegate reporter. For our particular use case, our scalability testing showed no discernible performance impact.

sigix:
Signed-off-by: Edward Bross <edward.bross@appian.com>

- Resolves #815

- adds the ability to reduce the number of spans reported by setting a filter and a defer threshold
- implemented via a new `FilteringReporter` that handles the filtering and buffering logic before forwarding spans to a delegate

### DIFF
--- a/jaeger-core/README.md
+++ b/jaeger-core/README.md
@@ -209,6 +209,28 @@ The Logstash configuration might look like this:
         </root>
 ```
 
+### Duration-Based Filtering
+
+When instrumenting a system that generates a large volume of spans (e.g., an interpreter that creates a span
+for every function call, operator, etc.) it can be useful to only report those spans that are involved in
+slow operations. Spans can be filtered if they are below a hard limit while other spans not reported unless a
+parent span exceeds a defer threshold:
+
+```java
+Configuration.ReporterConfiguration reporterConfig = Configuration.ReporterConfiguration.fromEnv()
+    .withFilterSpansUnder(10L)
+    .withDeferSpansUnder(100L);
+
+Configuration configuration = new Configuration("foo")
+    .withReporter(reporterConfig);
+
+Tracer tracer = configuration.getTracer();
+```
+
+This can allow for a significant reduction in the amount of data produced and collected without losing details
+for slow traces, but at the cost of increased memory consumption as spans are temporarily buffered in memory.
+Internal metrics are available to monitor the number of pending spans.
+
 ## Development
 
 Especially in unit tests, it's useful to have tracer that is not connected to tracing backend, but collects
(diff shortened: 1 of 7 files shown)


yurishkuro:
I prefer not to merge this because this sort of naive filtering can easily lead to broken traces, it is not particularly general as a mechanism, and in the current form can be implemented as an extension instead of being part of the core lib.

Cc @pavolloffay @jpkrohling - thoughts?

jpkrohling:
There's some discussion about filtering in general here: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/6341

In this particular case, I see how it could be useful to filter out spans quicker than a specific threshold, but I have the feeling that this is too little: users will want a way to filter spans based on other characteristics. According to our deprecation plan, we are not accepting new features in about one week, so, there won't be a way to do any other type of filtering.

If this comes with good and clear documentation that this is useful to specific use cases and that there are known side-effects (perhaps with concrete examples), I'd be OK in accepting this feature. Otherwise, I think it might cause more harm than good.

sigix:
@jpkrohling I added a section to jaeger-core README.md with documentation and a use case.

@yurishkuro - using OpenTelemetry, I can see how to implement this via extension instead of modification, but it was not evident to me how to do the same with this library -- the configuration of the reporter seemed to be completely internal to the `Configuration` class with no accessible extension points. If there is a way to do this via extension, that would be just fine.

yurishkuro:
There's no configuration based extension mechanism, but you can always instantiate a custom reporter and pass it to the tracer.

PR #824, opened 2022-01-03
Restore support for setting null as a tag value (Resolves #823)

### LINKED ISSUE #823: JaegerSpan::setTag no longer supports null values in version 1.7.0
`JaegerSpan::setTag` supported null values until version 1.7.0.
[This change](https://github.com/jaegertracing/jaeger-client-java/pull/794/files#diff-f636dac472d8de558af04a132aece0b1af0d1b901cc2ff7430d284565cb09c95R71) in JaegerSpan changed the tags map implementation from `HashMap` to `ConcurrentHashMap` which doesn't support null values.

**To Reproduce**
Call `JaegerSpan::setTag` with a null value. No exception until 1.7.0, but you get NPE with 1.7.0.

**Expected behavior**
No NPE when calling `JaegerSpan::setTag` with a null value.

**Version (please complete the following information):**
 - Jaeger version: 1.7.0

amirhadadi:
To maintain backwards compatibility, we better restore the ability to set null as a tag value.
This change reverts the use of `ConcurrentHashMap` for the tags map in `JaegerSpan` back to `HashMap`, as `HashMap` supports null values while `ConcurrentHashMap` does not.

### DIFF
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpan.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpan.java
@@ -68,7 +68,7 @@ public class JaegerSpan implements Span {
     this.startTimeMicroseconds = startTimeMicroseconds;
     this.startTimeNanoTicks = startTimeNanoTicks;
     this.computeDurationViaNanoTicks = computeDurationViaNanoTicks;
-    this.tags = new ConcurrentHashMap<String, Object>();
+    this.tags = new HashMap<>();
     this.references = references != null ? new ArrayList<Reference>(references) : null;
 
     // Handle SAMPLING_PRIORITY tag first, as this influences whether setTagAsObject actually
@@ -110,8 +110,8 @@ public class JaegerSpan implements Span {
     return Collections.unmodifiableList(references);
   }
 
-  public Map<String, Object> getTags() {
-    return new HashMap<String, Object>(tags);
+  public synchronized Map<String, Object> getTags() {
+    return new HashMap<>(tags);
   }
 
   @Override
--- a/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpan.java
+++ b/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpan.java
@@ -68,7 +68,7 @@ public class JaegerSpan implements Span {
     this.startTimeMicroseconds = startTimeMicroseconds;
     this.startTimeNanoTicks = startTimeNanoTicks;
     this.computeDurationViaNanoTicks = computeDurationViaNanoTicks;
-    this.tags = new HashMap<>();
+    this.tags = new ConcurrentHashMap<>();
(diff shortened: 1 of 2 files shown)


yurishkuro (changes requested):
Jaeger backend does not support null values for tags. I agree we should fix the NPE, but I'd fix it but removing the key from the map.

amirhadadi:
> Jaeger backend does not support null values for tags. I agree we should fix the NPE, but I'd fix it but removing the key from the map.

@yurishkuro I looked at usages of `JaegerSpan::getTags`.
One usage is in [`JaegerThriftSpanConverter`](https://github.com/jaegertracing/jaeger-client-java/blob/7544006e1aaf42c35d1e9dbc086608f6f6964e56/jaeger-thrift/src/main/java/io/jaegertracing/thrift/internal/reporters/protocols/JaegerThriftSpanConverter.java#L57). It ends up calling [`String.valueOf(tagValue)`](https://github.com/jaegertracing/jaeger-client-java/blob/7544006e1aaf42c35d1e9dbc086608f6f6964e56/jaeger-thrift/src/main/java/io/jaegertracing/thrift/internal/reporters/protocols/JaegerThriftSpanConverter.java#L126) converting null to "null", so it seems it will succeed in sending the tag.
Same thing goes for [`ThriftSpanConverter`](https://github.com/jaegertracing/jaeger-client-java/blob/7544006e1aaf42c35d1e9dbc086608f6f6964e56/jaeger-zipkin/src/main/java/io/jaegertracing/zipkin/internal/ThriftSpanConverter.java#L90).

[`V2SpanConverter`](https://github.com/jaegertracing/jaeger-client-java/blob/7544006e1aaf42c35d1e9dbc086608f6f6964e56/jaeger-zipkin/src/main/java/io/jaegertracing/zipkin/internal/V2SpanConverter.java#L117) OTOH would fail with NPE when calling `tagValue.toString()`.

So removing null values from the map would be a breaking change, depending on how spans are being converted.

yurishkuro:
> V2SpanConverter OTOH would fail with NPE when calling tagValue.toString()

So the null check I am suggesting will fix this.

amirhadadi:
> > V2SpanConverter OTOH would fail with NPE when calling tagValue.toString()
> 
> So the null check I am suggesting will fix this.

@yurishkuro yes, and that's good, but my main point is what I wrote before that about `ThriftSpanConverter` and `JaegerThriftSpanConverter`. They use `String.valueOf(value)` instead of `value.toString()`.
Since `String.valueOf(null)` == `"null"`, the tag value is sent correctly when it's null.
So now we will be changing that behavior and the tag with "null" as value won't be sent anymore.

yurishkuro:
We can make the null check add a string value "null", ie push this conversion higher in the stack to the sdk instead of the reporter

amirhadadi:
@yurishkuro done.

yurishkuro (approved):

PR #826, opened 2022-01-25
Bump apache thrift version to 0.15.0

eugeniyk:
- fix https://issues.apache.org/jira/browse/THRIFT-5375

- Bump version to 0.15.0; https://github.com/apache/thrift/blob/0.15.0/CHANGES.md#java


yurishkuro:
We're no longer accepting PRs to this library, please see the notice in the README.

From the CI run it looks like a non-breaking change, which means you can pin Thrift version in your dependency manager without requiring a new release of Jaeger SDK.

